PR #23, opened 2016-09-09
Slim down mexFunction

dweindl:
Slim down mexFunction. Chop into smaller functions and move parts to amici.c for code reuse in matlab-free environment. No changes in functionality.


amiwrap.c
```diff
@@ -72,348 +71,15 @@ void mexFunction(int nlhs, mxArray *plhs[], int nrhs, const mxArray *prhs[]) {
...
-                    
-                    status = fx0(x,udata);
-                    if (status != AMI_SUCCESS) goto freturn;
-                    status = fdx0(x,dx,udata);
-                    if (status != AMI_SUCCESS) goto freturn;
-                    status = fsx0(NVsx, x, dx, udata);
-                    if (status != AMI_SUCCESS) goto freturn;
-                    
-                    if(status == 0) {
+    booleantype setupBdone = false;
```
FFroehlich: Variable definitions should be done before any code is called, otherwise we do not fullfill the C89 standard and AMICI cannot be compiled on VS2014<.

dweindl:
We are aiming for C89 compatibility? Omg... :-(

PR #40, opened 2016-11-03
Update Master

FFroehlich:
There some rather important fixes currently sitting in the stable branch that have not made it to the master branch yet. Development has slowed down and code has been extensively tested, so I guess it's safe to merge now


FFroehlich:
yes uploading the model directories is necessary to test compilation because they cannot be generated without matlab.

PR #57, opened 2017-02-28
Staging

FFroehlich:
fixed SBML import


symbolic/am_and.m
```diff
@@ -1,6 +1,11 @@
...
-fun = a*b;
+% am_and is the amici implementation of the symbolic and function
+%
+% Parameters:
+%  a: first input parameter @type sym
+%  b: second input parameter @type sym
+%
+% Return values:
+%  fun: logical value, negative for false, positive for true
+fun = am_min(a,b);
```
dweindl: how is am_and the same as am_min?

symbolic/am_or.m
```diff
@@ -1,6 +1,11 @@
...
-% fun = f(a,b);
+% am_or is the amici implementation of the symbolic or function
+%
+% Parameters:
+%  a: first input parameter @type sym
+%  b: second input parameter @type sym
+%
+% Return values:
+%  fun: logical value, negative for false, positive for true
 fun = am_max(a,b);
```
dweindl: how is "or" the same as "max"?

symbolic/am_and.m
FFroehlich: with the mapping

true -> positive
false -> negative

minimum  is only positive (true) if both a and b are positive (true)

symbolic/am_or.m
FFroehlich: with the mapping

true -> positive
false -> negative

maximum is only positive (true) if at least one of a and b is positive (true)

dweindl: Got it.

dweindl: That mapping is done within the symbolic toolbox?

FFroehlich: that mapping is implicitely done in AMICI at runtime. In CVODES events are defined as roots with odd multiplicity. In SBML events always fire when triggers transition from false to true.

The advantage of the continuous mapping as that math is a little bit more rigorous and the "distance" to true/false can be used for regularisation of event observables that did not happen during simulation

dweindl: Ah, k, thx

@SBMLode/SBMLode.m
```diff
@@ -62,7 +64,7 @@
             model.checkODE();
         end
         
-        importSBML(this,filename)
+        importSBML(this,modelname)
```
FFroehlich: you are correct, the argument should be filename and not modelname.

@SBMLode/importSBML.m
```diff
@@ -457,42 +656,87 @@ function makeSubs(this,old,new)
...
+function z = delay(x,y)
+error('Events with delays are currently not supported!');
+end
+
+function x = stoich_initAssign_rule(y,initassignments_sym,initassignments_math,rulevars,rulemath)
+x = {};
+for iy = 1:length(y)
+    x{iy} = sym(y(iy).stoichiometry);
+    if(~isempty(sym(y(iy).id)))
+        if(~isempty(initassignments_sym) || ~isempty(rulevars))
```
FFroehlich: correct, fixed.

FFroehlich:
valgrind says everything is fine, any objections on merge?

paulstapor:
If amiexamples gives the same results as it does on the master, no.

FFroehlich:
I checked and results for all examples look fine now (we should really set up regression tests for this).

PR #67, opened 2017-04-05
Apply chain-rule factor in C-code

dweindl:
Move results processing from simulate_*.m to C code.
First step: Use for non-matlab code.

+ few small fixes/enhancements


src/amici.cpp
```diff
@@ -2568,24 +2588,116 @@ ReturnData *getSimulationResults(UserData *udata, ExpData *edata, int *pstatus)
...
+    if(sensi_meth == AMI_SS) {
+        if(rdata->am_dxdotdpdata)
+            for(int ip = 0; ip < np; ++ip)
+                for(int ix = 0; ix < nx; ++ix)
+                    dxdotdpdata[ip*nx + ix] *= p[ip] * coefficient;
+
+        if(rdata->am_dydpdata)
+            for(int ip = 0; ip < np; ++ip)
+                for(int iy = 0; iy < ny; ++iy)
+                    dydpdata[ip*nx + iy] *= p[ip] * coefficient;
```
FFroehlich: this does not cover second order sensitivities, also I would like to not have this only for #ifdef AMICI_WITHOUT_MATLAB but also with matlab (should be relatively straightforward, just remove corresponding functionality in generateM.m, this would also serve as a sanity check whether all cases are covered)

dweindl: Okay, will check for second order sensitivities. 

The goal is definitely to also use it for matlab code. However, we started to work on that quite some time ago, and things started to diverge, I would rather do that incrementally.

Btw, this whole pull request got a bit messy, sorry.

FFroehlich:
okay the implemented chain-rules should now be correct (I did not check them for event outputs yet). the standalone hvp version is not fully functional yet, but we are missing a way to pass the vector there anyways, so I think it is sufficient to work on that later.

@dweindl can you please check why the travis compilation is failing? I recently merged the master branch, so something seems to be off in this branch :/

I will valgrind over all examples and as soon as everything is fixed, we are good to go.

dweindl:
@FFroehlich Cool! One Travis problem was the double parentheses you fixed in 6e23cc4. The remaining one is a null pointer for memcpy in https://github.com/ICB-DCM/AMICI/blob/feature_chainrulefactor_to_ccode_dw/src/amici.cpp#L2427

FFroehlich:
okay valgrind now runs through fine and results for examples are looking good. I think we are ready to merge

PR #73, opened 2017-04-27
Add number of total model parameters to UserData.

dweindl:


models/model_dirac/model_dirac_deltax.cpp
```diff
@@ -19,12 +19,6 @@ status = w_model_dirac(t,x,NULL,user_data);
 
               } break;
```
FFroehlich: did you remove the event for compilation?

@amimodel/generateC.m
```diff
@@ -347,7 +347,7 @@ function generateC(this)
 fprintf(fid,'#include <include/udata_accessors.h>\n');
 fprintf(fid,'                \n');
 fprintf(fid,'                void init_modeldims(UserData *udata){\n');
-fprintf(fid,['                   npmodel = ' num2str(this.np) ';\n']);
+fprintf(fid,['                   nplist = ' num2str(this.np) ';\n']);
```
FFroehlich: this should be np and not nplist!

dweindl: Correct. Will fix.

models/model_dirac/model_dirac_deltax.cpp
dweindl: I did not change anything in the model. This is what I got out of running the first 3 lines of the example_dirac.

@amimodel/generateC.m
dweindl: Ah, you did already. Thx.

models/model_dirac/model_dirac_deltax.cpp
FFroehlich: okay this should be fixed now

PR #85, opened 2017-05-08
Misc cleanup & minor extensions

dweindl:
* replace magic numbers and defines by enums
* hdf5 IO& error handling
* remove redundant userdata fields
* ...

### DIFF
--- a/src/amici.cpp
+++ b/src/amici.cpp
@@ -417,7 +417,7 @@ ReturnData *setupReturnData(mxArray *plhs[], UserData *udata, double *pstatus) {
 /* ------------------------------------------------------------------------------------- */
 
 #ifndef AMICI_WITHOUT_MATLAB
-ExpData *setupExpData(const mxArray *prhs[], UserData *udata) {
+ExpData *setupExpData(const mxArray *prhs[], UserData *udata, int *status) {
     /**
      * setupExpData initialises the experimental data struct
      * @param[in] prhs user input @type *mxArray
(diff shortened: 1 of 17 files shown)


src/amici.cpp
```diff
@@ -214,9 +214,9 @@ UserData *setupUserData(const mxArray *prhs[]) {
 
     readOptionData(z2event)
     readOptionData(qpositivex)
-    readOptionScalar(sensi,int)
+    readOptionScalar(sensi,AMI_sensi_order)
```
FFroehlich: This needs to be handled differently now. For both AMI_SENSI_ORDER_SECOND and AMI_SENSI_ORDER_SECOND_DIRECTIONAL we will have a value sensi==2. So we need to change this value depending on the mode with which the model was augmented (this is currently realised via the AMI_o2mode_TAG which is initialised in the init_modeldims function)

FFroehlich:
I think with respect to the AMI_sensi_order_TAG, I would not include one enum for directional and full because this will be specified during compile time (hardcoded in init_modeldims) while the sensi_order flag is set at runtime (it would not make sense to specify directional when you only compiled for second order).

src/amici.cpp
dweindl: Ah, great. More intuitive. I will remove `AMI_SENSI_ORDER_SECOND_DIRECTIONAL` then.

dweindl: Should be fixed in f37a398.

PR #89, opened 2017-05-10
Jak/Stat example as additional test case

dweindl:


tests/cpputest/wrapTestModels.m
```diff
@@ -4,20 +4,38 @@
...
-
-    [exdir,~,~]=fileparts(which('example_steadystate.m'));
-    amiwrap('model_steadystate','model_steadystate_syms',exdir)
-          
+    
+    try
+        [exdir,~,~]=fileparts(which('example_steadystate.m'));
+        amiwrap('model_steadystate','model_steadystate_syms',exdir);
+    catch
+        cd(fileparts(mfilename('fullpath')));
```
FFroehlich: don't you want to compile again after cd?

dweindl: The 'catch' was only meant to change back to the original working directory, no real error handling is taking place. The problem throwing the exception in the first place will probably persist.

FFroehlich: ah okay!

dweindl:
Tests are failing on osx due to numerical errors. What would be a reasonable tolerance (currently 1e-8)?
`numstepsS` and `numrhsevalsS` also differ. Does it make sense to check those?

FFroehlich:
numrhsevals(S) and numsteps(S) may differ, so I would not test for them. as far as I can tell you are using default integration tolerances, which means that rtol is 1e-8, as I would expect a loss of 2 orders of magnitude and the problematic value is ~0.2 so an error of magnitude 1e-7 is still okayish.
Is it possible to check both absolute and relative error and only throw an error if one of the two exceeds an threshold?

FFroehlich:
Alternatively set rtol to 1e-10, an absolute accuracy of 1e-8 is something that we should aim for

dweindl:
Okay, I will remove numrhsevals(S) and numsteps(S).

Checking for relative accuracy makes more sense indeed. Thresholds for both relative and absolute can be implemented, yes.

FFroehlich:
Okay but keep in mind that **either** relative or absolute tolerances must hold, not necessarily both.

tests/cpputest/jakstat_adjoint/tests1.cpp
```diff
@@ -6,6 +6,7 @@
 #include <cstring>
 #include "wrapfunctions.h"
 
+#define TEST_RTOL_JAKSTAT 1e-06
```
FFroehlich: does not need to be jakstat specific, with rtol 1e-8, I would not expect more than 1e-6 for any of the models

PR #90, opened 2017-05-11
Various bugfixes

FFroehlich:


src/amici.cpp
```diff
@@ -64,6 +64,19 @@ void runAmiciSimulation(UserData *udata, const ExpData *edata, ReturnData *rdata
     applyChainRuleFactorToSimulationResults(udata, rdata, edata);
 
 freturn:
+    if(*pstatus<0){
+        if(llhdata)
+            *llhdata = amiGetNaN();
+        
+        if(sllhdata)
+            for (int ip=0; ip<nplist; ip++)
+            sllhdata[ip] = amiGetNaN();
```
dweindl: fillArray(sllhdata, nplist, amiGetNaN());

fillArray(s2llhdata, nplist*ng, amiGetNaN());

would save some space

dweindl: or put it to a separate function 
void invalidateReturnData(ReturnData*, UserData*);

src/amici_interface_matlabo2.cpp
```diff
@@ -0,0 +1 @@
+#include "amici_interface_matlab.cpp"
```
dweindl: What is this file required for?

FFroehlich: this is necessary to have debug symbols for both first and secondorder compilation, otherwise object file will be overwritten.

dweindl: Hm.. but why is it even compiled twice? The object file should be usable for both versions.

PR #93, opened 2017-05-12
Use BLAS for fsy and fsjy

dweindl:
Use BLAS to compute fsy and fsjy because wrapping time and file size explodes when working with larger models with several observables.


src/amici.cpp
```diff
@@ -2185,3 +2192,90 @@ void processUserData(UserData *udata) {
...
+
+
+        // sJy += multResult
+        for(int ig = 0; ig < ng; ++ig) {
+            if(ig == 0)
+                for(int ip = 0; ip < nplist; ++ip)
+                    sJy[ip] -= multResult[ip];
+            else
+                for(int ip = 0; ip < nplist; ++ip)
+                    s2Jy[ip + nplist * ig] -= multResult[ip+ nplist * ig];
```
FFroehlich: s2Jy[ip + nplist * (ig-1)] -= multResult[ip+ nplist * ig];

right?

dweindl: Right. Well spotted, even before the test failed. Which did fail, luckily. Done: 464734e.

PR #97, opened 2017-05-17
bugfix and license update

FFroehlich:


src/amici.cpp
```diff
@@ -1473,12 +1474,18 @@ void initHeaviside(int *status, UserData *udata, TempData *tdata) {
 
     int ie;
 
-
     froot(t,x,dx,rootvals,udata);
 
     for (ie = 0; ie<ne; ie++) {
         if (rootvals[ie]<=0) {
-            h[ie] = 0.0;
+            if(rootvals[ie]==0){
```
dweindl: remove this line?

dweindl: if, else if, else instead of nested if would be easier to follow.

src/amici.cpp
```diff
@@ -153,6 +153,7 @@ void *setupAMI(int *status, UserData *udata, TempData *tdata) {
 
         /* initialise heaviside variables */
         initHeaviside(status,udata,tdata);
+        if (*status != AMI_SUCCESS) return(NULL);
```
dweindl: Never gonna happen, since status is not changes in initHeaviside.

PR #107, opened 2017-05-31
compatibility update MSVS

FFroehlich:


src/symbolic_functions.cpp
```diff
@@ -301,12 +303,12 @@ double am_spline(double t, int num, ...) {
     double ss;
     double dudt;
     
-    double ts[num];
-    double us[num];
+    double *ts = new double[num]();;
```
dweindl: If a significant amount of time is lost for (de)allocation here, i would use alloca instead.

src/symbolic_functions.cpp
```diff
@@ -352,13 +361,13 @@ double am_spline_pos(double t, int num, ...) {
     double ss;
     double dudt;
     
-    double ts[num];
-    double us[num];
-    double uslog[num];
+    double *ts = new double[num]();;
```
dweindl: same here

src/symbolic_functions.cpp
```diff
@@ -404,13 +421,13 @@ double am_Dspline(int id, double t, int num, ...) {
     double ss;
     double dudt;
     
-    double ts[num];
-    double us[num];
-    double ps[num];
+    double *ts = new double[num]();;
```
dweindl: and here

PR #125, opened 2017-07-01
Feature edatatdata classes

FFroehlich:


include/amici.h
```diff
@@ -8,72 +8,109 @@
...
+#define AMICI_SPBCG       7
+#define AMICI_SPTFQMR     8
+#define AMICI_KLU         9
+
+#define AMICI_NORMAL      1
+#define AMICI_LOGNORMAL   2
+#define AMICI_ONEOUTPUT   5
+
+
+#define AMICI_ERROR_UDATA         -99
```
dweindl: :+1:

include/edata.h
```diff
@@ -1,25 +1,26 @@
...
-
 /** @brief struct that carries all information about experimental data */
-typedef struct edata {
+class ExpData {
+
+public:
+    /**
+     * @brief Default constructor
+     */
+    ExpData(const UserData *udata);
```
dweindl: `ExpData` should implement `ExpData()` since `udata` is not required.

src/amici.cpp
```diff
@@ -181,436 +113,355 @@ void *setupAMI(int *status, UserData *udata, TempData *tdata) {
...
-                /* allocate some more temporary storage */
-                NVsx = N_VCloneVectorArray_Serial(udata->nplist, tdata->am_x);
-                sdx = N_VCloneVectorArray_Serial(udata->nplist, tdata->am_x);
-                if (NVsx == NULL) return(NULL);
-                if (sdx == NULL) return(NULL);
-
+    
+    if (udata->sensi >= AMICI_SENSI_ORDER_FIRST) {
+        if (udata->sensi_meth == AMICI_SENSI_FSA) {
+            if (udata->nx>0) {
```
dweindl: We could replace all these multiple checks by (not yet existing) `bool UserData::needFSA()` to increase readability.

src/amici.cpp
```diff
@@ -181,436 +113,355 @@ void *setupAMI(int *status, UserData *udata, TempData *tdata) {
...
-
+    
+    if (tdata->dxB == NULL) return AMICI_ERROR_SETUPB;
+    memset(NV_DATA_S(tdata->dxB),0,sizeof(realtype)*udata->nx);
+    
+    if (tdata->xQB == NULL) return AMICI_ERROR_SETUPB;
+    memset(NV_DATA_S(tdata->xQB),0,sizeof(realtype)*udata->nJ*udata->nplist);
+    
     /* create backward problem */
     if (udata->lmm>2||udata->lmm<1) {
```
dweindl: Should replace magic numbers

src/amici.cpp
```diff
@@ -181,436 +113,355 @@ void *setupAMI(int *status, UserData *udata, TempData *tdata) {
...
         errMsgIdAndTxt("AMICI:mex:lmm","Illegal value for lmm!");
     }
     if (udata->iter>2||udata->iter<1) {
         errMsgIdAndTxt("AMICI:mex:iter","Illegal value for iter!");
```
dweindl: again

src/amici.cpp
```diff
@@ -181,436 +113,355 @@ void *setupAMI(int *status, UserData *udata, TempData *tdata) {
...
-        case AMI_SPGMR:
-            *status = AMISpgmrB(ami_mem, which, PREC_NONE, 5);
-            if (*status != AMI_SUCCESS) return;
-
-            *status = wrap_SetJacTimesVecFnB(ami_mem, which);
-            if (*status != AMI_SUCCESS) return;
-
+            
+        case AMICI_SPGMR:
+            status = AMISpgmrB(ami_mem, tdata->which, PREC_NONE, 5);
```
dweindl: What does "5" mean here and below?

@amifun/getArgs.m
```diff
@@ -78,11 +78,11 @@
         case 'y'
             this.argstr = '(realtype t, int it, realtype *y, N_Vector x, void *user_data)';
         case 'z'
-            this.argstr = '(realtype t, int ie, int *nroots, realtype *z, N_Vector x, void *user_data)';
+            this.argstr = '(realtype t, int ie, realtype *z, N_Vector x, void *user_data, void *temp_data)';
```
FFroehlich: for some functions UserData needs to be passed as void as the pointer is handled by CVODES/IDAS. For all other it's true that the casting is not necessary and I will fix this in the future.

FFroehlich: Regarding passing individual pointers or only the Data classes, we currently anyways use all the udata members and do not pass them as arguments and this cannot change due to restrictions from CVODES, consequently we will always have a dependence on UserData which limits reusability and I would thus prefer to only pass UserData/TempData/ReturnData/ExpData as this will simplify the code a bit and make changes easier.

PR #129, opened 2017-07-03
alloca-h error for windows fixed

LeonardSchmiester:
Fix for "alloca.h: No such file or directory" error for windows taken from
https://github.com/slembcke/Chipmunk2D/issues/84


FFroehlich:
I would still keep the definition of _alloca for MSVS in there

dweindl:
Agreed. @LeonardSchmiester : Can you keep the original `#if` and add the additions below with `#elif` and check if that works?

LeonardSchmiester:
Like this:
#if _MSC_VER && !__INTEL_COMPILER
    #include <malloc.h>
    #define alloca _alloca
#elif defined(WIN32) || defined(__WIN32) || defined(__WIN32__)
    // For alloca().
    #include <malloc.h>
#else
    #include <alloca.h>
#endif

dweindl:
Yes, does that compile?
Otherwise 
` #if _MSC_VER && !__INTEL_COMPILER || defined(WIN32) || defined(__WIN32) || defined(WIN32) `
without elif

LeonardSchmiester:
I tried for 2 of the examples and it worked

PR #131, opened 2017-07-03
DW misc

dweindl:


dweindl:
@FFroehlich : Shall we use that template style also for the other model code? It would be easier to read and edit, but it would require another dozen template files...

FFroehlich:
yes, I can take care of that.

PR #132, opened 2017-07-04
cleanup ffuns + implementation fJz & co

FFroehlich:


FFroehlich:
not ready for merging yet as some sensitivities are still incorrect. gccode currently directly uses xdata classes, but this can now easily be changed when we implement https://github.com/ICB-DCM/AMICI/issues/130

FFroehlich:
sorry for making such a long pull request, I realised I had to fix quite a bit of code to get this running properly

src/amici.cpp
```diff
@@ -1857,33 +1777,33 @@ int applyChainRuleFactorToSimulationResults(const UserData *udata, ReturnData *r
...
-                if (rdata->sz)
-                for(int ip = 0; ip < udata->nplist; ++ip)
-                for(int iz = 0; iz < udata->nztrue; ++iz)
-                for(int it = 0; it < udata->nt; ++it)
-                rdata->sy[(ip * udata->nztrue + iz)*udata->nt + it] = rdata->z[(udata->nztrue + ip*udata->nztrue + iz)*udata->nt + it];
+                    if (rdata->sz)
+                        for(int ip = 0; ip < udata->nplist; ++ip)
+                            for(int iz = 0; iz < udata->nztrue; ++iz)
+                                for(int it = 0; it < udata->nt; ++it)
+                                    rdata->sy[(ip * udata->nztrue + iz)*udata->nt + it] = rdata->z[(udata->nztrue + ip*udata->nztrue + iz)*udata->nt + it];
```
dweindl: `rdata->sy` or `rdata->sz`?

src/amici.cpp
```diff
@@ -1902,14 +1822,14 @@ rdata->s ## QUANT [(ip * N1 + IND1) * N2 + IND2] *= pcoefficient[ip];} \
...
+                for(int ix = 0; ix < udata->nx; ++ix)
+                    rdata->dxdotdp[ip*udata->nxtrue + ix] *= pcoefficient[ip];
         
         if (rdata->dydp)
-        for(int ip = 0; ip < udata->nplist; ++ip)
-        for(int iy = 0; iy < udata->ny; ++iy)
-        rdata->dydp[ip*udata->nxtrue + iy] *= pcoefficient[ip];
+            for(int ip = 0; ip < udata->nplist; ++ip)
+                for(int iy = 0; iy < udata->ny; ++iy)
+                    rdata->dydp[ip*udata->nxtrue + iy] *= pcoefficient[ip];
```
dweindl: s/nxtrue/nytrue/?

src/amici.cpp
```diff
@@ -1975,88 +1895,339 @@ rdata->s ## QUANT [(ip*N1 + N1T + IND1)*N2 + IND2] += udata->k[udata->nk-udata->
...
 
-int fsJy(realtype t, int it, realtype *sJy, realtype *s2Jy, realtype *dJydy, realtype *dJydp, realtype *y, realtype *sigmay, realtype *sy, realtype *dydp, realtype *my, void *user_data) {
+static int fsJy(int it, UserData *udata, TempData *tdata, const ExpData *edata, ReturnData *rdata) {
+    int status = AMICI_SUCCESS;
+    
+    // Compute dJydx*sx for current 'it'
+    // dJydx        nt x nJ x nx
+    // sx           nt x nx x nplist
+    
+    realtype *multResult = new double[udata->nJ * udata->nplist]();
```
dweindl: Should not mix type names

src/amici.cpp
```diff
@@ -1975,88 +1895,339 @@ rdata->s ## QUANT [(ip*N1 + N1T + IND1)*N2 + IND2] += udata->k[udata->nk-udata->
...
+    for(int iyt=0; iyt < udata->nytrue; ++iyt) {
+        if (amiIsNaN(edata->my[udata->nt * iyt + it]))
+            continue;
+        
+        // copy current (iyt) dJydy and dJydsigma slices
+        // dJydyTmp     nJ x ny
+        // dJydsigmaTmp nJ x ny
+        for(int iJ = 0; iJ < udata->nJ; ++iJ) {
+            for(int iy = 0; iy < udata->ny; ++iy) {
+                dJydyTmp[iJ + iy * udata->nJ] = tdata->dJydy[iyt + (iJ + iy * udata->nJ) * udata->nytrue];
```
dweindl: We have these non-contiguous vectors in so many places. Is this due to how we get data from cvodes or due to matlab col-major storage? Is there a chance to change the memory layout and avoiding all the extra copying or do we anyways need to slice in all directions?

src/amici.cpp
```diff
@@ -1975,88 +1895,339 @@ rdata->s ## QUANT [(ip*N1 + N1T + IND1)*N2 + IND2] += udata->k[udata->nk-udata->
...
+        for(int iJ = 0; iJ < udata->nJ; ++iJ)
+            dJydxTmp[iJ + ix * udata->nJ] = tdata->dJydx[it + (iJ + ix * udata->nJ ) * udata->nt];
+    }
+    
+    // C := alpha*op(A)*op(B) + beta*C,
+    amici_dgemm(AMICI_BLAS_ColMajor, AMICI_BLAS_NoTrans, AMICI_BLAS_NoTrans,
+                udata->nJ, udata->nplist, udata->nx,
+                1.0, dJydxTmp, udata->nJ,
+                sxTmp, udata->nx,
+                1.0, multResult, udata->nJ);
```
dweindl: Change beta to 0 instead of initializing multResult in line 1940.

src/amici.cpp
```diff
@@ -1975,88 +1895,339 @@ rdata->s ## QUANT [(ip*N1 + N1T + IND1)*N2 + IND2] += udata->k[udata->nk-udata->
...
+    for(int ix = 0; ix < udata->nx; ++ix)
+        for(int iJ = 0; iJ < udata->nJ; ++iJ)
+            dJzdxTmp[iJ + ix * udata->nJ] = tdata->dJzdx[tdata->nroots[ie] + (iJ + ix * udata->nJ) * udata->nmaxevent];
+    
+    // C := alpha*op(A)*op(B) + beta*C,
+    amici_dgemm(AMICI_BLAS_ColMajor, AMICI_BLAS_NoTrans, AMICI_BLAS_NoTrans,
+                udata->nJ, udata->nplist, udata->nx,
+                1.0, dJzdxTmp, udata->nJ,
+                sxTmp, udata->nx,
+                1.0, multResult, udata->nJ);
```
dweindl: Change beta to 0 instead of initializing multResult.

src/amici_interface_matlab.cpp
```diff
@@ -288,6 +291,26 @@ void amici_dgemm(AMICI_BLAS_LAYOUT layout, AMICI_BLAS_TRANSPOSE TransA, AMICI_BL
...
+    assert(layout == AMICI_BLAS_ColMajor);
+    
+    const ptrdiff_t M_ = M;
+    const ptrdiff_t N_ = N;
+    const ptrdiff_t lda_ = lda;
+    const ptrdiff_t incX_ = incX;
+    const ptrdiff_t incY_ = incY;
+    const char transA = amici_blasCBlasTransToBlasTrans(TransA);
+    
+    assert(layout == AMICI_BLAS_ColMajor);
```
dweindl: once (line 297) is enough.

dweindl:
```
CMake Error at CMakeLists.txt:43 (add_subdirectory):
  The source directory

    /home/dweindl/src/AMICI.github/tests/cpputest/neuron

  does not contain a CMakeLists.txt file.


CMake Error at CMakeLists.txt:45 (add_subdirectory):
  The source directory

    /home/dweindl/src/AMICI.github/tests/cpputest/neuron_o2

  does not contain a CMakeLists.txt file.
```

@amifun/gccode.m
```diff
@@ -83,158 +83,90 @@
...
+            cstr = regexprep(cstr,'var_x_([0-9]+)','x_tmp[$1]');
+            cstr = regexprep(cstr,'var_sx_([0-9]+)','sx_tmp[$1]');
+            % sxdot needs to preces xdot
+            cstr = regexprep(cstr,'var_sxdot_([0-9]+)','sxdot_tmp[$1]');
+            cstr = regexprep(cstr,'var_xdot_([0-9]+)','xdot_tmp\[$1\]');
+            cstr = regexprep(cstr,'var_xBdot_([0-9]+)','xBdot_tmp\[$1\]');
+            cstr = regexprep(cstr,'xdot_old_([0-9]+)','xdot_old_tmp[$1]');
+            cstr = regexprep(cstr,'xdot_([0-9]+)','xdot_tmp[$1]');
+            cstr = regexprep(cstr,'var_xB_([0-9]+)','xB_tmp[$1]');
+            cstr = regexprep(cstr,'var_qBdot_([0-9]+)','qBdot_tmp\[ip + udata->nplist*$1]');
```
FFroehlich: you do not need to escape the escape string and actually easier to read if you don't. So I will remove unescessary escapes.

src/amici.cpp
FFroehlich: for dJ(y/z)dx you need to store the it slices for the initialisation of the adjoint sensitivities.
for sx it is the way it is stored in CVODES and the return struct.
for dJ(y/z)d(y/z) you need to store this dependent on whether the corresponding datapoint is non-nan.

This could probably be reduced by integrating dJydp,dJydx and sJy (same for z) into a single function as some of the slices could be reused and iyt could be passed to dJydy to only compute the specific slice, but I would recommend doing that in a separate PR.

src/amici.cpp
FFroehlich: nope, we loop over iyt here, so we add to multResult in every iteration!

src/amici.cpp
FFroehlich: hmm, if we rearrange the ordering of dJydy such that it is ordered [iJ,iy,iyt], we  would not need any slicing right? This should be easily doable.

dweindl: Yes, this should be done separately - was just wondering. 

dJydy would have to be ordered [iJ + iy * nJ + iyt * nJ * nytrue], for matlab that would be [iJ,iy,iyt], yes.

src/amici.cpp
dweindl: There is no iyt here.

scripts/run-build.sh
```diff
@@ -123,10 +123,6 @@ if [ $? -ne 0 ] ; then
...
 
-
-# Run tests
-export DYLD_LIBRARY_PATH="${DYLD_LIBRARY_PATH}:${SUNDIALS_BUILD_PATH}/lib:${SUITESPARSE_ROOT}/lib"
-
-echo $DYLD_LIBRARY_PATH
-
 cd ${AMICI_PATH}
+
+echo 'export DYLD_LIBRARY_PATH="${DYLD_LIBRARY_PATH}:${SUNDIALS_BUILD_PATH}/lib:${SUITESPARSE_ROOT}/lib' > scripts/env.sh
```
FFroehlich: 👍

PR #140, opened 2017-07-13
Cleanup

dweindl:


@amimodel/compileC.m
```diff
@@ -224,24 +224,26 @@ function compileC(this)
...
+            amiciSourceList = [amiciSourceList, '"', fullfile(this.wrap_path, 'src', [amiciSourceBaseNames{ii}, '.cpp']), '" '];
+        end
         eval(['mex ' DEBUG COPT ...
             ' -c -outdir ' fullfile(this.wrap_path,'src') ' ' ...
-            fullfile(this.wrap_path,'src','amici.cpp') ' ' ...
-            includesstr]);
+            amiciSourceList ' ' includesstr]);
     end
-        
-    if(this.nxtrue == this.nx)
```
FFroehlich: did this check disappear on purpose?

models/model_dirac/CMakeLists.txt
```diff
@@ -83,15 +85,16 @@ ${MODEL_DIR}/model_dirac_xBdot.cpp
 ${MODEL_DIR}/model_dirac_xdot.cpp
 ${MODEL_DIR}/model_dirac_y.cpp
 ${MODEL_DIR}/model_dirac_z.cpp
-${MODEL_DIR}/wrapfunctions.cpp
+ ${MODEL_DIR}/wrapfunctions.cpp
```
FFroehlich: unneccessary space

@amimodel/compileC.m
dweindl: Yes, it was checked twice before. Before each compilation mex call. Now compilation is done in a single call.

models/model_dirac/CMakeLists.txt
dweindl: Fixed in the template, but did not rewrap test models. Done.

@amimodel/compileC.m
FFroehlich: Okay, should be fine then.

Issue #123, opened 2017-06-26
Add AmiciSolver class


FFroehlich:
currently selection between CVODES and IDAS is done via inclusion of the headers cvodewrap.h or idawrap.h, which is intransparent and error prone. Instead make abstract base class AmiciSolver with concrete classes CVODESSolver and IDASSolver. Class should be instatiated as member of AmiciModel class (https://github.com/ICB-DCM/AMICI/issues/122), which holds information whether ODEs or DAEs are specified.

dweindl:
All `wrap_*()` from `wrapfunctions.h` should then be moved to solver-specific files. 

Solver-dependent functions such as CVodeInit should then not be called with the specific model function directly, but with a solver-specific wrapper instead. 

This wrapper should receive a pointer to the AmiciModel, UserData, and TempData and then call the AmiciModel member functions.

Issue #122, opened 2017-06-26
Add AmiciModel class


dweindl:
Create classes for AMICI models, add `AmiciModel` as base class. Remove wrapfunctions.cpp. This will help to organize code better when using AMICI from C++. Furthermore, it ensures that multiple models can be evaluated within a single program. (Will close #118 as well).

dweindl:
All `model_*.cpp` files will be kept. AmiciModel will delegate to these functions. This allows using and testing individual functions independently where required and will accelerate compilation.

AmiciModel members should take User/Exp/Result/TempData as arguments and call functions in `model_*.cpp` with plain C arrays as arguments.
This will prevent lengthy recompilation of large models everytime AMICI code is changed.

PR #143, opened 2017-07-14
Misc

dweindl:
Untangle amici.cpp to prepare for #122 #123 and reduce compilation time
- Slim down amici.cpp: turn some functions into *Data members
- Get rid of compile-time dependencies of amici.cpp on model code


@amimodel/compileC.m
```diff
@@ -176,35 +80,35 @@ function compileC(this)
         end
     end
     
-    if(this.cfun(1).(this.funs{j}))
-        fprintf(['ffuns | ']);
-        ffuns = cellfun(@(x) fullfile(this.wrap_path,'models',this.modelname,[this.modelname '_' x '.cpp']),this.funs,'UniformOutput',false);
+    
+    funsForRecompile = this.funs(structfun(@(x) logical(x), this.cfun(1)));
+    if(numel(funsForRecompile))
```
dweindl: It is subsetted, it contains only the function names that have to be recompiled. We could sum over cfun, yes, but we need the subset afterwards anyways.

@amimodel/compileC.m
```diff
@@ -123,40 +35,32 @@ function compileC(this)
     
     % generate hash for file and append debug string if we have an md5
     % file, check this hash against the contained hash
-    cppsrc = {'symbolic_functions','spline','edata','rdata','udata','tdata'};
+    cppsrc = {'amici', 'symbolic_functions','spline','edata','rdata','udata','tdata', 'amici_interface_matlab', 'amici_misc', 'amici_model_functions'};
```
dweindl: No, not anymore. I changed that here, so it is resolved at compile-time. The *wrap.h functions have the same interface. The implementation doesn't matter.

@amimodel/compileC.m
```diff
@@ -4,113 +4,25 @@ function compileC(this)
...
     else
         o_suffix = '.o';
     end
     
+    amiciSourcePath = fullfile(this.wrap_path,'src');
+    modelSourceFolder = fullfile(this.wrap_path,'models',this.modelname);
+
+    [objectsstr, includesstr] = compileAMICIDependencies(this.wrap_path, o_suffix);
+        
+    includesstr = strcat(includesstr,' -I"', modelSourceFolder, '"');
```
dweindl: Will do. Didn't even know this existed.

dweindl: As far as I can see, this is currently not handled very well. Once the debug libraries have been built, the only way going back to optimized libs is deleting them manually or a change in the version suitesparse/sundials version number. We need to handle that smarter.

dweindl: I propose to have the debug and optimized object files in two different directories, and link them depending on `model.debug`.

@amimodel/compileC.m
FFroehlich: 👍

@amimodel/compileC.m
FFroehlich: Ah, I missed that you use the logical values to index this.funs.

auxiliary/compileAMICIDependencies.m
```diff
@@ -0,0 +1,343 @@
...
+    includesstr = strcat(includesstr,' -I"', fullfile(ssparse_path, 'KLU','Include'), '"');
+    includesstr = strcat(includesstr,' -I"', fullfile(ssparse_path, 'AMD','Include'), '"');
+    includesstr = strcat(includesstr,' -I"', fullfile(ssparse_path, 'COLAMD','Include'), '"');
+    includesstr = strcat(includesstr,' -I"', fullfile(ssparse_path, 'BTF','Include'), '"');
+    includesstr = strcat(includesstr,' -I"', fullfile(ssparse_path, 'SuiteSparse_config'), '"');
+    
+    % compile all the sundials objects if we haven't done so yet
+    sources_sundials = getSourcesSundials();
+    for j=1:length(sources_sundials)
+        if(~exist(fullfile(objectFolder,objects_sundials{j}), 'file') || del_sundials)
```
FFroehlich: can we also compile all the objects in a single call to mex as we do it for the model specific files?

dweindl: Done. Also for src/*.cpp.

PR #144, opened 2017-07-25
move SBMLode to external repository to avoid code duplication

FFroehlich:


FFroehlich:
In terms of documenation for other developers? I thought you simply need to cd to the submodule and git pull for an update.

dweindl:
Rather for the non-developers who are still cloning the repo.
Hm, with an existing clone, I had to do it manually. Maybe different with a new one, didn't check that.

FFroehlich:
Shouldn't be necessary at all, but I will test this in more detail.

PR #148, opened 2017-07-31
Feature newtonsolver

paulstapor:
Just for having a look at the moment. I still need to implement the Newton solver in an example and to valgrind it...


src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+     * @param[in] udata pointer to the user data struct @type UserData
+     * @param[in] tdata pointer to the temporary data struct @type TempData
+     * @param[out] tdata pointer to the temporary data struct @type TempData
+     * @param[out] rdata pointer to the return data struct @type ReturnData
+     * @return void
+     */
+
+    int ix;
+    int inewt = 0;
+    int stat1 = -1;
```
dweindl: What is `stat1`? If status, please name so and use descriptive status code from `amici_defines.h` or add a new one if appropriate.

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+    N_VScale(1.0, tdata->x, tmp_x);
+    N_VAbs(tmp_x, tmp_x);
+    x_tmp = N_VGetArrayPointer(tmp_x);
+    for (ix=0; ix<udata->nx; ix++) {
+        if (x_tmp[ix] < udata->atol) {
+            x_tmp[ix] = udata->atol;
+        }
+    }
+    N_VDiv(tdata->xdot, tmp_x, rel_x);
+    rel_res = sqrt(N_VDotProd(rel_x, rel_x));
```
dweindl: Should compute only if `res < udata->atol`, otherwise not necessary

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+    if (res < udata->atol || rel_res < udata->rtol) {
+        // Clean up worksapce
+        N_VDestroy_Serial(delta);
+        N_VDestroy_Serial(rel_x);
+        N_VDestroy_Serial(tmp_x);
+        
+        // Return Success;
+        return (0);
```
dweindl: `return AMICI_SUCCESS;` instead of commenting

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+                x_tmp[ix] = 0.0;
+            }
+        }
+        
+        // Compute new xdot
+        fxdot(tdata->t, tdata->x, tdata->dx, tdata->xdot, udata);
+        
+        // Compute residuals
+        res = N_VDotProd(tdata->xdot, tdata->xdot);
+        res_old = N_VDotProd(tdata->xdot_old, tdata->xdot_old);
```
dweindl: Does `res_old` need to be recomputed? Can't we use the one from the previous iteration?

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+        res = sqrt(res);
+        res_old = sqrt(res_old);
+        
+        // if the step was useful, the dampening can be enlarged
+        if (res<res_old) {
+            // update x and F with x_try and F_try
```
dweindl: What are F, x_try and F_try?

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+            N_VScale(1.0, tdata->x, tdata->x_old);
+            N_VScale(1.0, tdata->xdot, tdata->xdot_old);
+            
+            // Check condition
```
dweindl: More detail or can omit

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+            if (res < udata->atol) {
+                // Return number of Newton steps needed
+                rdata->newton_numsteps[ntry-1] = inewt + 1;
+                stat1 = 0;
+                break;
+            }
+            
+            if (stat1 != 0)
```
dweindl: `status != AMICI_SUCCESS`

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+     * @param[in] udata pointer to the user data struct @type UserData
+     * @param[in] tdata pointer to the temporary data struct @type TempData
+     * @param[out] tdata pointer to the temporary data struct @type TempData
+     * @return void
+     */
+    
+     int il;
+     int stat1 = -1;
+     double rho;
+     double rho1;
```
dweindl: What is rho and rho1?

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+     *
+     * @param[in] ami_mem pointer to the solver memory block @type *void
+     * @param[in] udata pointer to the user data struct @type UserData
+     * @param[in] tdata pointer to the temporary data struct @type TempData
+     * @param[out] tdata pointer to the temporary data struct @type TempData
+     * @return void
+     */
+    
+     int il;
+     int stat1 = -1;
```
dweindl: see above

src/amici.cpp
dweindl: And alpha, beta, omega

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+     N_Vector ns_tmp = N_VNew_Serial(udata->nx);
+     N_Vector ns_Jdiag = N_VNew_Serial(udata->nx);
+    
+     N_VScale(-1.0, tdata->xdot, tdata->xdot);
+    
+    // Get the diagonal of the Jacobian for preconditioning
+    fJDiag(tdata->t, ns_Jdiag, tdata->x, udata);
+    
+    // Ensure positivity of entries in ns_Jdiag
+    N_VConst(1.0, ns_p);
```
dweindl: I don't understand what happens with `ns_p`. It it initialized with one, used to scale `ns_tmp` by 1.0 and then set to 0. 

Why replace N_VLinearSum by N_VScale and directly set `ns_p` to 0?

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+     int stat1 = -1;
+     double rho;
+     double rho1;
+     double alpha;
+     double beta;
+     double omega;
+     double res;
+     double rel_res;
+    
+     N_Vector ns_p = N_VNew_Serial(udata->nx);
```
dweindl: What are all those N_Vector?

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+     */
+    
+    
+    double res_abs;
+    double res_rel;
+    double sim_time;
+    int status = (int) *rdata->status;
+    realtype *x_tmp;
+    N_Vector rel_x = N_VNew_Serial(udata->nx);
+    N_Vector tmp_x = N_VNew_Serial(udata->nx);
```
dweindl: To have both `tmp_x` and `x_tmp` is somewhat confusing and seems like to attract bugs.

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+        N_VScale(1.0, tdata->x, tmp_x);
+        N_VAbs(tmp_x, tmp_x);
+        x_tmp = N_VGetArrayPointer(tmp_x);
+        for (int ix=0; ix<udata->nx; ix++) {
+            if (x_tmp[ix] < udata->atol) {
+                x_tmp[ix] = udata->atol;
+            }
+        }
+        N_VDiv(tdata->xdot, tmp_x, rel_x);
+        res_rel = sqrt(N_VDotProd(rel_x, rel_x));
```
dweindl: No need to compute `res_rel` before having checked `res_abs`

src/amici.cpp
```diff
@@ -1623,6 +1997,55 @@ int workBackwardProblem(UserData *udata, TempData *tdata, ReturnData *rdata, con
...
+     *
+     * @param[in] udata pointer to the user data struct @type UserData
+     * @param[in] tdata pointer to the temporary data struct @type UserData
+     * @param[out] tdata pointer to the temporary data struct @type TempData
+     * @param[out] rdata pointer to the return data struct @type ReturnData
+     */
+    
+    int status = (int) *rdata->status;
+    double run_time;
+    clock_t starttime;
```
dweindl: Unused.

src/amici.cpp
FFroehlich: if it is the status flag, please call it `status`, if is a temporary value `status_temp` (although in most cases a temporary value for status does not make any sense as you still want to return that value in case the functions stops prematurely).

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+    
+    // Check, how fxdot is used exactly within AMICI...
+    fxdot(tdata->t, tdata->x, tdata->dx, tdata->xdot, udata);
+    res = sqrt(N_VDotProd(tdata->xdot,tdata->xdot));
+    
+    // Check for relative error, but make sure not to divide by 0!
+    // Ensure positivity of the state
+    N_VScale(1.0, tdata->x, tmp_x);
+    N_VAbs(tmp_x, tmp_x);
+    x_tmp = N_VGetArrayPointer(tmp_x);
```
FFroehlich: having `tmp_x` and `x_tmp` in the same function with different types is very confusing to me. I recommend renaming `tmp_x` to `x` or `x_newton`

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+    
+    return(stat1);
+    
+}
+
+/* ------------------------------------------------------------------------------------- */
+/* ------------------------------------------------------------------------------------- */
+/* ------------------------------------------------------------------------------------- */
+
+int getNewtonStep(UserData *udata, ReturnData *rdata, TempData *tdata, void *ami_mem, int ntry, int nnewt, N_Vector ns_delta) {
```
FFroehlich: What is the difference of this function compared to cvNewtonIteration in https://github.com/ICB-DCM/AMICI/blob/master/sundials/src/cvodes/cvodes.c ?

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+    return(AMICI_SUCCESS);
+}
+
+/* ------------------------------------------------------------------------------------- */
+/* ------------------------------------------------------------------------------------- */
+/* ------------------------------------------------------------------------------------- */
+
+int getNewtonSimulation(void *ami_mem, UserData *udata, TempData *tdata, ReturnData *rdata) {
+    /**
+     * getNewtonSimulation solves the forward problem, if the first Newton solver run did not succeed.
```
FFroehlich: documented arguments do not match function signature

src/amici.cpp
```diff
@@ -1424,7 +1794,11 @@ int workForwardProblem(UserData *udata, TempData *tdata, ReturnData *rdata, cons
                         }
                     } else {
                         if (udata->nx>0) {
-                            status = AMISolve(ami_mem, RCONST(udata->ts[it]), tdata->x, tdata->dx, &(tdata->t), AMICI_NORMAL);
+                            if (isinf(udata->ts[it])) {
+                                status = workSteadyStateProblem(udata, tdata, rdata, ami_mem, it);
```
FFroehlich: nice! this will work for both -inf (preequilibration) and +inf!

src/amici.cpp
```diff
@@ -1383,6 +1384,375 @@ int getDiagnosisB(int it, void *ami_mem, UserData *udata, ReturnData *rdata, Tem
...
+        for (int ix=0; ix<udata->nx; ix++) {
+            if (x_tmp[ix] < udata->atol) {
+                x_tmp[ix] = udata->atol;
+            }
+        }
+        N_VDiv(tdata->xdot, tmp_x, rel_x);
+        res_rel = sqrt(N_VDotProd(rel_x, rel_x));
+    
+        /* residuals are small? */
+        if (res_abs < udata->atol || res_rel < udata->rtol) {
```
FFroehlich: can we have a loop here that loops simulations until we reach steadystate or exceed the maximum number of iterations?

@amioption/amioption.m
```diff
@@ -51,6 +51,16 @@
         x0 = double.empty();
         % custom initial sensitivity
         sx0 = double.empty();
+        % newton solver: linear solver (BiCGStab, GMRES, KLU direct)
```
paulstapor: Done.
I added bicgstab as linear solver 10 to the solvers in amici_defines.h

@amioption/amioption.m
```diff
@@ -51,6 +51,16 @@
...
+        newton_linsol = 0;
+        % newton solver: preconditioning method (none, diagonal, incomplete LU)
+        newton_precon = 1;
```
paulstapor: Done.
I added an enumeration for the precon-method

@amioption/amioption.m
```diff
@@ -51,6 +51,16 @@
...
+        % newton solver: maximum newton steps
+        newton_maxsteps = 20;
+        % newton solver: maximum linear steps
+        newton_maxlinsteps = 150;
+        % preequilibration of system via newton solver
+        preequil = 0;
```
paulstapor: Done.

include/rdata.h
```diff
@@ -114,6 +114,21 @@ class ReturnData {
     /** employed order forward problem (dimension: nt) */
     double *order;
     
+    /** flag success of Newton solver */
```
paulstapor: Done.

include/rdata.h
```diff
@@ -114,6 +114,21 @@ class ReturnData {
...
+    double *newton;
+    
+    /** number of Newton steps steady state problem */
```
paulstapor: time in seconds, added.

include/rdata.h
```diff
@@ -114,6 +114,21 @@ class ReturnData {
...
+    double *newton_time;
+    
+    /** number of Newton steps steady state problem */
+    double *newton_numsteps;
+    
+    /** number of linear steps per Newton step steady state problem */
+    double *newton_numlinsteps;
+    
+    /** steady state found be Newton solver */
```
paulstapor: Done.

src/amici.cpp
paulstapor: Thanks for pointing out!
I have no clue. I never looked into the one from cvodes (and didn't know it exists) but adapted Glenns code into Amici. For the current commit, I would like to have my routine in Amici, since I know what it does and I don't want bigger changes before the merge. In the very near future, I will have a closer look at the cvodes one and maybe replace mine.

src/amici.cpp
paulstapor: That's what I've actually planned: I wanted to define an event, which is triggered if xdot is small. The problem: The event needs to be added to the other events before compiling...
That is maybe not super-nice and functionality of the Newton solver would be depending on the functionality of the events forever, so I decided against it.
A loop with single-steps might be a neat idea, I also thought about it and is probably the way to go. But this, I wanted to discuss with you offline in more detail before implemented. Right now, the merge ist more important and the current implementation works fine.

src/amici.cpp
paulstapor: Thanks!
It would, yes. Preequlibration however, I wanted to do via a flag which can be
 * 0 - no preequilibration
 * 1 - preequilibration of states only
 * 2 - preequilibration of states and sensitivities
However, doing this directly in the simulation would keep the code simpler...
We should discuss that offline.

src/amici.cpp
paulstapor: The whole getNwtonStep-routine is far from nice, I know. It is basically only the bicgstab algorithm, which Glenn took from the kinsol package...
I will replace it as soon as possible and take it out of amici.cpp, but for this short moment, it does what it should: it solves a linear system.

src/amici.cpp
paulstapor: As above.

src/amici.cpp
paulstapor: Not exactly. It's not used to scale ns_tmp, but it's added to it, as a constant vector. To be honest: I don't really know why, but it works...

src/amici.cpp
paulstapor: Done.

src/amici.cpp
paulstapor: Conceptual question: The solution is accepted if *one* of res_abs and res_rel is small enough, or if *both* are? The current implementation is in the sense that *one* is small? Shall both be?

FFroehlich: Only one needs to be small.

src/amici.cpp
FFroehlich: you don't need that, simply specifying -inf as starting time should do the trick. You cannot preequilibrate sensitivities. If you preequilibrate and want to compute sensitivities in the following, you need to (always) initialize the sx0 with the steady state sensitivities.

src/amici.cpp
dweindl: True.

paulstapor:
Okay, checks have passed. I think I addressed all of your remarks, or gave a reason why I would not change it in this pull request...
Anything left to do?

src/amici.cpp
FFroehlich: ☝️this was not fixed yet

src/amici.cpp
FFroehlich: ☝️ what about this?

paulstapor: now it is used again. Actually, starttime should be a clock_t type, tun_time a double. The way it was before (with starttime superfluous) was actually incorrect, since run_time was used as double and clock_t. For some reason, the compiler didn't worry. But since it is improper, I changed it back.

dweindl: It is used now, but I just noticed:
>             run_time = (double)((clock() - starttime) * 1000) / CLOCKS_PER_SEC;
This does not give you seconds, but 1e3 seconds.

src/amici.cpp
paulstapor: I'll take care of that, thanks

Issue #136, opened 2017-07-09
move id and z2event to getUserData()


FFroehlich:
These two are model dependent constant and should not be in amioptions. They must not be set by the user.

Issue #130, opened 2017-07-03
Decouple model functions from UserData and other amici core code


dweindl:
Change
```            
int wrap_init(void *cvode_mem, N_Vector x, N_Vector dx, realtype t){
    return CVodeInit(cvode_mem, xdot_model_steadystate, RCONST(t), x);
}
```
and others to  `return CVodeInit(cvode_mem, fxdot, RCONST(t), x);`

and
```
int fxdot(realtype t, N_Vector x, N_Vector dx, N_Vector xdot, void *user_data){
       return xdot_model_steadystate(t, x, xdot, user_data);
}
```

to `return xdot_model_steadystate(t, x, xdot, udata->x, udata->...., ....);`

This will avoid lengthy recompilation in case of amici updates.

dweindl:
Unfortunately, this is slightly more complicated since CVodeInit and IDAInit require different signatures. We would have to add another layer of abstraction. Maybe we wait with that until after #123 .

PR #156, opened 2017-08-03
Split up amici.cpp into classes

### LINKED ISSUE #99: Make more model parameters specifiable at simulation time
the parameter transformation could be set at simulation time but can currently only be set at compilation time, which is an unnecessary limitation.

dweindl:
* Better separation of responsibilities
  * Add Solver class #123, keep ami_mem contained there
  * Add Modell class #122
  * Split into ForwardProblem, BackwardProblem, SteadystateProblem
  * #136
  * Move temporary results from UserData to TempData
  * UserData is not altered by AMICI
  * ...
* More flexibility (e.g. #99)
* Cleanup of function signatures and includes

Affects #130
Closes #99, #122, #123, #136


include/idawrap.h
```diff
@@ -130,8 +131,21 @@ class IDASolver : public Solver
         return(0);
     }
 
-    int AMISetId(N_Vector id) {
-        return IDASetId(ami_mem, id);
+    int AMISetId(Model *model) {
```
FFroehlich: id should be a constant vector that is set at model initialisation. It must not be modified by the user.

dweindl:
Agreed. I will add Solver to TempData which will not be seen by the user. The initial idea was to be able to set solver options directly to the Solver object. That is not used so far, so it will be safer to hide it completely.

Issue #135, opened 2017-07-05
add more CI features in travis


FFroehlich:
- [x] code coverage via `gcov` and `coveralls`: https://github.com/eddyxu/cpp-coveralls
- [x] static analysis via `cppcheck`: https://arne-mertz.de/2017/04/continuous-integration-travis-ci/
- [x] memory leak via `valgrind`
~- [ ] coding style checks via `OCLint`?~

PR #172, opened 2017-08-26
Valgrind for travis

dweindl:
Ad #135: Run valgrind on the test suite.


dweindl:
Agreed. But since there is currently practically no dedicated osx code, there shouldn't be any specific leaks that would escape valgrind on linux.

PR #173, opened 2017-08-27
coveralls implementation

FFroehlich:
need CI to check whether this is actually working


scripts/run-coveralls.sh
```diff
@@ -0,0 +1,25 @@
...
+
+lcov --base-directory ${AMICI_PATH} --directory ${AMICI_PATH} --zerocounters -q
+
+for MODELSTR in $TESTMODELS; do
+    MODEL=${MODELSTR#model_}
+    cd ${AMICI_PATH}/tests/cpputest/build/${MODEL}
+    ./model_${MODEL}_test
+done
+
+cd ${AMICI_PATH}/build/CMakeFiles/amici.dir/src
```
dweindl: Why do we need to run it in this directory? 

Is there a (simple) way to get nicer file names in the result table (i.e. relative to $AMICI_PATH)?

FFroehlich: I am currently trying to figure this out, in the lcov report it is fine, but somehow coveralls messes things up, not quite sure why. Overall coveralls is quite poorly documented :(

PR #176, opened 2017-08-29
Feature newtonsolver

paulstapor:
Preequilibration is now possible by setting an option.

### DIFF
--- a/include/rdata.h
+++ b/include/rdata.h
@@ -132,8 +132,11 @@ class ReturnData {
     /** number of linear steps by Newton step for steady state problem */
     double *newton_numlinsteps;
 
-    /** steady state found be Newton solver */
-    double *xss;
+    /** preequilibration steady state found be Newton solver */
+    double *x0;
+    
+    /** preequilibration sensitivities found be Newton solver */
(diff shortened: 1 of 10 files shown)


include/rdata.h
```diff
@@ -132,8 +132,11 @@ class ReturnData {
     /** number of linear steps by Newton step for steady state problem */
     double *newton_numlinsteps;
 
-    /** steady state found be Newton solver */
-    double *xss;
+    /** preequilibration steady state found be Newton solver */
```
dweindl: "by"

include/rdata.h
```diff
@@ -132,8 +132,11 @@ class ReturnData {
...
+    double *x0;
+    
+    /** preequilibration sensitivities found be Newton solver */
```
dweindl: "by"

include/rdata.h
dweindl: Will that be clear to the user that this x0 is not his supplied initial states?

FFroehlich:
I would generally recommend fixing things that were critized, even if the pull request was approved ...

PR #186, opened 2017-09-02
Update UserData from hdf5

dweindl:
Read all options, x0, sx0, pbar and Newton solver options and make hdf5 input and UserData setters more robust.


FFroehlich:
okay was only createTestingData that failed due to problem with sparse newton solver, fixed in 4e06409

PR #192, opened 2017-09-18
bugfix for adjoints with user-provided initial sensitivities

paulstapor:
A little fix in the adjoint code: So far, Amici did not use initial states and sensitivities provided by the user via amioptions for adjoint senstiivity computation.
If there is no test case for this yet: I'll implement the Histone model as Amici example in two weeks, to have another test-example for  the Newton solver, preequilibration, and this.


paulstapor:
Thanks!

PR #196, opened 2017-10-13
Enhancement o2adjoints

paulstapor:
Atm, I just want to see possible memory leaks and so on....


include/amici_model.h
```diff
@@ -6,6 +6,10 @@
 #include <sundials/sundials_direct.h>
 #include <sundials/sundials_sparse.h>
 
+int fdJdx(realtype t, N_Vector x, N_Vector dx, void *user_data);
+int fdJdp(realtype t, N_Vector x, N_Vector dx, void *user_data);
+int fddxdotdpdp(realtype t, N_Vector x, N_Vector dx, void *user_data);
```
dweindl: These functions need to become virtual member functions of `Model` with an empty default implementation, otherwise linking non-second-order models won't work.

dweindl:
@paulstapor : What is the status of `enhancement_o2adjoints`? You closed the pull request, but the respective branch still exists.

paulstapor:
the improvement I tried to implement finally didn't work out as I hoped...

dweindl:
Ah, I see. So shall we delete this branch?

dweindl:
Oh, I see you did.

PR #197, opened 2017-10-14
Boost::serialization support for UserData and ReturnData 

dweindl:
Boost::serialization support for UserData and ReturnData and corresponding tests. Allows saving those data in various formats. Not relevant for use with Matlab.

### DIFF
--- a/include/edata.h
+++ b/include/edata.h
@@ -19,9 +19,9 @@ class ExpData {
 
     ~ExpData();
 
-    /** observed data (dimension: nytrue x nt, column-major) */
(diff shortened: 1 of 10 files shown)


include/edata.h
```diff
@@ -19,14 +19,15 @@ class ExpData {
 
     ~ExpData();
 
-    /** observed data */
+    /** observed data (dimension: nytrue x nt, column-major) */
```
FFroehlich: are you sure? in matlab (column major) this is supposed to be nt x nytrue.

dweindl: You are absolutely right.

PR #216, opened 2017-11-03
Use cmake variables for library suffix

dweindl:


dweindl:
@FFroehlich Can you please check if this works on your machine too?

FFroehlich:
also works on my machine :)

PR #222, opened 2017-11-17
decouple and refactor model code

FFroehlich:


CMakeLists.txt
```diff
@@ -83,9 +83,9 @@ set(AMICI_SRC_LIST
 add_library(${PROJECT_NAME} ${AMICI_SRC_LIST})
 
 target_link_libraries(${PROJECT_NAME}
-    ${SUNDIALS_LIB_DIR}/libsundials_nvecserial${CMAKE_SHARED_LIBRARY_SUFFIX}
-    ${SUNDIALS_LIB_DIR}/libsundials_cvodes${CMAKE_SHARED_LIBRARY_SUFFIX}
-    ${SUNDIALS_LIB_DIR}/libsundials_idas${CMAKE_SHARED_LIBRARY_SUFFIX}
+    ${SUNDIALS_LIB_DIR}/libsundials_nvecserial${CMAKE_STATIC_LIBRARY_SUFFIX}
+    ${SUNDIALS_LIB_DIR}/libsundials_cvodes${CMAKE_STATIC_LIBRARY_SUFFIX}
+    ${SUNDIALS_LIB_DIR}/libsundials_idas${CMAKE_STATIC_LIBRARY_SUFFIX}
```
dweindl: Why only static linking of sundials and not of suitesparse?

FFroehlich: the build scripts don't generate static versions of those libraries yet ;)

FFroehlich:
@dweindl  you may review this PR now :)

src/amici_hdf5.cpp
```diff
@@ -16,28 +17,32 @@
...
     UserData *udata = AMI_HDF5_readSimulationUserDataFromFileObject(
-        file_id, datasetPath, model);
+        file_id, datasetPath);
 
     H5Fclose(file_id);
 
     return (udata);
 }
 
+template <class hid_t>
```
dweindl: Why templated?

src/amici_hdf5.cpp
```diff
@@ -16,28 +17,32 @@
...
 UserData *AMI_HDF5_readSimulationUserDataFromFileObject(hid_t fileId,
-                                                        const char *datasetPath,
-                                                        Model *model) {
+                                                        const char *datasetPath) {
     assert(fileId > 0);
+    
+    int nx,nk,np;
+    getModelDims(&nx,&nk,&np);
```
dweindl: I added the `Model` argument so that it is clear for which model the data is to be loaded, that didn't look super nice. However, using getModelDims() makes it impossible to use this function in an application which uses two different models. I would argue that we should use getModel, getModelDims, ... nowhere other than in mexFunction and inside tests.

src/amici_hdf5.cpp
```diff
@@ -144,9 +151,10 @@ UserData *AMI_HDF5_readSimulationUserDataFromFileObject(hid_t fileId,
     if (AMI_HDF5_attributeExists(fileId, datasetPath, "pbar")) {
         status += AMI_HDF5_getDoubleArrayAttribute(fileId, datasetPath, "pbar",
                                                    &buffer, &length0);
-        if (length0 != 0 && length0 != (unsigned)udata->nplist)
+        if (length0 != 0 && length0 == (unsigned)udata->nplist())
             goto freturn;
-        udata->setPbar(buffer);
+        if(length0 == (unsigned)udata->nplist())
```
dweindl: We will never get there unless length0 == 0 and nplist == 0. Is this intended?

src/amici_interface_matlab.cpp
```diff
@@ -94,19 +100,19 @@ enum mexRhsArguments {
...
- * dimension checks @type Model
- * @return udata pointer to user data object @type UserData
+ * @param nrhs number of input arguments of the matlab call
+ * @param prhs pointer to the array of input arguments
+ * @return udata pointer to user data object
  */
-UserData *userDataFromMatlabCall(const mxArray *prhs[], int nrhs,
-                                 Model *model) {
+template <class mxArray>
+UserData *userDataFromMatlabCall(const mxArray *prhs[], int nrhs) {
```
dweindl: Similar as above: I think passing the model instance is much clearer than using the model wrapper functions.

src/amici_interface_matlab.cpp
```diff
@@ -160,6 +138,7 @@ UserData *userDataFromMatlabCall(const mxArray *prhs[], int nrhs,
         readOptionScalarInt(stldet, booleantype);
         readOptionData(qpositivex);
         readOptionScalarInt(sensi, AMICI_sensi_order);
+        // this needs to be set before we set parameters to obtain properly unscaled parameters
```
dweindl: Are the unscaled parameters updated when pscale is changed? I think they should be, otherwise this begs for trouble. But I agree that we should set pscale first.

src/amici_interface_matlab.cpp
```diff
@@ -376,7 +383,7 @@ ExpData *expDataFromMatlabCall(const mxArray *prhs[], const UserData *udata,
     int ne_mz = 0, nz_mz = 0, ne_sigmaz = 0,
         nz_sigmaz = 0; /* integers with problem dimensionality */
     ExpData *edata = new ExpData(udata, model);
-    if (edata->my == nullptr) {
+    if (edata->my.empty()) {
         // if allocation fails we throw an exception
```
dweindl: If allocation fails, std::bad_alloc should have been thrown already and we will never get here. I think we can remove this block.

@amifun/gccode.m
```diff
@@ -73,101 +73,93 @@
...
+            cstr = regexprep(cstr,'var_xdot_([0-9]+)','xdot[$1]');
+            cstr = regexprep(cstr,'var_xBdot_([0-9]+)','xBdot[$1]');
+            cstr = regexprep(cstr,'xdot_old_([0-9]+)','xdot_old[$1]');
+            cstr = regexprep(cstr,'xdot_([0-9]+)','xdot[$1]');
+            cstr = regexprep(cstr,'var_xB_([0-9]+)','xB[$1]');
+            cstr = regexprep(cstr,'var_dxB_([0-9]+)','dxB[$1]');
+            cstr = regexprep(cstr,'var_qBdot_([0-9]+)','qBdot[$1]');
+            cstr = regexprep(cstr,'var_v_([0-9]+)', 'v[$1]');
+            cstr = regexprep(cstr,'var_vB_([0-9]+)', 'vB[$1]');
+            cstr = regexprep(cstr,'var_JSparse_([0-9]+)', 'JSparse->data[$1]');
```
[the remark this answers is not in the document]
FFroehlich: the problem is that for the sparse matrices we need to set the rowptrs and colptrs which are part of the Slsmat struct. Currently these are also set in model_JSparse. We could introduce additional functions that set those, but in the end the data is meaningless without those indices.

src/steadystateproblem.cpp
```diff
@@ -256,38 +246,36 @@ void SteadystateProblem::getNewtonSimulation(const UserData *udata, TempData *td
...
     
     /* Loop over steps and check for convergence */
     double res_abs = INFINITY;
     double res_rel = INFINITY;
-    realtype *x_tmp;
     
     int it_newton = 0;
     while(res_abs > udata->atol && res_rel > udata->rtol) {
         /* One step of ODE integration */
-        solver->AMISolve(1e12, tdata->x, tdata->dx, &(tdata->t),
```
dweindl: Magic number

tests/cpputest/testfunctions.cpp
```diff
@@ -10,75 +10,71 @@
 #include <cmath>
 #include <unistd.h>
 
+
+extern amici::Model *getModel(const amici::UserData *udata);
```
dweindl: See comment above. I would not use those extern functions anywhere but in top-level code. It makes it very hard so see the dependencies. I don't see a downside to passing the model object as additional argument. (Maybe except for the UserDataFromHDF5, where it does not seem very plausible but that can be fixed).

@amifun/gccode.m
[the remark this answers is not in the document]
dweindl: My fault, I somehow thought it was `std::vector`, but it is `SlsMat` of course. Ignore that one.

dweindl:
TempData is gone :tada:.  Nice.

Since parameters, plist, ... are now in `Model`, I am wondering if it conceptually still makes sense to keep them inside `UserData`. Or if UserData was better renamed to AmiciOptions.

FFroehlich:
true! On the long run AmiciOptions should only be passed on construction of `ReturnData` (which can be recycled across simulations) and parameters are only passed when requesting simulation results (and stored in `Model`, which resets the model). However, this should probably be adressed in a different PR ;)

include/amici_exception.h
```diff
@@ -22,15 +28,69 @@ namespace amici {
...
+            free(symbols);
+            if (nFrames == nMaxFrames)
+            trace_buf << "[truncated]\n";
+            snprintf(trace, sizeof(trace), "(%s)", trace_buf.str().c_str());
+        }
+        
     private:
-        char msg[1000];
+        char msg[500];
+        char trace[500];
```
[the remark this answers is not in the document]
FFroehlich: for trace we could, for msg there is no printf like interface for strings right?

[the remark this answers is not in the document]
dweindl: Okay, `vsnprintf` to `std::string` will not make it nicer. So probably better to leave it that way.

include/amici_model.h
```diff
@@ -916,26 +769,193 @@ namespace amici {
...
+        
+        void getz(const int nroots, const ReturnData *rdata);
+        
+        void getrz(const int nroots, const ReturnData *rdata);
+        
+        /** Jacobian */
+        SlsMat J = nullptr;
+        
+        /** current state */
+        std::vector<double> x;
```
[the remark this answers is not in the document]
FFroehlich: yes and no ReturnData only contains (and probably must?) doubles as it interfaces with matlab.  It would be nice if we could at some point use` long double` or `single` aswell. This functions reads from ReturnData (via getx), but those functions are probably the best place to do perform the conversion anyways.

include/amici_model_ode.h
```diff
@@ -0,0 +1,318 @@
...
+         * @param k constants vector
+         * @param h heavyside vector
+         * @param xB Vector with the adjoint states
+         * @param w vector with helper variables
+         * @param dwdp derivative of w wrt p
+         * @return status flag indicating successful execution
+         **/
+        virtual int model_qBdot(realtype *qBdot, const int ip, const realtype t, const realtype *x, const realtype *p, const realtype *k, const realtype *h,
+                                const realtype *xB, const realtype *w, const realtype *dwdp) {
+            warnMsgIdAndTxt("AMICI:mex","Requested functionality is not supported as (%s) is not implemented for this model!",__func__);
```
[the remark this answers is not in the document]
FFroehlich: these functions are called from CVodes/IDSA which are probably not exception safe.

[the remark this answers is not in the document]
dweindl: I assumed they only operate on the Cvode/IdaCreate allocated memory block which will be taken care of.

include/amici_model.h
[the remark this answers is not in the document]
dweindl: Hm, then we should draw a clear line between `realtype` and `double` usage.

src/forwardproblem.cpp
```diff
@@ -493,146 +457,96 @@ void ForwardProblem::prepEventSensis(int ie, ReturnData *rdata,
...
      * @param[in] it index of data point @type int
-     * @param[in] udata pointer to the user data struct @type UserData
-     * @param[out] rdata pointer to the return data struct @type ReturnData
-     * @param[in] edata pointer to the experimental data struct @type ExpData
-     * @param[out] tdata pointer to the temporary data struct @type TempData
-     * @param[in] solver pointer to solver object @type Solver
-     * @param[in] model pointer to model specification object @type Model
      */
 
     if (model->nx > 0) {
```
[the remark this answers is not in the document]
FFroehlich: AmiSolve will not be called for nx==0 as far as I remember and calls in getDiagnosis will fail. Still better to check for this in getDiagnosis.

include/amici_model_ode.h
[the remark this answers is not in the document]
FFroehlich: might be, still as we now throw an recoverable error (`AMICI_RECOVERABLE_ERROR`), the solver will not automatically stop simulation, but attempt recovery by trying a smaller stepsize. This is functionality is only available with return codes.

include/amici_vector.h
```diff
@@ -0,0 +1,286 @@
...
+         * @return const pointer to data array
+         */
+        const realtype *data() const {
+            return vec.data();
+        }
+        
+        /** N_Vector accessor
+         * @return N_Vector
+         */
+        N_Vector getNVector() {
```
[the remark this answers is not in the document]
FFroehlich: `error: functions that differ only in their return type cannot be overloaded` ?

include/amici_model_ode.h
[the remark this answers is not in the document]
dweindl: Wait, how can "(%s) is not implemented for this model!" be recoverable?

include/amici_vector.h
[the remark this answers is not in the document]
dweindl: I meant changing this signature to const, not adding an additional function.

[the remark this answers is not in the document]
FFroehlich: currently there is `const N_Vector getNVector() const` and `N_Vector getNVector()`.

include/amici_model_ode.h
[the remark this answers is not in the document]
FFroehlich: It is not :D

include/amici_vector.h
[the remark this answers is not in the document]
dweindl: Ah, I see. But but it does not really make sense to have `const N_Vector getNVector() const` additionally, or does it? I guess `N_Vector getNVector() const` is enough.

dweindl:
R2016a/Linux, when running amiwrap: 
``` 
Error using mex
/home/dweindl/src/AMICI.github/src/amici.o: In function `amici::AmiException::storeBacktrace(int)':
amici.cpp:(.text._ZN5amici12AmiException14storeBacktraceEi[_ZN5amici12AmiException14storeBacktraceEi]+0x310): undefined reference to `dladdr'
collect2: error: ld returned 1 exit status
```

include/amici_vector.h
```diff
@@ -0,0 +1,286 @@
...
+            for (int idx = 0; idx < vaold.getLength(); idx++) {
+                nvec_array[idx] = vec_array.at(idx).nvec;
+            }
+        }
+        
+        /** accessor to data of AmiVector elements
+         * @param pos index of AmiVector
+         * @return pointer to data array
+         */
+        realtype *data(int pos) {
```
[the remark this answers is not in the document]
FFroehlich: `cannot initialize return object of type 'realtype *' (aka 'double *') with an rvalue
      of type 'const value_type *' (aka 'const double *')`

src/amici_solver_idas.cpp
```diff
@@ -408,111 +401,336 @@ void IDASolver::turnOffRootFinding() {
...
+                            model->plist[ip],N_VGetArrayPointer(dx),N_VGetArrayPointer(sx[ip]),N_VGetArrayPointer(sdx[ip]),
+                            model->w.data(),model->dwdx.data(),model->J->data,model->M.data(),&model->dxdotdp.at(ip*model->nx)) != AMICI_SUCCESS)
+                return AMICI_ERROR;
+            if(checkVals(model->nx,N_VGetArrayPointer(sxdot[ip]),"sensitivity rhs") != AMICI_SUCCESS)
+                return AMICI_ERROR;
+        }
+        return AMICI_SUCCESS;
+    }
                                             
 IDASolver::~IDASolver() { AMIFree(); }
```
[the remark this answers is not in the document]
FFroehlich: done

include/udata.h
```diff
@@ -14,11 +15,22 @@ void serialize(Archive &ar, amici::UserData &u, const unsigned int version);
 }}
 
 namespace amici {
+    
+class Model;
+
+template <class mxArray>
+UserData *userDataFromMatlabCall(const mxArray *prhs[], int nrhs);
```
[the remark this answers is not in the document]
FFroehlich: currently necessary to set this as friend function

src/amici_hdf5.cpp
FFroehlich: currently necessary to set this as friend function

src/amici_hdf5.cpp
FFroehlich: currently not possible as nx,nk,np need to be available for the creation of userData, which is used for the instantiation of the model. My current approach would be to create a model-specific ReturnData class that instantiates UserData and Model. We could then give a model specific name to getModelDims and reenable the use of multiple models, but I think this should not also be adressed in this monstrosity of a pull request

src/amici_interface_matlab.cpp
FFroehlich: see above

src/amici_hdf5.cpp
FFroehlich: I don't think so

src/amici_interface_matlab.cpp
FFroehlich: Now they are.

include/amici_vector.h
```diff
@@ -0,0 +1,286 @@
...
+            vec = rvec;
+            nvec = N_VMake_Serial(rvec.size(),rvec.data());
+        };
+        
+        /** swap functions for copy-and-swap idiom, swaps data between two AmiVectors
+         * see https://stackoverflow.com/questions/3279543/what-is-the-copy-and-swap-idiom
+         * @param first will be switched with second
+         * @param second will be switched with first
+         */
+        friend void swap( AmiVector& first, AmiVector& second ) {
```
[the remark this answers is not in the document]
FFroehlich: https://stackoverflow.com/questions/5695548/public-friend-swap-member-function

FFroehlich:
okay everything requested in the reviews should be adressed now, @dweindl can you have another look?

src/amici_hdf5.cpp
dweindl: I don't see the benefit of a model-specific ReturnData class. A ReturnData class that instantiates UserData and Model feels rather unnatural. Why not the other way around? This would involve only adding a convenience function something like Model::getReturnData() { return ReturnData(this) }

My main point is that getModelDims is only defined at link time, which makes it rather hard to follow program flow and to debug stuff. Simply passing Model down there is much clearing the getting it out of nowhere.

include/amici_model.h
```diff
@@ -1,5 +1,5 @@
-#ifndef AMICI_MODEL_H
-#define AMICI_MODEL_H
+#ifndef AMICI_fH
```
dweindl: Was this on purpose?

include/forwardproblem.h
```diff
@@ -43,36 +42,110 @@ class ForwardProblem {
...
+    AmiVectorArray *getsxptr() {
+        return &sx;
+    }
+    
+    /** accessor for pointer to sdx
+     * @return &sdx
+     */
+    AmiVectorArray *getsdxptr() {
+        return &sdx;
+    }
```
dweindl: @ all above: camel-case might be more readable.

include/symbolic_functions.h
```diff
@@ -3,30 +3,30 @@
...
-double am_Dspline(int id, double t, int num, ...);
-double am_Dspline_pos(int id, double t, int num, ...);
-double am_DDspline(int id1, int id2, double t, int num, ...);
-double am_DDspline_pos(int id1, int id2, double t, int num, ...);
+double spline(double t, int num, ...);
+double spline_pos(double t, int num, ...);
+double Dspline(int id, double t, int num, ...);
+double Dspline_pos(int id, double t, int num, ...);
+double DDspline(int id1, int id2, double t, int num, ...);
+double DDspline_pos(int id1, int id2, double t, int num, ...);
```
dweindl: All above: lower-case initial would be more conventional

src/amici_hdf5.cpp
FFroehlich: Currently the model only has a transient state and I would rather keep it that way. Still I agree that extending the results class might not be the best idea.  Maybe call it a simulator/kernel object that encapsulates all the other currently existing classes.

Still, currently instantiating the model requires instantiating udata which requires a call to getModelDims, so we cannot pass the model when getModelDims is called.

dweindl: Instantiating the model does not require udata.

Model::Model(const int nx, const int nxtrue,
              const int ny, const int nytrue, const int nz, const int nztrue,
              const int ne, const int nJ, const int nw, const int ndwdx,
              const int ndwdp, const int nnz, const int ubw, const int lbw,
              const AMICI_o2mode o2mode, const std::vector<realtype> p,
              const std::vector<realtype> k, const std::vector<int> plist,
              const std::vector<realtype> idlist, const std::vector<int> z2event)

FFroehlich: const std::vector p,
const std::vector k,
const std::vector plist
all come from udata, see respective wrapfunctions.cpp. If we do not know the size of plist at that point, we could aswell skip the initialization altogether as the size of many members depends on it.

dweindl: I see your point. Still I argue that AMICI core files should not depend on any functions that are only defined in model-specific files. 
In this function it can be replaced by reading those dimensions from the data file later on.
Maybe we should discuss (and document) the intended responsibilities and dependencies of udata, model, rdata soonish.

include/symbolic_functions.h
FFroehlich: true, but the regexp expression in gccode.m would be more of a pain

src/amici_hdf5.cpp
FFroehlich: agreed

PR #232, opened 2017-11-22
"Fixes" MSVS + forced recompilation on changes to x/p/k/root

FFroehlich:


src/amici_interface_matlab.cpp
```diff
@@ -449,6 +465,16 @@ ExpData *expDataFromMatlabCall(const mxArray *prhs[], const UserData *udata,
...
 }
+    
+    /** conversion from double to int with rounding,
+     * we know that values should be close to integer values so rounding can be performed by adding 0.5
+     * and performing a static cast subsequently
+     *  @param x input
+     *  @return int_x rounded and casted value
+     */
+    int round(const double x){
+        return(static_cast<int>(x+0.5));
```
dweindl: Is this function really necessary? Using `static_cast<int>(x)` should do the trick. I would not encourage passing something like 0.75 for an enum value...

dweindl: Or directly changing the respective `amioptions` field to `int` is probably more elegant.

FFroehlich: can we be sure that when MATLAB stores 2 as double it stores it at 2.000000.... and not as 1.9999999...?

dweindl: I would really really hope so, since 2 is not the result of any computation and it is well representable as float.

FFroehlich: hmm then lets check double(int(x)))-x=0 and throw an error if not?

dweindl: Would also work. Although handling it directly as `int` in `amioptions` would be nicer. There, the user can directly see what happened to his input.

FFroehlich: true ... however, implementing this in amioption is a bit nasty as MATLAB does not perform implicit conversion from double to int32, thus we would need to implement get/set functions for all int32 members to perform this implicit conversion, not something I am looking forward to :/

dweindl: :flushed:

FFroehlich: also explicit type specification for properties is an undocumented feature in matlab https://undocumentedmatlab.com/blog/setting-class-property-types and mtocpp does not support it (I have already filed a bug report over a year ago).

FFroehlich: ah, there is now a documented way to do it which also automatically typecasts but crashes matlab for versions prior to 2016a when loading the class  🎉  https://undocumentedmatlab.com/blog/setting-class-property-types-2

dweindl: Okay, okay, you convinced me :D

PR #243, opened 2018-01-10
Feature more adjoint options

paulstapor:
Added some options for the backward integration and fixed a little bug when calling Amici with forward sensitivities but without observables


src/amici_solver.cpp
```diff
@@ -148,18 +148,22 @@ void Solver::setupAMIB(BackwardProblem *bwd, const UserData *udata, Model *model
     AMISetUserDataB(bwd->getwhich(), model);
 
     /* Number of maximal internal steps */
-    AMISetMaxNumStepsB(bwd->getwhich(), 100 * udata->maxsteps);
+    AMISetMaxNumStepsB(tdata->which, udata->maxstepsB);
```
FFroehlich: tdata doesn't live here anymore, use `bwd->getwhich()`

Issue #245, opened 2018-01-11
Compilation failure on windows due to missing execinfo.h


dweindl:
Occurs in https://github.com/ICB-DCM/AMICI/blob/5cb96b75b4c9ba3405165e756b7a7c03d046d4bf/include/amici_exception.h#L10
This header is not available by default on windows.

HoangLUT:
Hello Dweindl,
What is that error really? I am using GitHub desktop on macOS and clone this repository and run 'installAMICI.m'. I test some models and it works just fine.
I did the same thing on my Windows PC and got that error. How should we fix it now? I am not sure why only Windows PC get that issue.
Thanks for your time.

FFroehlich:
I will try to fix this issue on Monday.

PR #246, opened 2018-01-15
fixes windows

FFroehlich:


FFroehlich:
see #245

include/amici_exception.h
```diff
@@ -7,9 +7,20 @@
 #include <cstdio>
 #include <cstring>
 #include <sstream>
-#include <execinfo.h>
-#include <dlfcn.h>    // for dladdr
-#include <cxxabi.h>   // for __cxa_demangle
+#define PLATFORM_WINDOWS 1
+#define PLATFORM_UNIX 2
+#if defined(_WIN32)
+    #define PLATFORM PLATFORM_WINDOWS // Windows
```
dweindl: Do we need the extra `PLATFORM`? Why not 
```
#if defined(_WIN32)
    #define PLATFORM_WINDOWS
#endif
...
```
Simplifies downstream usage to ` #ifdef PLATFORM_WINDOWS`

Issue #264, opened 2018-02-20
move second order code from matlab to c++ code


FFroehlich:
one of the few remaining things that are currently only accessible in matlab, but not from c/c++.

dweindl:
As soon as this is done, reenable checks for sensitivities in second-order models (testfunctions.cpp verifyReturnDataSensitivities)

paulstapor:
As I'm browsing through these issues right now...
I am wondering every now and then how to best re-implement second order stuff (even if someone else should do it at some point...)
I'm not perfectly convinced that the old matlab-style-way of augmenting the ODE model is really what we want... Also when doing second order, many things from first order can be reused, I think.

Maybe there's a way of doing this with a minimal wrapping overhead and with really most things done in C++ directly...
Any ideas? Obviously not urgent, just wanted to discuss it... (I am waiting for the compiler right now anyway... :D )

PR #265, opened 2018-02-22
Feature betterReturnData

FFroehlich:
matlab implementation now working


FFroehlich:
we need to think about how we gracefully transition the test as they are currently stored in ColMajor and results are now stored as RowMajor.
Also we are now mixing RowMajor and ColMajor dgemm calls, which should probably be avoided (eradicate last bits of RowMajor storage from generated function routines).
@dweindl what do you think?

dweindl:
Right, we will need some changes there. Maybe it would be good to port the current tests to matlab (keeping the same scripts for data generation, but running simulations and checking results in matlab). Shouldn't be too many lines of code. Could maybe be done on a separate, known-to-work branch.
This will ensure that everything is working while we rewire the C++ tests. For those tests, we will have to change HDF5 I/O only, right?

FFroehlich:
hmm i think it might be easier to simply change change hdf read-in (getDoubleArrayAttributeND) to check correctness, then change hdf5 test data generation and remove changes in read-in.

dweindl:
Easier, maybe. However, we might want to have some automated tests for matlab anyways. Running the examples works, but takes quite long and relies on the user knowing the correct graphs.

FFroehlich:
okay matlab tests are working now, next step would be to change hdf5 i/o.
I think we should also switch the hdf5 storage from RowMajor to ColMajor including recently introduced permutations. All transformation for matlab should be performed in matlab. We may want to take the opportunity to switch the amici hdf5 interface to cpp.

dweindl:
Actually, HDF5 should have things stored in row-major already.

src/amici.cpp
```diff
@@ -56,25 +56,46 @@ msgIdAndTxtFp warnMsgIdAndTxt = &printWarnMsgIdAndTxt;
...
+    try{
+        auto fwd = std::unique_ptr<ForwardProblem>(new ForwardProblem(rdata.get(),edata,&model,&solver));
+        fwd->workForwardProblem();
 
-    auto bwd = std::unique_ptr<BackwardProblem>(new BackwardProblem(fwd.get()));
-    bwd->workBackwardProblem();
+        auto bwd = std::unique_ptr<BackwardProblem>(new BackwardProblem(fwd.get()));
+        bwd->workBackwardProblem();
+    
+        rdata.get()->status = AMICI_SUCCESS;
```
dweindl: `rdata.get()->status` is equivalent to `rdata->status`. Thus, we should prefer the latter

src/amici.cpp
```diff
@@ -56,25 +56,46 @@ msgIdAndTxtFp warnMsgIdAndTxt = &printWarnMsgIdAndTxt;
...
+    } catch (amici::IntegrationFailure& ex) {
+        rdata.get()->invalidate(ex.time);
```
dweindl: I think we should move the exception handling one level up. Besides maybe `invalidate()`, the rest should be done in `mexFunction`, or otherwise left to the user. Currently the API is not very clear on where to check for exceptions and where to go after return codes.

src/amici_interface_cpp.cpp
```diff
@@ -37,19 +37,7 @@ namespace amici {
  * @return rdata pointer to return data object @type ReturnData
  */
 ReturnData *getSimulationResults(Model &model, const ExpData *edata, Solver &solver) {
```
dweindl: This one we can now get rid of completely, right?

src/amici.cpp
FFroehlich: agreed!

dweindl:
@ merge: i merged my recent changes with your here in #273. should merge that one first to avoid resolving conflicts twice

dweindl:
@FFroehlich edata fields are (at least according to comments) still in col-major. we will also change that as part of this pull request, right?

FFroehlich:
🤔  was planned so far, but probably makes sense.

dweindl:
CI result file in new format added.

testModels.m currently failing due to tolerances. How to handle?

TODO: edata col/row major

dweindl:
@FFroehlich Please check if the current tolerances in testModels.m make sense now. With the original ones, tests fail on my computer.

dweindl:
sigmay indexing here is still wrongly assuming col-major, right?  https://github.com/ICB-DCM/AMICI/blob/feature_betterReturnData/src/forwardproblem.cpp#L547 ff

FFroehlich:
okay @dweindl, from my side this is ready to be reviewed.

tests/testModels.m
```diff
@@ -83,10 +88,18 @@ function compareResults(sol,results)
...
-            if(ismember(ifield,{'s2llh', 's2x', 's2y', 's2z', 'sx'}))
-                checkAgreement(sol,results,ifield{1},1e-4,1e-3)
-            else
-                checkAgreement(sol,results,ifield{1})
+            if(~ismember(ifield,{'chi2'}))
+                if(~isempty(sol.s2llh))
+                    if(ismember(ifield,{'s2llh'}))
+                        checkAgreement(sol,results,ifield{1},1e-4,1e-3)
+                    elseif(ismember(ifield,{'x','sx','s2x','y','sy','s2y','z','sz','s2z','rz','srz','s2rz','sigmay','ssigmay','s2sigmay','sigmaz','ssigmaz','s2sigmaz','x0','sx0'}))
+                       % TODO: reimplement this as soon as
```
dweindl: ... as ...?

src/amici.cpp
```diff
@@ -80,19 +80,19 @@ std::unique_ptr<ReturnData> runAmiciSimulation(Solver &solver, const ExpData *ed
     } catch (amici::IntegrationFailure& ex) {
         rdata->invalidate(ex.time);
         rdata->status = ex.error_code;
-        amici::errMsgIdAndTxt("AMICI:mex:simulation","AMICI simulation failed at t = %f:\n%s\nError occured in:\n%s",ex.time,ex.what(),ex.getBacktrace());
+        amici::warnMsgIdAndTxt("AMICI:mex:simulation","AMICI simulation failed at t = %f:\n%s\nError occured in:\n%s",ex.time,ex.what(),ex.getBacktrace());
```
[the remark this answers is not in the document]
FFroehlich: warnings allows us to still return rdata, which is nice because you can actually inspect the simulation results to get some idea where things went wrong.

tests/testModels.m
FFroehlich: #264 is done :D

FFroehlich:
@dweindl we can delete the functions not hit in https://coveralls.io/builds/15938463/source?filename=src%2Famici_hdf5.cpp right?

dweindl:
@FFroehlich  Some at least. I'll take care of that.

dweindl:
@FFroehlich `testModels.m` fails at `model_neuron` for `sx`: `max(abs((expected - actual) ./ (rtol + abs(expected)))) = 1.2879e-05`

dweindl:
@FFroehlich In testModels.m there is (now) a problem with `pbar`: In C++ it is of length `nplist`, for the tests it is written of `nplist`, but in matlab simulate_* it is assumed to be of length `np`. 
How to fix? Is it going to be used somewhere soon?

PR #268, opened 2018-03-01
Using mostly the HDF5 C++ API. All HDF5 I/O should now be exception-s…

### LINKED ISSUE #254: Makeover for HDF5 I/O
Several functions there are currently not exception-safe. When fixing this, HDF5 calls should be changed to use the C++ API.

dweindl:
…afe. Closes #254.

### DIFF
--- a/include/amici_hdf5.h
+++ b/include/amici_hdf5.h
@@ -2,11 +2,20 @@
 #define AMICI_HDF5_H
 
 #include <string>
+#include <memory>
+#include <vector>
 
-#include <hdf5.h>
-#include <hdf5_hl.h>
+#include <H5Cpp.h>
 
 #undef AMI_HDF5_H_DEBUG
+
+#define AMICI_H5_SAVE_ERROR_HANDLER                                            \
+    herr_t (*old_func)(void *);                                                \
+    void *old_client_data;                                                     \
+    H5Eget_auto1(&old_func, &old_client_data);                                 \
+    H5Eset_auto1(NULL, NULL)
+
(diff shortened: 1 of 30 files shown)


dweindl:
@FFroehlich Works now with master. However, the error checking is not very strict, since our matlab-written data can be somewhat messy. For some functions we need to decide which fields are mandatory and which ones are optional. Otherwise some tests might silently be skipped. 
One issue is for example how to deal with empty attributes (when writing matlab `[]`). Should this be considered valid input?

FFroehlich:
hdf5 fields should only be skipped if the result is empty. If the testing routine cannot read a specific field, it should check whether the respective field in rdata is empty and throw an error otherwise.

include/amici_hdf5.h
```diff
@@ -2,28 +2,49 @@
...
 class Model;
 class Solver;
 
-/* Functions for reading and writing AMICI data to/from HDF5 files. */
+namespace hdf5 {
+
+
+/** @file Functions for reading and writing AMICI data to/from HDF5 files. */
+
 // TODO: proper type checking, exception instead of return code, c++ api; check if all fields are read and saved
```
FFroehlich: this is no longer TBD right?

FFroehlich: at least partially

dweindl: Partially. Let's leave it until we changed the rest and adapted to new storage format.

FFroehlich:
but lets adress the skipping of fields in #265. some of the "mandatory" fields are not written and NaN values/empty is not handled consistently.

dweindl:
Agreed.

PR #271, opened 2018-03-09
Write configuration for test cases in py, write results in C++

dweindl:
- Write configuration for test cases in py, writing results in C++. This allows to have fields with proper datatypes and dimensions and to have data written in row-major (-> #265).
- Array attributes changed to datasets due to size restriction as discussed.

~~Waiting for #264 : test for `sx` currently failing~~


FFroehlich:
can we ignore problems related to #264 for the time being and implement this in #265 ? implementing this and then rewriting everything shortly after is moot.

dweindl:
No problem. Let's just keep in mind that `sx` is not checked at all.

FFroehlich:
decrease of coverage is because of legacy code that will be removed later on right?

dweindl:
No, coveralls failed fully running the tests. Fixed now.

PR #273, opened 2018-03-10
Adapt tests and IO to new ReturnData and data layout

dweindl:


src/rdata.cpp
```diff
@@ -136,39 +136,36 @@ void ReturnData::applyChainRuleFactorToSimulationResults(const Model *model) {
...
         case AMICI_SCALING_LOG10:
-            coefficient = log(10.0);
-            for (int ip = 0; ip < nplist; ++ip)
-                pcoefficient.at(ip) = unscaledParameters[model->plist(ip)] * log(10);
+            coefficient.at(ip) = log(10.0);
+            pcoefficient.at(ip) = unscaledParameters.at(model->plist(ip)) * log(10);
             if (sensi == AMICI_SENSI_ORDER_SECOND && o2mode == AMICI_O2MODE_FULL)
-                for (int ip = 0; ip < np; ++ip)
-                    augcoefficient.at(ip) = unscaledParameters.at(ip) * log(10);
+                // TODO @REVIEW: unscaledParameters.at(model->plist(ip)) or unscaledParameters.at(ip)
```
FFroehlich: `unscaledParameters.at(ip)`, here we always have the default order set at compile time (as we need to encode this in the states), which is not affected by the states.
We should actually change this, but for now the previuous implementation was correct.

include/rdata.h
```diff
@@ -134,6 +136,7 @@ class ReturnData {
     std::vector<int> numnonlinsolvconvfailsB;
 
     /** employed order forward problem (dimension: nt) */
+    // TODO @Review: should be int?
```
FFroehlich: yep, should be int.

PR #283, opened 2018-03-14
Feature 259,  Generate more informative output when detected Inf/NaN

dweindl:


include/amici/model.h
```diff
@@ -763,6 +763,16 @@ namespace amici {
         
         const realtype gett(const int it, const ReturnData *rdata) const;
 
+        /**
```
FFroehlich: we dont need double doxygen documentation

dweindl: Why double? The one below is a different function.

PR #288, opened 2018-03-14
Add new tests for sensitivities with empty sens_ind / plist

dweindl:


tests/testModels.m
```diff
@@ -23,6 +26,10 @@ function testModels()
             model_rtol = 1e-5;
         end
         for itest = 1:length(info.Groups(imodel).Groups)
+            if(any(strcmp(ignoredTests, info.Groups(imodel).Groups(itest).Name)))
```
FFroehlich: `ismember(info.Groups(imodel).Groups(itest).Name, ignoredTests)` would be more consistent with the rest of the script.

PR #297, opened 2018-04-06
Add ExpData copy constructor and others

dweindl:


FFroehlich:
I would expect that `ExpData::ExpData(const ExpData &other)` also copies the contents and not only the dimensions. Did I miss something here?

src/edata.cpp
```diff
@@ -9,20 +9,36 @@ namespace amici {
...
-    sigmay.resize(model.nt() * model.nytrue);
-    mz.resize(model.nMaxEvent() * model.nztrue);
-    sigmaz.resize(model.nMaxEvent() * model.nztrue);
+    : ExpData(model.nytrue, model.nztrue, model.nt(), model.nMaxEvent())
+{
+}
+
+ExpData::ExpData(const ExpData &other)
+    : ExpData(other.nytrue, other.nztrue, other.nt, other.nmaxevent,
+              other.my, other.sigmay, other.mz, other.sigmaz)
```
dweindl: @FFroehlich Here is the content.

FFroehlich: ah, I am not sure what I was looking for :D

Issue #249, opened 2018-01-21
amidata - new fields


thomassligon:
I am running pre-equilibrium with the data format defined by amidata.

The objective function contains one pair of simulations for each experiment. 
The first simulation starts with some proteins set to zero initial concentration, which grows and then levels off due to transcription, translation, and degradation. This simulation runs until a steady state is reached. It requires conditions such as "gene Lyn on" or "gene Lyn off". 

The second simulation represents the time course of the experiment. It requires the same conditions as simulation one, but other conditions representing the experimental conditions, such as "allergen present".

This means that I require two conditions for each experiment, one for each simulation.

It would be nice to have another condition, such as data.condition2, or data.conditionSteadyState.

I am also using a field data.comments. Since it is not used in the code, it is OK today, but it might be nice to have it available in the code for usage in plotting.

FFroehlich:
I see the appeal of having a second condition for the steady state. However, things become more complicated if the same steady state is used for multiple different conditions. In the long run, this issue will probably be adressed (in more generality) as soon as we implement SED-ML support, which has been planned for quite a while.

PR #302, opened 2018-05-15
Feature249 preequilibration condition

dweindl:
Will implement #249

### DIFF
--- a/src/forwardproblem.cpp
+++ b/src/forwardproblem.cpp
@@ -99,9 +99,21 @@ void ForwardProblem::workForwardProblem() {
 
     /* if preequilibration is necessary, start Newton solver */
     if (solver->getNewtonPreequilibration()) {
+        // Are there dedicated preequilibration parameters provided?
+        if(edata->fixedParametersPreequilibration.size()) {
+            if(edata->fixedParametersPreequilibration.size() != (unsigned) model->nk())
+                throw AmiException("Number of fixed parameters (%d) in model does not match preequilibration parameters in ExpData (%zd).",
+                                   model->nk(), edata->fixedParametersPreequilibration.size());
+            model->setFixedParameters(edata->fixedParametersPreequilibration);
+        } else if(edata->fixedParameters.size()) {
+            if(edata->fixedParameters.size() != (unsigned) model->nk())
+                throw AmiException("Number of fixed parameters (%d) in model does not match ExpData (%zd).",
+                                   model->nk(), edata->fixedParameters.size());
+            model->setFixedParameters(edata->fixedParameters);
+        }
+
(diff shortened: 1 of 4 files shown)


src/forwardproblem.cpp
```diff
@@ -86,22 +86,34 @@ void ForwardProblem::workForwardProblem() {
...
     }
     
     int ncheck = 0; /* the number of (internal) checkpoints stored so far */
     realtype tlastroot = 0; /* storage for last found root */
 
     /* if preequilibration is necessary, start Newton solver */
     if (solver->getNewtonPreequilibration()) {
+        // Are there dedicated preequilibration parameters provided?
+        if(edata && edata->fixedParametersPreequilibration.size()) {
+            warnMsgIdAndTxt("asdf", "sefsdf");
```
FFroehlich: i guess we can remove those warning messages now?

src/forwardproblem.cpp
```diff
@@ -113,6 +125,15 @@ void ForwardProblem::workForwardProblem() {
...
     }
 
+    if(edata && edata->fixedParameters.size()) {
+        // fixed parameter in model are superseded by those provided in edata
+        // Note: this changes those parameters of `model` permanently
+        if(edata->fixedParameters.size() != (unsigned) model->nk())
+            throw AmiException("Number of fixed parameters (%d) in model does not match ExpData (%zd).",
+                               model->nk(), edata->fixedParameters.size());
+        model->setFixedParameters(edata->fixedParameters);
+    }
```
FFroehlich: I think if `edata->fixedParameters.size()` is not provided here, we probably want to revert back to the model specified parameters that we may have overwritten with `edata->fixedParameters`.

src/forwardproblem.cpp
dweindl: Sorry, yes, should not have gotten in here.

src/forwardproblem.cpp
dweindl: You are right.

dweindl: (Overwritten by `edata->fixedParametersPreequilibration` you mean?)

FFroehlich: yes

src/forwardproblem.cpp
```diff
@@ -129,6 +131,8 @@ void ForwardProblem::workForwardProblem() {
             throw AmiException("Number of fixed parameters (%d) in model does not match ExpData (%zd).",
                                model->nk(), edata->fixedParameters.size());
         model->setFixedParameters(edata->fixedParameters);
+    } else if (edata && solver->getNewtonPreequilibration() && originalFixedParameters.size()) {
+        model->setFixedParameters(originalFixedParameters);
```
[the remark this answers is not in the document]
FFroehlich: yes

PR #306, opened 2018-05-22
various bugfixes for longstanding issues

FFroehlich:


LoosC:
hmm I still get the error: use of undeclared identifier 'Dmax'; did you mean 'amici::Dmax' 
dydx[1+0*27] = Dmax(1,x[0],1.0E1,0.0);
in my example model using Matlab2016b.

Issue #303, opened 2018-05-15
SBML import fails with MATLAB R2018a


thomassligon:
The attached file was created in Cell Designer and validated by SBML.org, but ODE import fails with

```
Error using sym>convertChar (line 1448)
Character vectors and strings in the first argument can only specify a variable or number. To evaluate character vectors and strings representing symbolic
expressions, use 'str2sym'.

Error in sym>tomupad (line 1214)
        S = convertChar(x);

Error in sym (line 211)
                S.s = tomupad(x);

Error in cell2sym (line 31)
Csym = cellfun(conv, C, 'UniformOutput', false);

Error in sym>tomupad (line 1226)
    xsym = cell2sym(x);

Error in sym (line 211)
                S.s = tomupad(x);

Error in SBMLode/importSBML>cleanedsym (line 686)
csym = sym(sanitizeString(strrep(str,'time','__time_internal_amici__')));

Error in SBMLode/importSBML (line 80)
all_rulemath = cleanedsym({model.rule.formula});

Error in SBMLode (line 63)
            model.importSBML(filename);

Error in testAsthmaM07cd2>import (line 57)
    ODE = SBMLode([modelName '.xml']);

Error in testAsthmaM07cd2 (line 10)
ODE = import(modelName); % only first time
```
[ODEimportError.zip](https://github.com/ICB-DCM/AMICI/files/2005694/ODEimportError.zip)

dweindl:
@thomassligon The script fails because of a wrong filename (ASTHMA_V40_M07cd vs ASTHMA_V40_M07cd2). 

After correcting that, sbml import works. Cannot reproduce this error wit the latest AMICI version on R2017b/Linux.

thomassligon:
Daniel, thanks for the quick response! I see that I uploaded the wrong script, cd2 instead of cd, so that (trivial) problem is fixed. In addition, if ODE import and amiwrap works, then everything is OK. However, I just retested, and I still get the same error. I am running R2018a/Windows. Could you try with R2018a/Linux?

thomassligon:
I just retested on R2017b/Windows and it seems to be working OK, except that I am getting tons of this warning:
```
Warning: Support of character vectors will be removed in a future release. Character vectors can be used only for variable names and numbers. Instead, to
create symbolic expressions first create symbolic variables using 'syms'. To evaluate character vectors and strings representing symbolic expressions, use
'str2sym'. 
```
Maybe R2018a is the future release they were talking about.

[...]

Issue #234, opened 2017-11-30
Warning: Support of character vectors will be removed in a future release.


thomassligon:
When I import SBML, I get this warning 208 times:
```
Warning: Support of character vectors will be removed in a future release. Character vectors can be used only for variable names and numbers. Instead,
to create symbolic expressions first create symbolic variables using 'syms'. To evaluate character vectors and strings representing symbolic
expressions, use 'str2sym'. 
> In sym>convertExpression (line 1581)
  In sym>convertChar (line 1486)
  In sym>tomupad (line 1236)
  In sym (line 215)
  In cell2sym (line 31)
  In sym>tomupad (line 1248)
  In sym (line 215)
  In SBMLode/importSBML>cleanedsym (line 686)
  In SBMLode/importSBML (line 208)
  In SBMLode (line 63)
  In testAMICI (line 6) 
```
This makes it difficult to see the real messages.
The solution is easy, and is contained in the warning itself.
Just replace importSBML line 686
`csym = sym(sanitizeString(strrep(str,'time','__time_internal_amici__')));`
by
`csym = str2sym(sanitizeString(strrep(str,'time','__time_internal_amici__')));`

FFroehlich:
This is a known issue: https://github.com/ICB-DCM/AMICI/issues/50

The last time I checked, the warning did not recommend usage of `str2sym`. I would thus assume that this functionality is not backwards compatible and we probably need to perform version checking.

Still thanks for pointing this out :)

thomassligon:
Yes, I checked the MATLAB documentation. str2sym was introduced in R2017b.

dweindl:
In R2018a this no longer raises a warning but SBML-import fails, see #303.

PR #308, opened 2018-05-23
fix issue #303

thomassligon:
More text will appear in pull request.

### DIFF
--- a/matlab/SBMLimporter/@SBMLode/importSBML.m
+++ b/matlab/SBMLimporter/@SBMLode/importSBML.m
@@ -101,11 +101,7 @@ this.state = species_sym;
 nx = length(this.state);
 
 % extract corresponding volumes
-if verLessThan('matlab','R2017b')
-    compartments = sym(sanitizeString({model.species.compartment}));
-else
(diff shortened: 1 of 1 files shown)


thomassligon:
Here is an explanation of the changes (all in module importSBML.m).
In lines 104, 452, 459, and 701 you will find a version check (verLessThan) and a line where sym has been replaced by str2sym. That change was documented in issues #234 and #303 and is due to a change in MATLAB.
In lines 259 and 267 you will find a change of reactants{:} to sym(reactants) and similar for products. This is because the first case passes a cell array to the cat function, and the second passes a symbolic array. The cell array invokes a special feature of MATLAB that converts it to many separate parameters, as documented in
https://www.mathworks.com/help/matlab/matlab_prog/comma-separated-lists.html
I assume that they symbolic array is what was intended. Please check to see if that is true, and please retest in R2018a on your machine.

FFroehlich:
please use recently introduced the function `betterSym` instead of implementing the version check at every occurence of sym https://github.com/ICB-DCM/AMICI/blob/master/matlab/auxiliary/betterSym.m

dweindl:
Closing after months of inactivity.

PR #312, opened 2018-05-25
fixes for issues 121, 190, 284 and 296

FFroehlich:


include/amici/defines.h
```diff
@@ -1,8 +1,18 @@
...
+/** MS definition of PI and other constants */
+#define _USE_MATH_DEFINES
+#ifndef M_PI
+/** define PI if we still have no definition */
+#define M_PI 3.14159265358979323846
+#endif
 
 namespace amici {
 
+constexpr double pi = M_PI;
```
dweindl: Can we merge that with the lines above and conditionally set `amici::pi` without introducing `M_PI` in the public header? (And consequently only use `amici::pi`)

matlab/@amimodel/generateC.m
```diff
@@ -113,11 +124,8 @@ function generateC(this)
 fprintf(fid,'\n');
 fprintf(fid,'\n');
 
-fprintf(fid,'#define pi M_PI\n');
-fprintf(fid,'\n');
 fprintf(fid,'#ifdef __cplusplus\n#define EXTERNC extern "C"\n#else\n#define EXTERNC\n#endif\n');
```
dweindl: Not strictly part of this PR, but we should remove that. Does not make sense since everything is C++ now.

PR #327, opened 2018-05-31
Miscellaneous

dweindl:
More robust packaging, various smaller extensions fixes and cleanup of Python API.


dweindl:
@FFroehlich I think there is something wrong with the sensitivities. Can you please have a look at the sensitivities at the bottom of: https://github.com/ICB-DCM/AMICI/blob/dw_misc/python/examples/example_steadystate/ExampleSteadystate.ipynb
Is dwdp missing in dydp in python sbml import?

PR #330, opened 2018-06-09
feature(python) Allow sigma parameters

dweindl:
- Allow sigma parameters in python
- Fix rdata.ssigma indexing error


dweindl:
NOTE: 
- I think tests are failing due to wrong expected results. For model_jakstat_adjoint saved sllh w.r.t. sigma parameters p[14..16] are 0.0 and probably wrong.
- steadystate example notebook now seems to have correct sensitivities (one single sigma for one single observable)
- python gradient check indicates that for model_jakstat_adjoint sllh w.r.t. sigma parameters p[14..16] is still wrong (3 sigma parameters)

dweindl:
sllh is now in line with finite differences for all examples

@FFroehlich Please have a look at the other arrays indexing substitutions with num2str in https://github.com/ICB-DCM/AMICI/blob/dw_misc/matlab/@amifun/gccode.m#L145 , this looks somewhat incoherent. Seem to be okay, but might be more coherent to change dsigmazdp as well, and maybe also they remaining ones.

include/amici/forwardproblem.h
```diff
@@ -168,77 +169,75 @@ class ForwardProblem {
...
-    /** state derivative of data likelihood */
+    /** state derivative of data likelihood (dimension nJ x nx x nt, ordering =?) */
     std::vector<realtype> dJydx;
-    /** state derivative of event likelihood */
+    /** state derivative of event likelihood (dimension nJ x nx x nMaxEvent, ordering =?) */
     std::vector<realtype> dJzdx;
     
     /** data likelihood */
-    std::vector<realtype> Jy;
+    std::vector<realtype> Jy; // TODO remove?
```
FFroehlich: I think this can be safely removed

include/amici/model.h
```diff
@@ -1229,68 +1047,72 @@ namespace amici {
         
         const realtype *getsrz(const int nroots, const int nplist, const ReturnData *rdata) const;
 
-        /** Jacobian */
+
+
+        /** Sparse Jacobian (dimension: nx x nx x nnz)*/
```
FFroehlich: dimension is only nnz

include/amici/forwardproblem.h
```diff
@@ -168,77 +169,75 @@ class ForwardProblem {
 
     void applyEventSensiBolusFSA();
     
-    /** array of index which root has been found */
+    /** array of index which root has been found  (dimension: ne * ne * ne * nmaxevent, ordering = ?) */
```
FFroehlich: dimension should be ne * ne * nmaxevent (also change init).

matlab/@amifun/gccode.m
```diff
@@ -143,7 +143,7 @@
             cstr = regexprep(cstr,'var_sigma_y_([0-9]+)','sigmay[$1]');
             cstr = regexprep(cstr,'var_sigma_z_([0-9]+)','sigmaz[$1]');
             cstr = regexprep(cstr,'var_dsigma_zdp_([0-9]+)',['dsigmazdp[ip*' num2str(model.nz) ' + $1]']);
```
FFroehlich: dsigmazdp should work the same way as `dsigmaydp`

src/model_header.ODE_template.h
```diff
@@ -30,6 +30,7 @@ extern void dwdx_TPL_MODELNAME(realtype *dwdx, const realtype t, const realtype
 extern void dxdotdp_TPL_MODELNAME(realtype *dxdotdp, const realtype t, const realtype *x, const realtype *p, const realtype *k, const realtype *h, const int ip, const realtype *w, const realtype *dwdp);
 extern void dydx_TPL_MODELNAME(double *dydx, const realtype t, const realtype *x, const realtype *p, const realtype *k, const realtype *h);
 extern void dydp_TPL_MODELNAME(double *dydp, const realtype t, const realtype *x, const realtype *p, const realtype *k, const realtype *h, const int ip);
+extern void dsigmaydp_TPL_MODELNAME(double *dsigmaydp, const realtype t, const realtype *p, const realtype *k, const int ip);
```
FFroehlich: dsigma_ydp

src/model.cpp
```diff
@@ -588,6 +880,7 @@ void Model::fsigma_z(const realtype t, const int ie, const int *nroots,
 void Model::fdsigma_zdp(const realtype t) {
     std::fill(dsigmazdp.begin(),dsigmazdp.end(),0.0);
     for(int ip = 0; (unsigned)ip < plist_.size(); ip++)
+        // TODO dsigmazdp.data() or slice?
```
FFroehlich: fdsigma_zdp code should be roughly the equivalent code as for fdsigma_ydp (that might currently not have been the case), the only thing that changes is that for individual ne, we have to check z2event for which iz we may write values.

src/forwardproblem.cpp
```diff
@@ -615,41 +574,37 @@ void ForwardProblem::getDataSensisFSA(int it) {
...
         }
     }
 
-    for (int iy = 0; iy < model->nytrue; iy++) {
-        if (edata) {
+    if(edata) {
+        for (int iy = 0; iy < model->nytrue; iy++) {
             if (isNaN(edata->sigmay[it * rdata->nytrue + iy])) {
-                model->fdsigma_ydp(it, rdata);
+                // @ REVIEW: it seems redundant to call this for every iy, should be only called once per it, not?
```
FFroehlich: yes, also looks redundant to me, actually, most of this code block seems redundant with the current model->fdsigma_ydp implementation. I wanted to move the respective code to the model class anyways as its more consistent with the other code to have the responsability there.

include/amici/forwardproblem.h
dweindl: Just to be sure, just Jy or all 4 of them?

FFroehlich: all 4 of them

src/forwardproblem.cpp
dweindl: Postponed to #332

PR #334, opened 2018-06-12
fix for adjoints with preequilibration and sensitivities

### LINKED ISSUE #333: Incorrect sensitivities when using preequilibration with adjoint sensitivities
When using preequilibration with adjoint sensitivities `sx0` is (silently) not set. This leads to wrong `sllh`.

paulstapor:
fix #333


paulstapor:
That should already do the job... so let's not wait until Friday, if this is enough

src/steadystateproblem.cpp
```diff
@@ -70,8 +70,7 @@ void SteadystateProblem::workSteadyStateProblem(ReturnData *rdata,
     getNewtonOutput(rdata, model, newton_status, run_time, it);
 
     /* Compute steady state sensitvities */
-    if (rdata->sensi_meth == AMICI_SENSI_FSA &&
-        rdata->sensi >= AMICI_SENSI_ORDER_FIRST)
+    if (rdata->sensi >= AMICI_SENSI_ORDER_FIRST)
```
dweindl: && rdata->sensimeth != AMICI_SENSI_NONE

paulstapor:
Just for the moment, I added the other solver which can also be used for preequilibration and changed the naming in the way we discussed...
I hope things are ready for merging now...

PR #337, opened 2018-06-12
fix(python) Fix python model compilation, add to ci, improve hdf5 search, python ci in venv

dweindl:


dweindl:
@FFroehlich Now I remember why I named the function dsigmaydp instead of dsigma_ydp: Python interface expects the variable to have the same name as the function. However the variable was so far (inconsistently) called dsigmaydp. How to address this?

- Change matlab code generation to variable name dsigma_ydp?
or
- Have Python generate function dsigmaydp?

FFroehlich:
I think the functions + variables should be called `dsigmaydp` to be consistent with all other functions/code. A long time ago there was some conflict with other variables, but that should no longer be an issue (sigma_y was the symbolic function and sigmay was mix of symbolic + values in ExpData, but we don't want to keep those seperate anyways).

dweindl:
Okay, that means we remove all underscores in `sigma_*` in all interfaces?

FFroehlich:
Yes.

dweindl:
In the matlab part, I will only change that for C++ code generation. Otherwise we'd have to adapt syms files as well, which would break quite some things.

dweindl:
@FFroehlich Please add your macOS hdf5 include and library directories to https://github.com/ICB-DCM/AMICI/blob/fix_declaration/python/sdist/setup.py#L60

PR #345, opened 2018-06-22
ci(python) Install amici package to virtual environment and run tests there (Closes #342)

### LINKED ISSUE #342: [python, ci] Install AMICI package in virtual environment
Running `scripts/buildAll.sh` builds AMICI, creates the python package and installs it. This will overwrite any previous installation of AMICI and might not always be intended.

I would prefer having the CI scripts create a virtual python environment within the build folder and run all tests there. Running it in a controlled environment will also ensure that all dependencies of the package are listed properly.

dweindl:
#342


scripts/run-python-tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Test python model wrapping inside virtual environment
+
+AMICI_PATH="`dirname \"$0\"`"
+AMICI_PATH="`( cd \"$AMICI_PATH/..\" && pwd )`"
+
+cd ${AMICI_PATH}/tests
+
+${AMICI_PATH}/build/venv/bin/pip3 install scipy
```
FFroehlich: `source /build/venv/bin/activate`?

dweindl: There was an issue with `activate` that made me use above. I don't recall the details anymore.

FFroehlich:
why is coverage decreasing here? I would guess that cpp coverage files from python tests changed location?

dweindl:
Will merge as part of #337. Closing.

PR #346, opened 2018-06-22
Add symbol names to python-wrapped models and make them available in python (Closes #64)

### LINKED ISSUE #64: Include symbol names in simulation results
I would find it very helpful to have the names of the model species from the model_*_syms file included in the simulation results.

It would also be nice to be able to get a list of the symbol names for the fixed and variable parameters of the compiled model. Not sure where to put them, though.

dweindl:


include/amici/model.h
```diff
@@ -504,6 +504,32 @@ namespace amici {
          */
         int checkFinite(const int N,const realtype *array, const char* fun) const;
 
+
+        /**
+         * @brief Get names of the model parameters
+         * @return the names
+         */
+        virtual std::vector<std::string> getParameterNames() const { return std::vector<std::string>(); }
```
FFroehlich: can we have smarter defaults here? Maybe something like `unnamedParameter1,unnamedParameter2,...`

dweindl: Technically yes. I thought this way it is easier to check if names are set or not. Creating default names by the user is trivial. I find it unlikely that the user really wants to use any of the default names we could come up with here.

FFroehlich: Fair enough, but I think we should then at least print a warning/error message or even throw an exception.

dweindl: Hmm... I think I'd prefer: 
```
auto names = model.getParameterNames();
if(names.empty()) {
// ... 
}
```
over
```
try {
auto names = model.getParameterNames();
} catch (AmiException &e) {
// ... 
}
```

FFroehlich: You actually need to do:
```
auto names = model.getParameterNames();
if(names.empty() && model.getNumberOfParameters()>0) {
// ... 
}
```
The fact that the result for zero length parameters and no set Names is the same bothers me. Maybe introduce a function `isSetParameterNames()` that checks `if(names.empty() && model.getNumberOfParameters()>0)`?

dweindl: Okay. I think I'd prefer that. I think `hasParameterNames()` would also be coherent with libsbml. We can also reopen that discussion after having used it in practice a bit more.

FFroehlich: sounds good!

Issue #332, opened 2018-06-12
[core] Cleanup ForwardProblem (esp. getDataSensisFSA)


dweindl:
`ForwardProblem::getDataSensisFSA` looks somewhat inefficient and redundant with `Model::fdsigma_ydp`. This should be cleanup up. Ideally while clarifying responsibilities of `Model` and `ForwardProblem`. For example, it is at the moment not clear who is responsible for populating ReturnData. I would prefer keeping that out of `Model`.

dweindl:
Collecting other core cleanup TODOs here as well:

`AmiVector *BackwardProblem::get*ptr` return non-const pointers to private members. This should be changed. They are used by `Solver::setupAMIB`. The respective lines there would better be kept inside  `Model` making those getters unnecessary.

PR #376, opened 2018-07-27
Feature 223 Implement res, sres and FIM for better interoperability with least squares solvers

FFroehlich:


include/amici/model.h
```diff
@@ -456,6 +456,35 @@ namespace amici {
...
+         * @param it time index
+         * @param rdata ReturnData instance to which result will be written
+         * @param edata ExpData instance containing observable data
+         */
+        void fres(const int it, ReturnData *rdata, const ExpData *edata);
+        
+        /** chi-squared function
+         * @param it time index
+         * @param rdata ReturnData instance to which result will be written
+         * @param edata ExpData instance containing observable data
```
dweindl: No `edata` there

include/amici/model.h
```diff
@@ -456,6 +456,35 @@ namespace amici {
...
+         */
+        void fsres(const int it, ReturnData *rdata, const ExpData *edata);
+        
+        /** fisher information matrix function
+         * @param it time index
+         * @param rdata ReturnData instance to which result will be written
+         * @param edata ExpData instance containing observable data
```
dweindl: No edata there

include/amici/rdata.h
```diff
@@ -101,12 +101,16 @@ class ReturnData {
      * nplist x ny, row-major) */
     std::vector<realtype> ssigmay;
     
-    /** observable (dimension: nt x ny, row-major) */
+    /** observable (dimension: nt*ny) */
```
dweindl: still row-major?

include/amici/rdata.h
```diff
@@ -101,12 +101,16 @@ class ReturnData {
...
     std::vector<realtype> res;
 
-    /** parameter derivative of residual (dimension: nt x ny x nplist,
+    /** parameter derivative of residual (dimension: nt*ny x nplist,
```
dweindl: Why * instead x?

src/model.cpp
```diff
@@ -1052,37 +1052,88 @@ void Model::fdJrzdsigma(const int nroots,const ReturnData *rdata,
...
+ * @param x Vector with the states
+ */
 void Model::fdwdx(const realtype t, const N_Vector x) {
     fw(t,x);
     std::fill(dwdx.begin(),dwdx.end(),0.0);
     fdwdx(dwdx.data(),t,N_VGetArrayPointer(x), unscaledParameters.data(),fixedParameters.data(),h.data(),w.data());
 }
 
+void Model::fres(const int it, ReturnData *rdata, const ExpData *edata) {
+    if (edata && !rdata->res.empty()) {
```
dweindl:
```
if (!edata || rdata->res.empty()) {
   return 
}
```
Saves us a level of indentation and increases readability :)

src/model.cpp
```diff
@@ -1052,37 +1052,88 @@ void Model::fdJrzdsigma(const int nroots,const ReturnData *rdata,
...
+            int iyt = iy + it * rdata->ny;
+            if (isNaN(edata->my.at(iyt_true)))
+                continue;
+            rdata->res.at(iyt_true) = (rdata->y.at(iyt) - edata->my.at(iyt_true))/rdata->sigmay.at(iyt);
+        }
+    }
+}
+    
+void Model::fchi2(const int it, ReturnData *rdata) {
+    if (!rdata->res.empty()) {
```
dweindl: see previous

src/model.cpp
```diff
@@ -1052,37 +1052,88 @@ void Model::fdJrzdsigma(const int nroots,const ReturnData *rdata,
...
+        for (int iy = 0; iy < nytrue; ++iy) {
+            int iyt_true = iy + it * rdata->nytrue;
+            rdata->chi2 += pow(rdata->res.at(iyt_true), 2);
+        }
+    }
+}
+    
+void Model::fsres(const int it, ReturnData *rdata, const ExpData *edata) {
+    if (edata && !rdata->sres.empty()) {
```
dweindl: see previous

src/model.cpp
```diff
@@ -1052,37 +1052,88 @@ void Model::fdJrzdsigma(const int nroots,const ReturnData *rdata,
...
+            for (int ip = 0; ip < nplist(); ++ip) {
+                rdata->sres.at(iyt_true + ip * rdata->nt * rdata->nytrue) =
+                rdata->sy.at(iy + rdata->ny*(ip + it*nplist()))/rdata->sigmay.at(iyt);
+            }
+        }
+    }
+}
+    
+void Model::fFIM(const int it, ReturnData *rdata) {
+    if (!rdata->sres.empty()) {
```
dweindl: see previous :)

include/amici/model.h
```diff
@@ -456,6 +456,35 @@ namespace amici {
...
+         */
+        void fFIM(const int it, ReturnData *rdata);
```
dweindl: Regarding all newly added functions:
None of them requires access to `Model`, or did I miss something? Wouldn't it be better to move them to `ReturnData` or completely free them?

At least all entities not depending on `ExpData` could also be computed on demand to save memory.

tests/testModels.py
```diff
@@ -129,6 +129,24 @@ def grad(x0, symbol='llh', x0full=None, plist=[], verbose=False):
         plist = [ip]
         err_norm = check_grad(func, grad, p[plist], 'llh', p, [ip])
         print('sllh: p[%d]: |error|_2: %f' % (ip, err_norm))
+
+    rdata = amici.runAmiciSimulation(model, solver, edata)
+
+
+    if solver.getSensitivityMethod() == amici.AMICI_SENSI_FSA:
```
dweindl: leastsquares_applicable = solver.getSensitivityMethod() == amici.AMICI_SENSI_FSA

src/rdata.cpp
```diff
@@ -77,6 +75,9 @@ ReturnData::ReturnData(Solver const& solver, const Model *model)
     if (sensi >= AMICI_SENSI_ORDER_FIRST){
         sllh.resize(nplist, getNaN());
         sx0.resize(nx * nplist, getNaN());
+        FIM.resize(nplist * nplist, 0.0);
+        res.resize(nt * model->nytrue, 0.0);
+        sres.resize(nt * model->nytrue * nplist, 0.0);
```
dweindl: Depends on `sy`, shouldn't that be moved inside `if (sensi_meth == AMICI_SENSI_FSA ...` below?

include/amici/rdata.h
FFroehlich: its 1D ...

include/amici/rdata.h
FFroehlich: its 1D ...

include/amici/rdata.h
FFroehlich: because its actually a 2D matrix.

include/amici/model.h
FFroehlich: those that do not directly depend on ExpData indirectly depend on it either through rdata.res or rdata.sres. Yes, this could in theory be moved to ReturnData, but conceptually I think that these computation belong more in the model than ReturnData. In particular if we look at #332, these functions may easily end up with calls to Model->fsy().

include/amici/rdata.h
dweindl: If it is nt*ny it still need to be clear if e.g. the second element corresponds to (t0,y1) or (t1,y0), not?

include/amici/rdata.h
dweindl: See above. Maybe I don't get it, but for me it doesn't seem to matter whether one calls this 1D, flattened 2D, or flattened 3D. The order of the elements should still be defined somehow.

include/amici/model.h
dweindl: Hm... okay... `Model` is already so bloated. Maybe that can be cleaned up at a later stage...

python/sdist/amici.egg-info/PKG-INFO
```diff
@@ -0,0 +1,37 @@
+Metadata-Version: 1.2
```
dweindl: I guess the egg-info were accidentally added?

python/sdist/amici/amici.py
```diff
@@ -0,0 +1,3283 @@
+# This file was automatically generated by SWIG (http://www.swig.org).
```
dweindl: Also this one

PR #390, opened 2018-08-05
added check for ny > 0, due to simulation failure

paulstapor:
Currently, the routines fy, fJy, etc. are called in forward simulation, regardless if model->ny > 0 or not. This leads to simulation failure, if model->ny == 0.
This commit should fix the problem.


FFroehlich:
please move the dimension checks to the individual functions in `Model` that fail. We dont want to have `ForwardProblem` to take care of that

PR #392, opened 2018-08-06
Fix matlab interface

paulstapor:
Due to the change of type of observedData in edata, the Matlab interface had to be adapted.

### DIFF
--- a/src/interface_matlab.cpp
+++ b/src/interface_matlab.cpp
@@ -238,7 +238,7 @@ ExpData *expDataFromMatlabCall(const mxArray *prhs[],
         
         mxArray *dataSigmaYT;
         mexCallMATLAB(1, &dataSigmaYT, 1, &dataSigmaY, "transpose");
-        std::vector<realtype> observedDataSigma = mxArrayToVector(dataSigmaYT, nt_sigmay * nt_sigmay);
+        std::vector<realtype> observedDataSigma = mxArrayToVector(dataSigmaYT, ny_sigmay * nt_sigmay);
         edata->setObservedDataStdDev(observedDataSigma);
     } else {
         throw AmiException("Field Sigma_Y not specified as field in data struct!");
--- a/src/interface_matlab.cpp
+++ b/src/interface_matlab.cpp
@@ -175,7 +175,7 @@ std::vector<realtype> mxArrayToVector(mxArray *mexData, int dataLength){
  * dimension checks
  * @return edata pointer to experimental data object
  */
-ExpData *expDataFromMatlabCall(const mxArray *prhs[],
+std::unique_ptr<ExpData> expDataFromMatlabCall(const mxArray *prhs[],
                                Model const &model) {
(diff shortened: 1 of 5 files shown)


src/interface_matlab.cpp
```diff
@@ -158,6 +158,15 @@ void amici_daxpy(int n, double alpha, const double *x, const int incx, double *y
     FORTRAN_WRAPPER(daxpy)(&n_, &alpha, x, &incx_, y, &incy_);
 }
 
+/** conversion from mxArray to vector<realtype>
+ *  @param x input
+ *  @return int_x casted value
+ */
+std::vector<realtype> mxArrayToVector(mxArray *mexData, int dataLength){
+    std::vector<realtype> dataVector(dataLength);
```
FFroehlich: I'm fine with this, but @dweindl is a strong proponent of `auto`. Could you replace all `std::vector<realtype>` initializations with `auto`?

FFroehlich: also you might want to rebase as everything before d0cd3b7 is still from another PR.

paulstapor: I implemented the auto where it was possible and would do a squash commit anyway... In that case: also rebase?

src/interface_matlab.cpp
```diff
@@ -178,7 +187,11 @@ ExpData *expDataFromMatlabCall(const mxArray *prhs[],
...
     ExpData *edata = new ExpData(model);
-    if (edata->my.empty() && edata->mz.empty()) {
+    
+    /* Check whether data and events are empty */
+    std::vector<realtype> observedData = edata->getObservedData();
+    std::vector<realtype> observedEvents = edata->getObservedEvents();
+    if (observedData.empty() && observedEvents.empty()) {
         // if my and mz are both empty, no (or empty) data was provided
         // in that case we simply return a nullptr for easier checking.
         delete(edata);
```
dweindl: @FFroehlich: I don't get what is supposed to happen in these lines.  Do we still need that?

FFroehlich: We should be able to completely remove this as data fields are no longer mandatory.

paulstapor:
Can/Shall we merge now?

FFroehlich:
Rebase is still missing. See https://www.atlassian.com/git/tutorials/merging-vs-rebasing for a step-by-step guide how to do that. For a list of interactive rebase commands see https://help.github.com/articles/about-git-rebase/.
What you want to do is drop commits db0a035 to d0cd3b7. In the future, start a PR from a clean branch.

dweindl:
> Can/Shall we merge now?

I think now we are good. Feel free to drop some of the less informative commit messages from the squash message.

paulstapor:
I tried to rebase but didn't pull the correct master and messed up some stuff. I resolved the conflicts which occurred, so everything is on the state we want it to have.
I'm sorry that the history will not be 100% clean and will make it better next time, but fixing it now would simply take me longer than it is worth it in this case, I think...

PR #399, opened 2018-08-10
Maintenance fixes

FFroehlich:
Still want to fix a few python/sbml related issues


include/amici/defines.h
```diff
@@ -91,7 +91,7 @@ typedef enum AMICI_sensi_meth_TAG {
 } AMICI_sensi_meth;
 
 /** linear solvers for CVODES/IDAS */
-enum LinearSolver {
+enum class LinearSolver {
```
dweindl: I like using enum class.

FFroehlich: I was not sure, but we should also be able to replace the typedef enums right?

src/solver_cvodes.cpp
```diff
@@ -446,7 +452,42 @@ void CVodeSolver::turnOffRootFinding() {
...
+const Model *CVodeSolver::getModel() const {
+    if (!solverMemory)
+        throw AmiException("Solver has not been allocated, information is not available");
+    auto cv_mem = (CVodeMem) solverMemory.get();
+    return static_cast<Model *>(cv_mem->cv_user_data);
+}
+
+bool CVodeSolver::getMallocDone() const {
+    if (!solverMemory)
+        throw AmiException("Solver has not been allocated, information is not available");
```
dweindl: Should this throw? Would this not just be `false`? (Same in getAdjMallocDone)

tests/cpputest/testfunctions.cpp
```diff
@@ -49,6 +49,8 @@ void simulateVerifyWrite(const std::string& hdffileOptions, const std::string& h
 
     // simulate & verify
     auto rdata = runAmiciSimulation(*solver, edata.get(), *model);
+    
+    auto rdata_resimulation = runAmiciSimulation(*solver, edata.get(), *model);
```
dweindl: Document why we want this? Otherwise this is likely to be optimized out ;)

include/amici/defines.h
dweindl: Yes, they can also be replaced. That's from back in C time.
Since the will be always specified with the class qualifier, we can also rid of any common prefixes.
By convention, enum class member names are not capitalized. Just as comment, we can keep the existing ones the way they are.

FFroehlich:
The remaining issue is due to a bug in SWIG that was fixed no less than 4 years ago (https://github.com/swig/swig/issues/147), we need swig3.0.3 (see https://github.com/SGpp/SGpp/issues/1), but apparently travis downloads 3.0.2 although 3.0.8 is the latest version for xenial (https://packages.ubuntu.com/xenial/amd64/swig3.0/download). @dweindl any idea how to fix this?

dweindl:
The problem is that semi-supported xenial image. Apt is installing software from the trusty and trusty-backports repository. There, the newest one is indeed only [3.0.2](https://packages.ubuntu.com/trusty-backports/swig3.0).

Should be possible to install the xenial package anyways. Will need `sudo: true` then and maybe change back to manual apt invocation.
Another option would be dodging the official images completely and use a custom docker image instead with an up-to-date ubuntu. Not sure if it takes 15min or a day to get this working, though.

dweindl:
Works ;)

PR #410, opened 2018-08-17
Feature preequilibration sensitivities via FSA

FFroehlich:


src/edata.cpp
```diff
@@ -94,6 +94,11 @@ ExpData::ExpData(ReturnData const& rdata, std::vector<realtype> sigma_y, std::ve
 }
     
 void ExpData::setTimepoints(const std::vector<realtype> &ts) {
+    if (ts.size() > 2)
```
dweindl: `if(!std::is_sorted(ts.begin(), ts.end())`

src/steadystateproblem.cpp
```diff
@@ -303,87 +316,54 @@ void SteadystateProblem::getSteadystateSimulation(ReturnData *rdata, Solver *sol
         if (it_newton >= solver->getMaxSteps())
             throw NewtonFailure(AMICI_TOO_MUCH_WORK,"getSteadystateSimulation");
     }
+    if (it<1) {
+        if (newtonSimSolver->getSensitivityOrder()>SensitivityOrder::none)
```
dweindl:
```
if (newtonSimSolver->getSensitivityOrder() > SensitivityOrder::none)  {
  if(it < 1) ...
else ...
} 
```

FFroehlich: nope `newtonSimSolver->getSensitivityOrder()` and `dolver->getSensitivityOrder()` may differ

FFroehlich: but probably can combine two with && and use elseif

tests/example_steadystate.py
```diff
@@ -232,6 +269,42 @@ def writeSensiAdjSimPreeq(filename):
...
+    ex.data['condition'] = ex.modelOptions['kappa']
+    ex.data['conditionPreequilibration'] = [0.1, 0.4, 0.7, 0.5]
+    ex.data['t'] = np.linspace(0, 5, 10)
+
+    ex.modelOptions['ts'] = np.linspace(0, 5, 10)
+    ex.modelOptions['FSASteadyStateSensitivityFlag'] = 1
+    ex.solverOptions['sensi'] = 1
+    ex.solverOptions['sensi_meth'] = 2
+    ex.solverOptions['newton_preeq'] = True
+    ex.solverOptions['atol'] = 10 ** -16
```
dweindl: Are they saved as float or int? Int might cause problems when reading. 1e-16 would for sure be float.

src/steadystateproblem.cpp
dweindl: Ah, true. Missed that.

tests/example_steadystate.py
FFroehlich: I am not even sure where the generated options are used ...

dweindl: Should be okay, I guess... https://github.com/ICB-DCM/AMICI/blob/1551d4c68910fe4b636d9da0facbdf13b39f77a9/tests/example.py#L10

dweindl: Sorry, no, might a problem. https://github.com/ICB-DCM/AMICI/blob/1551d4c68910fe4b636d9da0facbdf13b39f77a9/tests/example.py#L22

src/steadystateproblem.cpp
```diff
@@ -239,26 +246,29 @@ void SteadystateProblem::getSteadystateSimulation(ReturnData *rdata, Solver *sol
...
-        newton_sim = createSteadystateSimSolver(solver, model, *t);
-        model->fx0(x);
+        newtonSimSolver = createSteadystateSimSolver(solver, model, *t);
     } else {
         /* Carry on simulating from last point */
         *t = model->t(it-1);
         model->fx0(x);
+        model->fsx0(sx, x);
         /* Reinitialize old solver */
         solver->reInit(*t, x, &dx);
```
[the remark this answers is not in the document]
FFroehlich: Actually I was wondering why `fx0` was called here in the first place, as we are continuing the simulation calling these here does not make a whole lot of sense to me.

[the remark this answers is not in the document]
paulstapor: The Newton solver rewrite x. For the simulation in preequlibration case, we want to start however from x0, not from the last iterate of the Newton solver...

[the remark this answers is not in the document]
FFroehlich: Yes, but preequilibration is only done in the case `if (it<1)` right?
These calls happen in `else`, i.e. when the user specified `inf` as a timepoint. I think there we actually probably want to continue from the last simulated timepoint and not from the initial conditions (also we ignore model->getInitialStates here)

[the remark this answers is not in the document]
paulstapor: You are right. These two lines shouldn't be here...

PR #411, opened 2018-08-18
fix #408 questionmark operator

### LINKED ISSUE #408: Edata constructor: Question mark operator wrong
Question mark operator is `condition ? true_case : false_case` in C++ (unlike python)...
In: 
https://github.com/ICB-DCM/AMICI/blob/master/src/edata.cpp#L79
Same some lines below.

yannikschaelte:


FFroehlich:
its the "ternary operator" but apparently I only know how to call it, not how to use it ...

yannikschaelte:
anything to be done about the codecov failings?

FFroehlich:
write tests ;)
for the moment we do not require codecov or codacy to pass, so you can simply ignore them and merge.

PR #418, opened 2018-08-29
Feature convenience wrappers, fixes #413

### LINKED ISSUE #413: [python] pandas support for edata and rdata
Would make life a lot easier if edata and rdata could be exported to pandas DataFrame with overlapping column headers.

FFroehlich:


python/amici/sbml_import.py
```diff
@@ -1486,28 +1486,45 @@ class TemplateAmici(Template):
...
+                                filter_by_name=False):
     """Turn assignment rules into observables.
 
     Arguments:
         sbml_model: an sbml Model instance
-        filter_function: callback function taking assignment variable as input and returning True/False to indicate if the respective rule should be turned into an observable
+        filter_function: callback function taking assignment variable as input
+        and returning True/False to indicate if the respective rule should be
+        turned into an observable
+        filter_by_name: boolean indicating whether assignment variables are
```
dweindl: We can also just pass the respective sbml parameter object and let the user call getId() or getName() or anything else directly... Breaking change, but barely used yet anyways.

include/amici/model.h
```diff
@@ -374,6 +381,39 @@ namespace amici {
          */
         void setFixedParameters(std::vector<realtype> const&k);
 
+        /**
+         * @brief Get value of model parameter with the specified Id
+         * @param par_id parameter id
+         * @return parameter value
+         */
+        realtype getFixedParameterById(std::string par_id) const;
```
dweindl: `std::string const& par_id `?

src/model.cpp
```diff
@@ -372,6 +372,30 @@ void Model::setParameterScale(std::vector<ParameterScaling> const& pscale) {
...
     return originalParameters;
 }
+    
+realtype Model::getParameterById(std::string par_id) const {
+    if(!hasParameterIds())
+        throw AmiException("Could not access parameters by id as they are not set");
+    
+    auto par_ids = getParameterIds();
+    for(int ip = 0; ip < np(); ++ip)
+        if(par_ids.at(ip) == par_id)
```
dweindl: please use `std::find(par_ids.begin(), par_ids.end(), par_id)`

FFroehlich:
I don't like the fact that reinit_fixed_parameter_initial_conditions has to be specified at compile time. However, this is currently necessary as we cannot ensure correct results for models where individual initial conditions depend on the fixedParameters and Parameters.
I thought about making specifiable at runtime as a private member of `Model` and making the setter function overridable to allow replacing it with an error throwing function.

Regarding the presimulation testcase, I am still not sure what the best way of generating a c++ test case is. Generate static code that is included in the repository by default? Otherwise we would probably have to bootstrap the build process. As a compiled amici would be necessary for the generation of model files.

python/amici/sbml_import.py
```diff
@@ -1108,12 +1181,20 @@ def writeFunctionFile(self, function, assume_pow_positivity):
...
         body = self.getFunctionBody(function)
-        if assume_pow_positivity:
-            body = [re.sub(r'(^|\W)pow\(', r'\1amici::pos_pow(', line) for line in body]
+        if assume_pow_positivity \
+                and 'assume_pow_positivity' in self.functions[function].keys()\
+                and self.functions[function]['assume_pow_positivity']:
+            body = [re.sub(r'(^|\W)pow\(', r'\1amici::pos_pow(', line) for line
+                    in body]
+            # execute this twice to catch cases where the ending ( would be the
+            # starting (^|\W) for the following match
```
dweindl: This will fail for 'pow(pow(pow(` etc.. Maybe directly go for:
```
while True:
        b = re.sub(r'(^|\W)pow\(', r'\1amici::pos_pow(', a)
        if b == a: break 
        else: a = b
```

src/forwardproblem.cpp
```diff
@@ -186,11 +190,58 @@ void ForwardProblem::handlePreequilibration()
...
+        for (int ip = 0; ip < model->nplist(); ip++)
+            for (int ix = 0; ix < model->nx; ix++)
+                rdata->sx0[ip * model->nx + ix] = sx.at(ix, ip);
+        
+        if(solver->getSensitivityMethod() == SensitivityMethod::forward)
+            solver->sensReInit(&sx, &sdx);
+    }
+}
+    
+void ForwardProblem::handlePresimulation(int *ncheck)
```
dweindl: Do we need both preequilibration and presimulation? Wouldn't preequilibration just be the special case of presimulation with t_presim=Inf?

src/model.cpp
```diff
@@ -372,6 +373,32 @@ void Model::setParameterScale(std::vector<ParameterScaling> const& pscale) {
 std::vector<realtype> const& Model::getParameters() const {
     return originalParameters;
 }
+    
+realtype Model::getParameterById(std::string const& par_id) const {
+    if(!hasParameterIds())
+        throw AmiException("Could not access parameters by id as they are not set");
+    
+    auto par_ids = getParameterIds();
```
dweindl: Just to note: With the current implementation, the parameter vector is copied for each getParameterId call. This might be significant when iterating over thousands of parameters.

Effort could already be halved, by reusing par_ids instead of (the current implementation of) hasParameterIds.

Same for the other related functions.

src/model.cpp
```diff
@@ -372,6 +373,32 @@ void Model::setParameterScale(std::vector<ParameterScaling> const& pscale) {
...
+    auto it = std::find(par_ids.begin(), par_ids.end(), par_id);
+    
+    if(it != par_ids.end())
+        return originalParameters.at(it - par_ids.begin());
+    else
+        throw AmiException("Could not find parameter with specified id");
```
dweindl: Save the else ;)
```
    if(it == par_ids.end())
        throw AmiException("Could not find parameter with specified id"); 
    return originalParameters.at(it - par_ids.begin());
```

dweindl: Could reduce redundancy: 
```
getValueById(getParmeterNames(), parameters, "parameter")
getValueById(getFixedParmeterNames(), fixedparameters, "fixed parameter")
....
```

src/model.cpp
```diff
@@ -380,6 +407,76 @@ void Model::setParameters(const std::vector<realtype> &p) {
...
+    
+    if(it != par_ids.end()) {
+        originalParameters.at(it - par_ids.begin()) = value;
+        unscaleParameters(this->unscaledParameters.data());
+    } else {
+        throw AmiException("Could not find parameter with specified id");
+    }
+}
+    
+void Model::setParametersByIdRegex(std::string const& par_id_regex, realtype value) {
```
dweindl: Would it be helpful to return the number of matches?

src/steadystateproblem.cpp
```diff
@@ -69,23 +71,10 @@ void SteadystateProblem::workSteadyStateProblem(ReturnData *rdata,
     /* Compute steady state sensitvities */
     
     if (solver->getSensitivityOrder() >= SensitivityOrder::first &&
-        solver->getSensitivityMethod() != SensitivityMethod::none) {
+        (newton_status == 1 || newton_status == 3 || model->getSteadyStateSensitivityMode() != SteadyStateSensitivityMode::simulationFSA))
```
dweindl: Let's use an enum for newton_status.

tests/testSBML.py
```diff
@@ -41,6 +80,69 @@ def runTest(self):
...
+        self.assertTrue(
+            np.isclose(
+                rdata[0]['y'],
+                df_obs[list(model.getObservableIds())].values
+            ).all()
+        )
+        df_res = amici.getResidualsAsDataFrame(model, edata, rdata)
+
+        model.getParameterById('p1')
+        model.setParameterById('p1',2.0)
```
dweindl: +`self.assertTrue(model.getParameterById('p1') == 2.0)` ... ?

.travis.yml
```diff
@@ -50,7 +50,7 @@ matrix:
 
 install:
     - export BASE_DIR=`pwd`
-    - pip3 install --user --upgrade pip setuptools wheel pkgconfig doxypypy scipy
+    - pip3 install --user --upgrade pip setuptools wheel pkgconfig doxypypy scipy pandas
```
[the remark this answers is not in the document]
dweindl: Ah, k. I think I would require it. Shouldn't really be a limitation for anybody. Otherwise the example notebook fails, which not too welcoming...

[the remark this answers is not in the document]
FFroehlich: Riiight ... Good Point!

include/amici/edata.h
```diff
@@ -317,6 +317,10 @@ class ExpData {
     /** condition-specific parameters for pre-equilibration of size Model::nk() or empty.
      * Overrides Solver::newton_preeq */
     std::vector<realtype> fixedParametersPreequilibration;
+    /** condition-specific parameters for pre-simulation of size Model::nk() or empty. */
+    std::vector<realtype> fixedParametersPresimulation;
+    /** duration of pre-simulation */
```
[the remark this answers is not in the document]
FFroehlich: presimualation is [t_start-t_presim,t_start], simulation is [t_start,...] will add doc.

FFroehlich:
>I don't like the fact that reinit_fixed_parameter_initial_conditions has to be specified at compile time. However, this is currently necessary as we cannot ensure correct results for models where individual initial conditions depend on the fixedParameters and Parameters.
I thought about making specifiable at runtime as a private member of Model and making the setter function overridable to allow replacing it with an error throwing function.

>Regarding the presimulation testcase, I am still not sure what the best way of generating a c++ test case is. Generate static code that is included in the repository by default? Otherwise we would probably have to bootstrap the build process. As a compiled amici would be necessary for the generation of model files.

@dweindl

dweindl:
>I don't like the fact that reinit_fixed_parameter_initial_conditions has to be specified at compile time. However, this is currently necessary as we cannot ensure correct results for models where individual initial conditions depend on the fixedParameters and Parameters.
I thought about making specifiable at runtime as a private member of Model and making the setter function overridable to allow replacing it with an error throwing function.

Introducing a flag / function into Model to indicate parameter-dependency of initial conditions sounds reasonable.

>Regarding the presimulation testcase, I am still not sure what the best way of generating a c++ test case is. Generate static code that is included in the repository by default? Otherwise we would probably have to bootstrap the build process. As a compiled amici would be necessary for the generation of model files.

- I'd prefer to keep tests at the lowest level possible without much overhead so they can be run often. Swigging and building the python model takes extra time. 
- For travis it would be good to generate model-c++ code there, to avoid testing with outdated files. So there the python module should be built beforehand. For the non-sbml models we'll have to keep the pregenerated code. Would be good to get rid of them over time though.
- During coding, it should be easily possible to regenerate models as well as recompiling only existing models and running the tests.

python/amici/sbml_import.py
FFroehlich: Don't you think that 2 passes will always be fine?

`re.sub(r'(^|\W)pow\(', r'\1amici::pos_pow(', 'pow(pow(pow(')` yields `'amici::pos_pow(pow(amici::pos_pow('`

`re.sub(r'(^|\W)pow\(', r'\1amici::pos_pow(',re.sub(r'(^|\W)pow\(', r'\1amici::pos_pow(', 'pow(pow(pow('))` yields `'amici::pos_pow(amici::pos_pow(amici::pos_pow('`

src/model.cpp
FFroehlich: I don't like the idea of duplicating code for the sake of a meager 2 fold increase in efficency. I would guess that the compiler anyway inlines `!hasParameterIds()` detects that `getParameterIds()` is called twice.

python/amici/sbml_import.py
dweindl: Oh right, it always matches every second and thus should work for any number. So it's fine.

src/model.cpp
dweindl: The duplication would be `par_ids.empty()`, I think that would be justifiable. ;) 
Not doing that won't kill us either.

src/forwardproblem.cpp
FFroehlich: We may want to to preequilibration AND presimulation at the same time. Long term this might be replaced by events that can also update fixed parameters, but nobody knows when that will be implemented ...

src/model.cpp
```diff
@@ -373,31 +373,56 @@ void Model::setParameterScale(std::vector<ParameterScaling> const& pscale) {
 std::vector<realtype> const& Model::getParameters() const {
     return originalParameters;
 }
+
+realtype getValueById(std::vector<std::string> ids, std::vector<realtype> values,
```
dweindl: Pass ids and values as const&

src/model.cpp
```diff
@@ -373,31 +373,56 @@ void Model::setParameterScale(std::vector<ParameterScaling> const& pscale) {
...
+                      std::string const& id, const char* value_name, const char* id_name) {
+    auto it = std::find(ids.begin(), ids.end(), id);
+    if(it != ids.end())
+        return values.at(it - ids.begin());
+    
+    throw AmiException("Could not find %s with specified %s", value_name, id_name);
+}
+    
+void setValueById(std::vector<std::string> ids, std::vector<realtype> values, realtype value,
```
dweindl: This changes only a copy of values (and discards it). Pass ids as `const&`, values as `&`

src/model.cpp
```diff
@@ -373,31 +373,56 @@ void Model::setParameterScale(std::vector<ParameterScaling> const& pscale) {
...
+                      std::string const& id, const char* value_name, const char* id_name) {
+    auto it = std::find(ids.begin(), ids.end(), id);
+    if(it != ids.end())
+        values.at(it - ids.begin()) = value;
+    else
+        throw AmiException("Could not find %s with specified %s", value_name, id_name);
+}
+    
+int setValueByIdRegex(std::vector<std::string> ids, std::vector<realtype> values, realtype value,
```
dweindl: See previous

FFroehlich:
Looks like travis can't recover from the switch from integration service to github app, I will close this and reopen a new PR.

PR #439, opened 2018-09-06
Improve packaging

dweindl:
To maybe fully work on windows...

### DIFF
--- a/python/amici/__init__.py
+++ b/python/amici/__init__.py
@@ -22,7 +22,7 @@ Attributes:
     amiciSrcPath: absolute path of the amici source directory
     amiciModulePath: absolute root path of the amici module
     hdf5_enabled: boolean indicating if amici was compiled with hdf5 support
-    has_clibs: boolean indicating if this is the full package with swig 
+    has_clibs: boolean indicating if this is the full package with swig
                interface or the raw package without
 """
 
--- a/python/amici/__init__.py
+++ b/python/amici/__init__.py
@@ -16,7 +16,6 @@ help(modelName)
 ```
 
 Attributes:
-    dirname: absolute path to parent directory of this file
     amici_path: absolute root path of the amici repository
     amiciSwigPath: absolute path of the amici swig directory
     amiciSrcPath: absolute path of the amici source directory
--- a/python/amici/__init__.py
+++ b/python/amici/__init__.py
@@ -22,22 +22,27 @@ Attributes:
     amiciSrcPath: absolute path of the amici source directory
     amiciModulePath: absolute root path of the amici module
     hdf5_enabled: boolean indicating if amici was compiled with hdf5 support
+    has_clibs: boolean indicating if this is the full package with swig 
+               interface or the raw package without
 """
 
 import os
 
-# If this file is inside the amici package, import swig interface,
-# otherwise we are inside the git repository, then don't
-dirname = os.path.dirname(__file__)
 hdf5_enabled = False
-if os.path.isfile(os.path.join(dirname, 'amici.py')):
+has_clibs = False
+
+try:
+    from . import amici
+    from .amici import *
+    hdf5_enabled = True
+    has_clibs = True
+except (ImportError, ModuleNotFoundError, AttributeError):
     try:
-        from . import amici
-        from .amici import *
-        hdf5_enabled = True
-    except AttributeError:
         from . import amici_without_hdf5 as amici
         from .amici_without_hdf5 import *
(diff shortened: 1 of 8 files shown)


python/amici/__init__.py
```diff
@@ -51,12 +55,14 @@
...
-    getDataObservablesAsDataFrame, getSimulationObservablesAsDataFrame, \
-    getSimulationStatesAsDataFrame, getResidualsAsDataFrame
+if has_clibs:
+    # these module require the swig interface
+    from .sbml_import import SbmlImporter, assignmentRules2observables, \
+        constantSpeciesToParameters
+    from .numpy import rdataToNumPyArrays, edataToNumPyArrays
+    from .pandas import getEdataFromDataFrame, \
+        getDataObservablesAsDataFrame, getSimulationObservablesAsDataFrame, \
+        getSimulationStatesAsDataFrame, getResidualsAsDataFrame
```
yannikschaelte: looks reasonable, but hope it won't get more cases than clibs and hdf5 :)

python/amici/sbml_import.py
```diff
@@ -1611,7 +1617,9 @@ def printWithException(self,math):
 
         """
         try:
-            return self.codeprinter.doprint(math)
+            ret = self.codeprinter.doprint(math)
+            ret = re.sub(r'(^|\W)M_PI(\W|$)', r'\1amici::pi\2', ret)
```
yannikschaelte: why are there both M_PI and amici::pi again?

python/amici/sbml_import.py
```diff
@@ -295,6 +295,7 @@ def sbml2amici(self,
                    sigmas=None,
                    verbose=False,
                    assume_pow_positivity=False,
+                   compiler=None
```
yannikschaelte: required for windows ('mingw32'), just wonder if amici can memorize this at compile time

python/amici/setup.template.py
```diff
@@ -25,22 +32,22 @@ def getAmiciLibs():
...
+h5pkgcfg = getHdf5Config()
+
+blaspkgcfg = getBlasConfig()
+linker_flags.extend(blaspkgcfg.get('extra_link_args', []))
+
+libraries = [*getAmiciLibs(), *blaspkgcfg['libraries']]
+if hdf5_enabled:
+    libraries.extend(['hdf5_hl_cpp', 'hdf5_hl', 'hdf5_cpp', 'hdf5'])
+
+sources = ['swig/TPL_MODELNAME.i', *getModelSources()]
```
yannikschaelte: honestly can't tell if all flags and options are set correctly, but looks reasonable

python/amici/setuptools.py
```diff
@@ -0,0 +1,134 @@
...
+    # try for hdf5 in standard locations
+    hdf5_include_dir_hints = ['/usr/include/hdf5/serial',
+                              '/usr/local/include',
+                              '/usr/include', # travis ubuntu xenial
+                              '/usr/local/Cellar/hdf5/1.10.2_1/include'] # travis macOS
+    hdf5_library_dir_hints = ['/usr/lib/x86_64-linux-gnu/', # travis ubuntu xenial
+                              '/usr/lib/x86_64-linux-gnu/hdf5/serial',
+                              '/usr/local/lib',
+                              '/usr/local/Cellar/hdf5/1.10.2_1/lib'] # travis macOS
+
```
yannikschaelte: standard for windows?

python/amici/sbml_import.py
dweindl: I did not find a way to convince CCodeprinter to directly use amici::pi. So here M_PI is replaced by amici::pi.

python/amici/sbml_import.py
dweindl: Could be convenient. Would have to create a config file in package directory. Feel free to open a separate issue on that.

python/amici/setuptools.py
dweindl: Not sure there is anything remotely standard on Windows. If you have something, we can add it..

yannikschaelte: Well, the Program Files (x86) folders, but we can see that whence hdf5 for windows works.

python/amici/__init__.py
dweindl: Yes, that's a bit messy. The problem is mostly, that for testing it is very helpful if you can import the package without requiring the compiled extensions. And the second problem is that there is no good way yet for having hdf5 available everywhere.

PR #446, opened 2018-09-17
Better tolerance control for the steady state problem

FFroehlich:


FFroehlich:
For the simulation based approach, I am wondering whether we should also do a convergence test on the sensitivities, i.e., also compute wrms_ip for all sx[ip] (i.e., replace call to fxdot with fsxdot(ip)) and ensure that also all of those are smaller than 1?

paulstapor:
> For the simulation based approach, I am wondering whether we should also do a convergence test on the sensitivities, i.e., also compute wrms_ip for all sx[ip] (i.e., replace call to fxdot with fsxdot(ip)) and ensure that also all of those are smaller than 1?

Sounds like a lot of overhead for big models, but it is actually necessary...
I would suggest to add a conditional test, which is only applied, if the steady state for the state variable ODE is reached. In the best case, the sensitivity check has to be performed only once then. In the worst case, it has to be repeated only for a few iterations...

PR #448, opened 2018-10-04
Feature nonnegativity

FFroehlich:


FFroehlich:
Initial evaluation suggest that this implementation has a pretty hefty negative impact on simulation time. We probably need to come up with a good testcase to do a more rigorous evaluation.

paulstapor:
The changes look actually all fine so far. I'm just wondering whether positivity checking is maybe superfluous during backward integration and quadrature, which might save some computation time...

Anyhow: things should work nicely, backward integration positivity checking would just affect the performance, but not yield any incorrect results. So the PR is fine for me...

FFroehlich:
> The changes look actually all fine so far. I'm just wondering whether positivity checking is maybe superfluous during backward integration and quadrature, which might save some computation time...
> 
> Anyhow: things should work nicely, backward integration positivity checking would just affect the performance, but not yield any incorrect results. So the PR is fine for me...

We create a copy of the state vector in `x_pos` and accordingly do not modify the solver state (which according to the cvodes documentation is something that we explicitely do not want to do)

According to prelimary testing, the checks themselves do not have too much impact on performance (<5%, the most severe being sxdot and qBdot). Still we might want to add a check that completely avoids additional computation if no entries in qpositivex are set.

include/amici/model.h
```diff
@@ -691,6 +691,19 @@ namespace amici {
          */
         void setTimepoints(std::vector<realtype> const& ts);
 
+        
+        /**
+         * @brief gets flags indicating whether states should be treated as non-negative
+         * @return vector of flags
+         */
+        std::vector<bool> getQPositiveX() const;
```
dweindl: What does the Q stand for?

src/model.cpp
```diff
@@ -548,6 +548,16 @@ void Model::setTimepoints(const std::vector<realtype> &ts) {
         throw AmiException("Encountered non-monotonic timepoints, please order timepoints such that they are monotonically increasing!");
     this->ts = std::move(ts);
 }
+    
+std::vector<bool> Model::getQPositiveX() const {
```
dweindl: return const ref?

src/model.cpp
```diff
@@ -1315,7 +1328,14 @@ bool operator ==(const Model &a, const Model &b)
             && (a.ts == b.ts)
             && (a.nmaxevent == b.nmaxevent)
             && (a.pscale == b.pscale)
+            && (a.qpositivex == b.qpositivex)
             && (a.tstart == b.tstart);
 }
 
+void Model::computeX_pos(N_Vector x) {
```
dweindl: To keep `Model` from exploding we could make that a free function. Also improves testability.

src/model.cpp
```diff
@@ -1315,7 +1328,14 @@ bool operator ==(const Model &a, const Model &b)
...
+    for (int ix = 0; ix < x_pos.getLength(); ++ix) {
+        x_pos.at(ix) = qpositivex.at(ix) && NV_Ith_S(x, ix) < 0 ? 0 : NV_Ith_S(x, ix);
```
dweindl: Add parentheses to not make me look up operator precedence again? :)

include/amici/model.h
```diff
@@ -1737,13 +1750,19 @@ namespace amici {
...
         /** sensitivity initialisation (size nx * nplist, ordering = ?) */
-        std::vector<realtype>sx0data;
+        std::vector<realtype> sx0data;
 
         /** timepoints (size nt) */
-        std::vector<realtype>ts;
+        std::vector<realtype> ts;
+        
+        /** vector of flags indicating whether state variables are to be assumed to be positive */
+        std::vector<bool> qpositivex;
```
dweindl: Is there a specific reason for having a bool array instead of an index array? The latter would avoid iterating over nx elements in case there are no non-negativity constraints (and safe space).

include/amici/model.h
FFroehlich: 😅 thats how it was called in d2d

src/model.cpp
FFroehlich: Ah, we want this to be a member function so we can preallocate `x_pos`.

include/amici/model.h
dweindl: :thinking: Maybe we can go for something like `nonNegativeStates` `nonNegativeStateIndices`?

src/model.cpp
dweindl: Can make it a free function and pass the preallocated array as second argument.

include/amici/model.h
FFroehlich: To copy `x`, we need to iterate over all elements anyways. Having an index array would require two loops. I think a cleaner solution would be to check once if any `qpositivex` is true and if not make computeX_pos not copy anything.

PR #450, opened 2018-10-09
Feature exception handling

FFroehlich:
Interfaces generated via swig can now properly handle exceptions :)

### DIFF
--- a/src/amici.cpp
+++ b/src/amici.cpp
@@ -103,15 +103,7 @@ std::unique_ptr<ReturnData> runAmiciSimulation(Solver &solver, const ExpData *ed
         rdata->invalidate(model.t0());
         rdata->status = AMICI_ERROR;
         amici::warnMsgIdAndTxt("AMICI:mex:simulation","AMICI simulation failed:\n%s\nError occured in:\n%s",ex.what(),ex.getBacktrace());
-    } catch (std::exception const& ex) {
-        rdata->invalidate(model.t0());
-        rdata->status = AMICI_ERROR;
-        amici::warnMsgIdAndTxt("AMICI:mex:simulation","AMICI simulation failed:\n%s",ex.what());
-    } catch (...) {
-        rdata->invalidate(model.t0());
-        rdata->status = AMICI_ERROR;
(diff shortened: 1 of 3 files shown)


src/amici.cpp
```diff
@@ -103,15 +103,7 @@ std::unique_ptr<ReturnData> runAmiciSimulation(Solver &solver, const ExpData *ed
...
-        rdata->invalidate(model.t0());
-        rdata->status = AMICI_ERROR;
-        amici::warnMsgIdAndTxt("AMICI:mex:simulation","AMICI simulation failed:\n%s",ex.what());
-    } catch (...) {
-        rdata->invalidate(model.t0());
-        rdata->status = AMICI_ERROR;
-        amici::warnMsgIdAndTxt("AMICI:mex", "Unknown internal error occured");
-    }
+    } 
     model.setFixedParameters(originalFixedParameters);
```
dweindl: Hmm..  this leaves the model in an unexpected state. I guess we should catch, revert to original values and rethrow?
Otherwise this might lead to some surprises if you reuse the model after catching.

src/amici.cpp
```diff
@@ -103,7 +103,15 @@ std::unique_ptr<ReturnData> runAmiciSimulation(Solver &solver, const ExpData *ed
...
+        model.setTimepoints(originalTimepoints);
+        throw ex;
+    } catch (...) {
+        model.setFixedParameters(originalFixedParameters);
+        model.setTimepoints(originalTimepoints);
+        throw std::runtime_error("Unknown internal error occured!");
+    }
     model.setFixedParameters(originalFixedParameters);
     model.setTimepoints(originalTimepoints);
     rdata->applyChainRuleFactorToSimulationResults(&model);
```
dweindl: We assume that if it is not an integration exception, there are no usable sensitivities at all? Otherwise we would have to include this line before rethrowing.

FFroehlich: If we throw and exception here, rdata will not be returned.

dweindl: Obviously... :man_facepalming:

PR #451, opened 2018-10-09
Feature better python support for STL vectors

FFroehlich:
this was simpler than expect ...


FFroehlich:
why is coverage dropping so heavily here 🤔

PR #460, opened 2018-11-10
Separate sbml import

FFroehlich:
Refactoring of sbml_import.py. ODE creation is now done in ode_export.py, which should make it easier to also import ODEs from other source. symbolic computations have been abstracted to simplify future extensions.

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -17,24 +17,24 @@ from string import Template
 
 from . import amiciSwigPath, amiciSrcPath, amiciModulePath
 
-'''
-prototype for generated C++ functions, keys are the names of functions 
+"""
+prototype for generated C++ functions, keys are the names of functions
 
 signature: str
-    defines the argument part of the function signature, input variables 
+    defines the argument part of the function signature, input variables
     should have a const flag
 
 assume_pow_positivity: bool
-    identifies the functions on which assume_pow_positivity will have an 
-    effect when specified during model generation. generally these are 
-    functions that are used for solving the ODE, where negative values may 
+    identifies the functions on which assume_pow_positivity will have an
+    effect when specified during model generation. generally these are
+    functions that are used for solving the ODE, where negative values may
     negatively affect convergence of the integration algorithm
 
 sparse: bool
-    specifies whether the result of this function will be stored in sparse 
-    format. sparse format means that the function will only return an array of 
+    specifies whether the result of this function will be stored in sparse
+    format. sparse format means that the function will only return an array of
     nonzero values and not a full matrix.
-'''
+"""
 functions = {
     'J': {
         'signature':
@@ -1301,28 +1301,28 @@ class ODEModel:
         """
         # compute total derivative according to chainrule
         # Dydz = dydx*dxdz + dydz
-        vars = dict()
-        vars['dydx'] = dict()
-        vars['dxdz'] = dict()
-        vars['dydz'] = dict()
+        variables = dict()
+        variables['dydx'] = dict()
+        variables['dxdz'] = dict()
+        variables['dydz'] = dict()
 
-        vars['dydx']['name'] = f'd{eq}d{chainvar}' \
+        variables['dydx']['name'] = f'd{eq}d{chainvar}' \
             if dydx_name is None else dydx_name
-        vars['dxdz']['name'] = f'd{chainvar}d{var}' \
+        variables['dxdz']['name'] = f'd{chainvar}d{var}' \
             if dxdz_name is None else dxdz_name
-        vars['dydz']['name'] = f'd{eq}d{var}'
+        variables['dydz']['name'] = f'd{eq}d{var}'
 
         # if the symbol appears in the function signature,
         # we can use the respective symbol instead of the equation
-        for var in vars:
-            varname = vars[var]["name"]
+        for var in variables:
+            varname = variables[var]["name"]
             if var_in_function_signature(name, varname):
-                vars[var]['sym'] = self.sym(varname)
+                variables[var]['sym'] = self.sym(varname)
             else:
-                vars[var]['sym'] = self.eq(varname)
+                variables[var]['sym'] = self.eq(varname)
 
         self._eqs[name] = \
-            vars['dydx']['sym'] * vars['dxdz']['sym'] + vars['dydz']['sym']
+            variables['dydx']['sym'] * variables['dxdz']['sym'] + variables['dydz']['sym']
 
     def _multiplication(self, name, x, y,
                         transpose_x=False, sign=1):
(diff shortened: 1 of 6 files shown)


python/amici/ode_export.py
```diff
@@ -0,0 +1,2266 @@
...
+        ----------
+
+        """
+        for var in self.variable_prototype:
+            self.generateSymbol(var)
+
+    def generateSparseSymbol(self, name):
+        """
+        Generates the sparse symbolic identifiers, symbolic identifiers,
+        sparse eqations, column pointers and row values for a symbolic variable
```
dweindl:
```suggestion
        sparse equations, column pointers and row values for a symbolic variable
```

python/amici/ode_export.py
```diff
@@ -0,0 +1,2266 @@
...
+            self._eqs[name] = self.eq('J').transpose()
+
+        elif name == 'JDiag':
+            self._eqs[name] = getSymbolicDiagonal(self.eq('J'))
+
+        elif name == 'x0_fixedParameters':
+            k = self.sym('k')
+            self._eqs[name] = sp.DenseMatrix([
+                eq
+                if any([sym in eq.free_symbols
```
dweindl: This is not particularly easy to read...

python/amici/ode_export.py
```diff
@@ -0,0 +1,2266 @@
...
+
+        else:
+            raise Exception(f'Unknown equation {name}')
+
+        if name in ['Jy', 'dydx']:
+            self._eqs[name] = self._eqs[name].transpose()
+
+    def symNames(self):
+        """
+        Returns a list of names of generates symbolic variables
```
dweindl:
```suggestion
        Returns a list of names of generated symbolic variables
```

python/amici/ode_export.py
```diff
@@ -0,0 +1,2266 @@
...
+                                  f'amici-{self.modelName}')
+        self.modelPath = os.path.abspath(output_dir)
+        self.modelSwigPath = os.path.join(self.modelPath, 'swig')
+
+        # Signatures and properties of generated model functions (see
+        # include/amici/model.h for details)
+
+        self.model = ode_model
+
+        self.functions = functions
```
dweindl: We should use a deepcopy here, otherwise there might be trouble when working with multiple ODE exporters. Not sure how often that will happen, but better safe than sorry.

python/amici/ode_export.py
```diff
@@ -0,0 +1,2266 @@
...
+
+        self.allow_reinit_fixpar_initcond = False
+
+    def compileModel(self):
+        """
+        Generates the native C++ code and compiles it into a simulateble module
```
dweindl:
```suggestion
        Generates the native C++ code and compiles it into a simulatable module
```

python/amici/ode_export.py
```diff
@@ -0,0 +1,2266 @@
...
+    """
+    varname = varname.replace('sparse', '')
+    return name in functions \
+           and re.search(
+                    f'const (realtype|double) \*{varname}[0]*[,)]+',
+                    functions[name]['signature']
+                )
+
+
+class ODEExporter:
```
dweindl: I propose separating C++ generation from compilation. Once there core files have been generated, there should be other functions/classes generating and compiling the interface files. This will me more modular and maintainable. 
It will allow people who just want to generate C++ code to not have swig installed. And if somebody feels like: It will also be easy to add another class for matlab wrapper generation, so that we can dump matlab sbml import completely.

python/amici/ode_export.py
```diff
@@ -0,0 +1,2266 @@
...
+            symbol = self.model.sparseeq(function)
+        else:
+            symbol = self.model.eq(function)
+
+        # function header
+        lines = [
+            '#include "amici/symbolic_functions.h"',
+            '#include "amici/defines.h" //realtype definition',
+            'using amici::realtype;',
+            '#include <cmath> ',
```
dweindl:
```suggestion
            '#include <cmath>',
```

python/amici/sbml_import.py
```diff
@@ -3,273 +3,191 @@
...
+        # check the error log just once after all conversion/validation calls.
         self.checkLibSBMLErrors()
 
         self.sbml = self.sbml_doc.getModel()
 
-
     def checkLibSBMLErrors(self):
-        """Generate AMICI C++ files for the model provided to the constructor.
+        """
+        Generate AMICI C++ files for the model provided to the constructor.
```
dweindl: This doesn't look right

python/amici/sbml_import.py
```diff
@@ -928,805 +846,237 @@ def computeModelEquations(self, observables=None, sigmas=None):
...
-        speciesSyms = self.symbols['species']['sym']
+        speciesSyms = self.symbols['species']['identifier']
 
         # add user-provided observables or make all species observable
-        if(observables):
-            # Replace logX(.) by log(., X) since symengine cannot parse the former
-            # also replace symengine-incompatible sbml log(basis, x)
+        if observables:
+            # Replace logX(.) by log(., X) since symengine cannot parse the
+            # former also replace symengine-incompatible sbml log(basis, x)
```
dweindl:
```suggestion
            # former. Also replace symengine-incompatible sbml log(basis, x)
```

python/amici/sbml_import.py
```diff
@@ -928,805 +846,237 @@ def computeModelEquations(self, observables=None, sigmas=None):
...
-                lines += self.getSymLines(symbol[:, iobs], variableName, 12)
-                lines.append(' ' * 12 + 'break;')
-            lines.append('}')
-        else:
-            if('colPtrs' in self.functions[function].keys()):
-                rowVals = self.functions[function]['rowVals']
-                colPtrs = self.functions[function]['colPtrs']
-                lines += self.getSparseSymLines(symbol, rowVals, colPtrs, variableName, 4)
+        constants = [
+            (sp.Symbol('avogadro'), sp.Symbol('6.02214179*1e23')),
```
dweindl:
```suggestion
            (sp.Symbol('avogadro'), sp.Symbol('6.02214179e23')),
```

python/amici/ode_export.py
```diff
@@ -0,0 +1,2266 @@
...
+                'sign': -1,
+            },
+            'xBdot': {
+                'x': 'JB',
+                'y': 'xB',
+                'sign': -1,
+            },
+        }
+
+    def import_from_sbml_importer(self, si, flux_as_expressions=True):
```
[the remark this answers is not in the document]
FFroehlich: setting the fluxes as expressions is a bit of a hack and requires access to protected members ODEModel. Accordingly I would rather keep them with ODEModel. I have added documentation to make this more evident

PR #462, opened 2018-11-11
Fix Critical Bug in ExpData construction from DataFrames

FFroehlich:
Previously, `edata.fixedParametersPreequilibration` or `edata.fixedParametersPresimulation` were not set if they were the same as `edata.fixedParameters` ...


dweindl:
Add test to prevent this happening again?

PR #465, opened 2018-11-17
Fix Swig ExpData constructor interface

FFroehlich:


FFroehlich:
will add testcase to check sllh in sbml_import tests

scripts/buildAmici.sh
```diff
@@ -19,7 +19,21 @@ make
 
 make python-sdist
 set -x
+
+rm -f ${AMICI_PATH}/python/sdist/amici/*.cxx
+rm -f ${AMICI_PATH}/python/sdist/amici/*.so
+rm -f ${AMICI_PATH}/python/sdist/amici/amici.py
+rm -f ${AMICI_PATH}/python/sdist/amici/amici_without_hdf5.py
+
+# test install from archive
```
dweindl: What about putting the python part to a separate file? I tend to use buildAmici.sh often in the case when I am not interested in the python interface. And especially two times installing amici and all its dependencies is quite time consuming.

FFroehlich: True!

PR #467, opened 2018-11-19
Replace symengine by sympy

FFroehlich:
symengine does not seem to be ready to be used in a production environment as it still contains critical bugs that are hard to anticipate and may very well lead to incorrect simulations.


.travis.yml
```diff
@@ -76,6 +74,7 @@ script:
     - scripts/runNotebook.sh python/examples/example_*/
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./scripts/run-valgrind.sh; fi
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ./scripts/run-doxygen.sh; fi
```
dweindl: Unrelated: I thought about splitting CI jobs. mtoc++, doxygen, ... and others which don't require AMICI compilation could be singled out and run in parallel. This could cut down travis runtime quite a bit. What do you think?

thomassligon:
Did you submit bug reports to Mathworks? All of the bug reports I submitted as part of GenSSI 2.0 were handled, with some quick workaround suggestions and code improvements (but the latter often came 1-2 versions later, i.e. 6-12 months).

Tom
(Dr. Thomas S. Ligon)
Tom@thomassligon.info<mailto:Tom@thomassligon.info>
Frohnloher Str. 6a
81475 Muenchen
Germany
Tel. +49(89)74575075

dweindl:
Hi Tom, it's about https://github.com/symengine/symengine.py https://github.com/symengine/symengine , not about the Matlab Symbolic Toolbox.

.travis.yml
FFroehlich: aren't we anyways limited to 2 concurrent jobs? If not thats a valid point

Issue #469, opened 2018-11-19
Model.clone() does not copy `steadyStateSensitivityMode`


FFroehlich:

FFroehlich:
same for `Solver.absoluteToleranceSensi` and `Solver.relativeToleranceSensi`

FFroehlich:
same for `Model.reinitializeFixedParameterInitialStates`

FFroehlich:
We _need_ better unit tests for this (probably best testable in python as dynamic member access is tricky in c++)

dweindl:
That's why I started getting rid of explicit copy constructors. Don't implement tests, change the class that the default copy constructor does the correct job.

PR #470, opened 2018-11-19
Fix Solver and Model copy constructors

FFroehlich:


FFroehlich:
fixes #469

include/amici/solver.h
```diff
@@ -48,6 +48,8 @@ class Solver {
         sensi = other.sensi;
         atol = other.atol;
         rtol = other.rtol;
+        atol_sensi = other.atol_sensi;
+        rtol_sensi = other.rtol_sensi;
```
dweindl: Okay, as hotfix, but better handling of the solver memory would avoid dealing with the long list of members in copy, copyassign, move, equality, ..... Which will get out of sync soon again ;). I guess the same members are also still missing in serialization functions.

tests/testCPP.py
```diff
@@ -0,0 +1,113 @@
...
+    def testCopyConstructors(self):
+        for obj in [self.model, self.solver]:
+            for attr in dir(obj):
+                with self.subTest(obj=obj, attr=attr):
+                    if callable(getattr(obj, attr)):
+                        # only use setter and getter functions
+                        if attr.startswith('get') \
+                                and 'set' + attr[3:] in dir(obj):
+
+                            if attr.endswith('ById') \
```
dweindl: 8 nesting levels? how can i review that?

PR #480, opened 2018-11-30
Make computing sx0_fixedParameters optional

dweindl:


dweindl:
(ping)

dweindl:
I don't like what codecov is doing here... Can we make that not required and rather consider that during review?

PR #482, opened 2018-12-04
fix(steadystate) fix solver copying in createSteadyStateSimSolver

FFroehlich:
a couple of tolerances were not correctly copied, lets use the default copy function and avoid all related problems


src/steadystateproblem.cpp
```diff
@@ -264,38 +263,30 @@ void SteadystateProblem::getSteadystateSimulation(ReturnData *rdata, Solver *sol
...
 }
 
 std::unique_ptr<CVodeSolver> SteadystateProblem::createSteadystateSimSolver(
         Solver *solver, Model *model, realtype tstart)
 {
-
-    
-    /* Create new CVode object */
+    /* Create new CVode solver object */
+    auto simsolver_ptr = dynamic_cast<CVodeSolver*>(solver->clone());
```
dweindl: Leaks `simsolver_ptr` on `throw` below. Put into `std::unique<Solver>` and convert later, or use as is

PR #489, opened 2018-12-12
Fix a couple of python package generation warnings

dweindl:


python/sdist/setup.py
```diff
@@ -39,7 +39,7 @@
 
 # Python version check. We need >= 3.6 due to e.g. f-strings
 if sys.version_info < (3, 6):
-    sys.exit('PEtab requires at least Python version 3.6')
+    sys.exit('amici requires at least Python version 3.6')
```
dweindl: @FFroehlich  Thx ;)

PR #490, opened 2018-12-12
Fix ParameterScaling issues

dweindl:


FFroehlich:
can you please add a testcase?

Issue #488, opened 2018-12-12
void Model::setParameterScale(std::vector<ParameterScaling> const& pscale) does not check input dimensions


dweindl:
should only accept vector of size `np`

PR #501, opened 2018-12-18
Release v0.8.2

### LINKED ISSUE #493: newline does not exist in older MATLAB releases
`amimodel.compileAndLinkModel>getCompilerVersionString` fails because it contains the `newline` function which was introduced in [R2016b](https://fr.mathworks.com/help/matlab/ref/newline.html). Replacing it by `char(10)`, as recommended in the link above, fixes the problem.

dweindl:
Let's close a bunch of issues...

Changes:

Features / improvements:
* Speedup symbolic processing for ODE generation in python

Bugfixes:
* Fix(python) Add missing deepcopy (introduced in 6847ba675f2088854db583199b8754aaa6e01576)
* Fix(core) Prevent parameter scaling length mismatch (#488)
* Fix(python) Set distutils dependency to current version to fix </usr/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'long_description_content_type'>
* fix(python) add symlink to version.txt to be included in package

Backwards-compatibility:
* Replace 'newline' by literal to restore <R2016b compatibility (Fixes #493)

Maintenance:
* Remove obsolete swig library build via cmake and related file copying
* Provide issue template for bug reports
* Providing valid SBML document to import is not optional anymore
* Update documentation and tests
* Add python version check and raise required version to 3.6 to prevent cryptic error messages when encountering f-strings


matlab/@amimodel/compileAndLinkModel.m
```diff
@@ -331,7 +331,7 @@ function updateFileHash(fileFolder,hashFolder,filename)
 
 function versionstring = getCompilerVersionString()
     [~,systemreturn] = system([mex.getCompilerConfigurations('c++').Details.CompilerExecutable ' --version']);
-    newlinePos = strfind(systemreturn,newline);
+    newlinePos = strfind(systemreturn, '\n');
```
FFroehlich: this does not work for os x, we actually do need `sprintf('\n')` here

PR #506, opened 2019-01-07
Merge v0.8.2 back into develop 

dweindl:


dweindl:
Does not work with current branch protection. Will create PR from separate branch.

PR #510, opened 2019-01-13
Various bugfixes

FFroehlich:


include/amici/edata.h
```diff
@@ -53,19 +53,19 @@ class ExpData {
      * @param observedEventsStdDev (dimension: nmaxevent x nztrue, row-major)
      */
     ExpData(int nytrue, int nztrue, int nmaxevent,
-            std::vector<realtype> ts,
-            std::vector<realtype> observedData,
-            std::vector<realtype> observedDataStdDev,
-            std::vector<realtype> observedEvents,
-            std::vector<realtype> observedEventsStdDev);
+            std::vector<realtype> const& ts,
```
dweindl: I think in most cases constness interferes here with the move inside the setters (works, but move has no effect).

include/amici/exception.h
```diff
@@ -39,79 +39,85 @@ namespace amici {
...
                     char *demangled = NULL;
                     int status = -1;
                     if (info.dli_sname[0] == '_')
-                        demangled = abi::__cxa_demangle(info.dli_sname, NULL, 0, &status);
-                    snprintf(buf, sizeof(buf), "%-3d %*p %s + %zd\n",
-                             i-2, int(2 + sizeof(void*) * 2), callstack[i],
-                             status == 0 ? demangled :
-                             info.dli_sname == 0 ? symbols[i] : info.dli_sname,
-                             (char *)callstack[i] - (char *)info.dli_saddr);
+                        demangled = abi::__cxa_demangle(info.dli_sname, NULL, 0,
```
dweindl: Maybe we can at some point define a `clang-format` style for coherent formatting. Started in `.clang-format` (#168), but never really used.

include/amici/model.h
```diff
@@ -927,7 +927,14 @@ namespace amici {
...
          */
         int checkFinite(const int N,const realtype *array, const char* fun) const;
-
+        
+        /**
+         * @brief Returns the amici version that was used to compile the model
+         * @return ver amici version string
+         */
+        virtual std::string getAmiciVersion() const {
+            throw AmiException("version not set during compilation");
```
dweindl:
```suggestion
            throw AmiException("Version not set during code generation");
```

python/amici/sbml_import.py
```diff
@@ -1107,3 +1109,34 @@ def checkLibSBMLErrors(sbml_doc, show_warnings=False):
...
+        Arguments:
+            sym: symbolic expressions @type sympy.Basic
+
+        Returns:
+
+        Raises:
+            raises SBMLException if an unsupported function is encountered
+    """
+
+    unsupported_functions = [sp.factorial, sp.ceiling]
```
dweindl: Can we add `piecewise` here? Occurred in some models of the benchmark collection.

scripts/run-cppcheck.sh
```diff
@@ -8,7 +8,32 @@ AMICI_PATH=$(cd $SCRIPT_PATH/.. && pwd)
...
+
+
+# suppress header warnings for standard libraries
+grep -v "Cppcheck cannot find all the include files" cppcheck.txt > cppcheck_tmp.txt
+mv cppcheck_tmp.txt cppcheck.txt
+
+grep -v "'AmiVectorArray' does not have a operator=" cppcheck.txt > cppcheck_tmp.txt
+mv cppcheck_tmp.txt cppcheck.txt
+
+grep -v "Member variable 'ExpData::nytrue_' is not initialized in the constructor" cppcheck.txt > cppcheck_tmp.txt
```
dweindl: Shouldn't we rather fix those? Either do in every constructor or set to zero in the class definition.

src/model.cpp
```diff
@@ -602,9 +605,11 @@ const std::vector<int> &Model::getParameterList() const {
 }
 
 void Model::setParameterList(const std::vector<int> &plist) {
-    for(auto const& idx: plist)
-        if(idx < 0 || idx >= np())
-            throw AmiException("Indices in plist must be in [0..np]");
+    int np = this->np(); // cannot capture 'this' in lambda expression
+    if(std::any_of(plist.begin(), plist.end(),
```
dweindl: :+1:

include/amici/exception.h
FFroehlich: this was formatted by clang format :D

scripts/run-cppcheck.sh
FFroehlich: this is a false positive, they are initialized in the delegate constructors which cppcheck does not recognize

include/amici/exception.h
dweindl: Ah, I thought the old version was. Alright then ;).

dweindl:
Let's get that merged?

PR #513, opened 2019-01-14
feature(steadystate) use klu_refactor instead of always factoring de …

### LINKED ISSUE #421: [newton] Use klu_refactor instead of repeatedly factorizing the problem de-novo
We should be able to use `klu_refactor` instead of repeatedly using `klu_factor`, right?

see https://github.com/PetterS/SuiteSparse/blob/master/KLU/Source/klu_refactor.c

FFroehlich:
…novo, fixes #421

### DIFF
--- a/src/newton_solver.cpp
+++ b/src/newton_solver.cpp
@@ -9,6 +9,8 @@
 #include "amici/rdata.h"
 #include "amici/edata.h"
 
+#include "sundials/sundials_math.h"
+
 #include <cstring>
 #include <ctime>
 #include <cmath>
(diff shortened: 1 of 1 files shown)


src/newton_solver.cpp
```diff
@@ -255,22 +256,46 @@ void NewtonSolverSparse::prepareLinearSystem(int ntry, int nnewt) {
...
+        if (!klu_refactor(Jtmp->indexptrs, Jtmp->indexvals, Jtmp->data,
+                          symbolic, numeric, &common))
+            throw NewtonFailure(common.status, "klu_refactor");
+        /* check cheap estimate of rcond to see if we need to recompute
+         factorization
+         */
+        if (!klu_rcond(symbolic, numeric, &common))
+            throw NewtonFailure(common.status, "klu_rcond");
+
+        if (common.rcond < SUNRpowerR(UNIT_ROUNDOFF, 2 / 3)) {
```
dweindl: Too late, but: `2/3` performs integer division. The result might not be as expected. Is this intended or did you mean `2.0/3.0`? (same 5 lines below)

FFroehlich: duh. should be `2.0/3.0`

PR #521, opened 2019-01-15
Reduce logsize

FFroehlich:
size of travis log was always >10k lines, which makes it a bit hard to navigate. Let's remove the biggest offenders.

### DIFF
--- a/scripts/buildCpputest.sh
+++ b/scripts/buildCpputest.sh
@@ -23,6 +23,6 @@ fi
 cd cpputest-master
 mkdir -p build
 cd build
-cmake -DTESTS=OFF -DBUILD_TESTING=OFF -DC++11=ON .. -quiet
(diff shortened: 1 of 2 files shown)


scripts/buildCpputest.sh
```diff
@@ -14,15 +14,15 @@ export CPPUTEST_BUILD_DIR=${AMICI_PATH}/ThirdParty/cpputest-master/
...
-    unzip cpputest-master.zip
+    unzip -q cpputest-master.zip
     #cd cpputest-master/ && ./autogen.sh && ./configure && make
 fi
 
 cd cpputest-master
 mkdir -p build
 cd build
-cmake -DTESTS=OFF -DBUILD_TESTING=OFF -DC++11=ON ..
+cmake -DTESTS=OFF -DBUILD_TESTING=OFF -DC++11=ON .. -quiet
```
dweindl: Hmm.. my cmake doesn't have this option.

PR #523, opened 2019-01-16
lets check whether brew update ist still needed

FFroehlich:


FFroehlich:
this looks like its going to produce a travis shitshow, lets not do that right now.

PR #526, opened 2019-01-16
ci(travis) Fold script section outputs

dweindl:


FFroehlich:
is this really looking as intended?

dweindl:
> is this really looking as intended?

Unfortunately not... My intentions were better...

dweindl:
Good enough now...

Issue #398, opened 2018-08-10
[core] Change Event Logic


FFroehlich:
Apparently we are currently abusing CVodeReInit to treat events, which required some changes to CVodes code to properly handle events. This deprives us of the option to us CVodeReInit to reinitialize the solver for a new problem.

Instead of using CVodeReInit, we should test whether we can achieve the same results by forcing the order to 1 when events occur. This should allow easier upgrading and better compatibility with sundials (version 4.0 is soon to be released, which would be a good opportunity to make these changes) and enable a more straightforward reuse of `Solver` objects.

FFroehlich:
fixed in c27df28062f7b96fb8fe43415053445c041fdf89

PR #549, opened 2019-01-21
Sundials 4.0.2

FFroehlich:


FFroehlich:
okay, this is going to be a large PR

src/sundials_matrix_wrapper.cpp
```diff
@@ -1,126 +1,83 @@
...
-
-
-
-DlsMatWrapper::DlsMatWrapper(long int M, long int N)
-    : matrix(NewDenseMat(M, N))
+    
+SUNMatrixWrapper::SUNMatrixWrapper(int M, int N)
+: matrix(SUNDenseMatrix(M, N))
 {
-    if((M*N > 0) && !matrix)
```
dweindl: That gave me a compiler warning. Multiplication in boolean context or something like that.

FFroehlich:
I think most of the heavy lifting is done, but some issues still remain:

- [ ] create RAII wrappers for sundials classes `SUNNonlinearSolver`  and `SUNLinearSolver`
- [ ] implement creation `SUNNonlinearSolver` (based on `iter`) and `SUNLinearSolver` (based on `linsol`) instances in `solver_cvodes` and `solver_idas` interfaces and remove deprecated function
- [ ] replace functionality `solver::ReInit` and `solver::sensReInit`, `solver::reInitB` and `solver::quadReInitB` by resetting order to zero and updating respective vectors in the solver memory (see #398) 

@dweindl could you take care of the first two?

dweindl:
@FFroehlich Compiles now and tests pass, except for model_robertson_test. Will clean up more.

dweindl:
While we are on it: sundials 4.0.2 was released last week ;)

src/solver.cpp
```diff
@@ -213,154 +212,173 @@ void Solver::getDiagnosisB(const int it, ReturnData *rdata, int which) const {
...
+        throw AmiException("Invalid choice of solver: %d", static_cast<int>(linsol));
+    }
+}
+
+void Solver::initializeNonLinearSolverB(AmiVector *xB, const int which, Model *model)
+{
+    switch(iter) {
+    case NonlinearSolverIteration::newton:
+        switch(ism) {
+        case InternalSensitivityMethod::staggered:
```
FFroehlich: We never use backwards integration + sensitivities, so this should not be necessary.

FFroehlich:
Okay, I think everything but the events stuff should now be working

FFroehlich:
okay `example_adjoint` is working now. Still need to implement the event stuff for IDAS and after that we probably will have to do some general cleanup, but it looks like might be able to merge this PR soon.

dweindl:
Great! Will start looking into the cleaning part.

dweindl:
Tests are passing for you? I have 5 failing here.

PR #554, opened 2019-01-22
[WIP] Redirect C/C++ output to show up in ipython notebooks

dweindl:
Works now with IPython, Python, pycharm.


dweindl:
I would make it mandatory unless it turns out to be a problem on some system. I think using amici within a notebook is not such an exotic use case. Installing it later might be more annoying to most users.

However, I just noticed that it interferes with running `python setup.py` :-/

PR #562, opened 2019-01-28
Hotfix matlab & Newton solver, fix #560

### LINKED ISSUE #560: Dimension mismatch when writing rdata->newton_numsteps to matlab results
Hi there

I think there might be a possible bug in the file AMICI/matlab/@amimodel/compileAndLinkModel.m. There should be an ellipsis at the end of line 186.

A second problem I ran into whilst using the __Matlab interface__: Previously, in version 0.8.1, I was able to compile my model using amiwrap, and this returned as output the simulate_${MODELNAME}.m, which I could call without getting any error. However, now, in __version 0.9.1__, although I can still compile my model using amiwrap, running the resulting output simulate_${MODELNAME}.m generates the following error:
```
'Unexpected Standard exception from MEX file.↵What() is:Dimension mismatch when writing rdata->newton_numsteps to matlab results↵..'
```

This error is thrown from line 210 in AMICI/src/returndata_matlab.cpp. As I'm not sure what I happening, I had to post this issue.

Other details:
 * macOS Mojave (version 10.14.2)
 * Clang++ compiler (Apple LLVM version 10.0.0, clang-1000.11.45.5, Target: x86_64-apple-darwin18.2.0)
 * Matlab version R2017a

paulstapor:
The matlab interface for the Newton solver had a little flaw. Besides that, there was a sign error in the Newton solver. So this PR is also relevant for the other interfaces. I think it should therefore rather go to master than to develop.


paulstapor:
It seems like the test for the steady state model is off now...
It might be, that the test results for the steady state model are corrupted... or that there is something screwed up du to the fact I pushed also the header file for this model. However, the only change in that file was the line stating the Amici commit...
Any ideas on that?

paulstapor:
Sure, we should just recreate the testing data as well...

paulstapor:
@FFroehlich :
I recreated the test results and uploaded them. I'm currently not authorized to merge this PR, so you might want to do it, if you're fine with the changes.

PR #566, opened 2019-01-28
Fix 559

FFroehlich:


src/model.cpp
```diff
@@ -1571,7 +1571,6 @@ int Model::checkFinite(const int N, const realtype *array, const char *fun) cons
     auto result = amici::checkFinite(N, array, fun);
 
     if(result != AMICI_SUCCESS) {
-        amici::checkFinite(ts.size(), ts.data(), "ts");
```
dweindl: When using adjoints, a check for inf might be appropriate. Not sure if that is done in any other place

include/amici/defines.h
```diff
@@ -19,6 +19,7 @@ constexpr double pi = 3.14159265358979323846;
 
 /* Return codes */
 #define AMICI_RECOVERABLE_ERROR          1
+#define AMICI_UNRECOVERABLE_ERROR       -1
```
dweindl: Is there a specific reason to have AMICI_UNRECOVERABLE_ERROR and AMICI_TOO_MUCH_WORK indistinguishable?

src/misc.cpp
```diff
@@ -7,21 +7,26 @@
...
+        if (isInf(array[idx])) {
+            warnMsgIdAndTxt(
+                "AMICI:Inf",
+                "AMICI encountered an Inf value at index %i of %i in %s!", idx,
+                N, fun);
+            return AMICI_RECOVERABLE_ERROR;
         }
     }
-    return(AMICI_SUCCESS);
+    return (AMICI_SUCCESS);
```
dweindl:
```suggestion
    return AMICI_SUCCESS;
```

include/amici/defines.h
FFroehlich: no, value of `AMICI_UNRECOVERABLE_ERROR` does not matter, just needs to be strictly negative.

src/model.cpp
FFroehlich: hmm good point!

PR #593, opened 2019-02-11
Release 0.9.4

### LINKED ISSUE #586: Solver diagnostics for adjoint sensitivities not written to ReturnData for the last timepoint
... since `BackwardProblem::handleDatapointB`, and thus, `Solver::getDiagnosisB` is never called.

Should they also be reported if the next timepoint is never reached? (CVODES mxsteps error)

dweindl:
Minor fixes only:
- fix(core) Get solver diagnostics for first(last) timepoint (#588) (Closes #586)
- fix(ci) Fix autodeploy (Closes #589)


include/amici/solver.h
```diff
@@ -75,7 +75,7 @@ class Solver {
      * @param model pointer to the model object
      */
 
-    void setupAMIB(BackwardProblem *bwd, Model *model);
+    void setupB(BackwardProblem *bwd, Model *model);
```
yannikschaelte: why shorter? + maybe remove the empty line before?

src/backwardproblem.cpp
```diff
@@ -178,36 +127,19 @@ void BackwardProblem::handleEventB(const int iroot) {
     model->updateHeavisideB(&rootidx[iroot * model->ne]);
 }
 
-/**
- * handleDataPoint executes everything necessary for the handling of data
- * points for the backward problems
- *
- * @param it index of data point @type int
- */
+
```
yannikschaelte: I assume these docs all went to the h files?

include/amici/solver.h
dweindl: AMI prefix is a relict from C times. We have amici namespace now.

src/backwardproblem.cpp
dweindl: Assumption correct.

PR #600, opened 2019-02-12
Fix sign in undamped Newton step.

paszkow:
First, in NewtonSolver::getStep() a passed-in delta vector gets
a reverse sign by calling 'delta->minus();' and then its passed to
NewtonSolver::linsolveSPBCG via NewtonSolver::solveLinearSystem
where the sign is again reversed (xdot = *ns_delta; xdot.minus();)

Therefore -gamma must be applied in the Newton step otherwise the
Newton method does not converge at all.


paulstapor:
I'll look into that today. However, the Newton solver worked fine with the direct solvers lately... So I'm not sure: Would your change also work for the non-iterative solvers?

paszkow:
So probably this change will make direct solvers not work anymore. Another possible fix would be to restore a previous sign of gamma and remove the line xdot.minus() from the function NewtonSolverIterative::linsolveSPBCG.

paulstapor:
Okay, thanks! I'll check that out this afternoon and see, whether I get that done.

FFroehlich:
Sounds like we need to implement a small unit test.

paulstapor:
Okay, I just checked it:
just flipping the sign spoils convergence for direct solvers. So we need to go for something more specific here. I'll dig a bit deeper into that...

paulstapor:
Okay, for some reason, the hand-crafted, iterative linear solver returns an update vector which is just flipped when compared to the direct solvers...
It's not exactly a beautiful solution, but the easiest possible fix for that which does not spoil anything, would be 
inserting on line after line 372 of newton_solver.cpp  with
`rhs->minus();`
At lest locally on my computer, that does the job..

dweindl:
@FFroehlich @paulstapor Is this to be merged?

FFroehlich:
I'd say yes.

PR #603, opened 2019-02-13
feature(python) PEtab import via CLI

dweindl:


dweindl:
@FFroehlich : Are you using `sbml_import.constantSpeciesToParameters`? I would move that to PEtab.
Are you using `sbml_import.assignmentRules2observables`? I'd remove that here. There is a better version of that in PEtab, where also the format is documented.

FFroehlich:
Nope, not using any of those functions.

dweindl:
> Nope, not using any of those functions.

`assignmentRules2observables` is used in `testPYSB.py` and `testSBML.py`. So I will leave it in, I guess.

PR #608, opened 2019-02-15
Fix edata to df

### LINKED ISSUE #607: DataFrame export: Have an option to use the id for observables as column headers
In https://github.com/ICB-DCM/AMICI/blob/master/python/amici/pandas.py#L8 (``amici.pandas.getDataObservablesAsDataFrame(model, edata_list)``, observable names are used as column headers. Since we use that in pypesto, it would be good to have an option ``use_ids=True`` or sth like that to use the observable ids instead.

In addition, one could think about using the id also when the name is empty ("").

yannikschaelte:
Give ``amici.pandas.py`` functions an option ``by_id=False``. This is always carried on to the ``_get_names_or_ids`` function which now won't try to use names if True. The default behavior is not changed.

Handling of empty values not implemented yet, since there is a check that fails when there are multiple empty names anyway.

fixes #607 .

### DIFF
--- a/python/amici/pandas.py
+++ b/python/amici/pandas.py
@@ -5,12 +5,16 @@ import copy
 from .numpy import edataToNumPyArrays
 from amici import ExpData
 
-def getDataObservablesAsDataFrame(model, edata_list):
+def getDataObservablesAsDataFrame(model, edata_list, by_id=False):
     """ Write Observables from experimental data as DataFrame
 
     Arguments:
         model: Model instance
         edata_list: list of ExpData instances with experimental data
+        by_id: bool (optional, default = False)
+            If True, uses observable ids as identifiers in dataframe,
+            otherwise the possibly more descriptive observable names
+            are used.
 
     Returns:
         pandas DataFrame with conditions and observables
@@ -19,7 +23,7 @@ def getDataObservablesAsDataFrame(model, edata_list):
 
     """
 
-    cols = _get_extended_observable_cols(model)
+    cols = _get_extended_observable_cols(model, by_id=by_id)
     df_edata = pd.DataFrame(columns=cols)
 
     for edata in edata_list:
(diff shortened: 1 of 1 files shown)


python/amici/pandas.py
```diff
@@ -2,174 +2,247 @@
...
-    """ Convert a list of ExpData to pandas DataFrame
+def getResidualsAsDataFrame(model, edata_list, rdata_list, by_id=False):
+    """ Convert a list of ExpData to pandas DataFrame.
 
     Arguments:
-        model: Model instance
-        edata_list: list of ExpData instances with experimental data
-        rdata_list: list of ReturnData instances corresponding to ExpData
+        model: Model instance.
+        edata_list: list of ExpData instances with experimental data.
```
FFroehlich: could also mention here that we allow passing of single instances/

python/amici/pandas.py
```diff
@@ -2,174 +2,247 @@
...
+            descriptive names.
 
     Returns:
-        pandas DataFrame with conditions and observables
+        pandas DataFrame with conditions and observables.
 
     Raises:
 
     """
+    if isinstance(edata_list, amici.amici.ExpData):
```
FFroehlich: why not `amici.ExpData`? could also add `amici.ExpDataPtr` here (or whatever the swig export of the smart pointer is called)

python/amici/pandas.py
```diff
@@ -189,226 +262,272 @@ def _fill_conditions_dict(datadict, model, edata) -> dict:
...
+    if not by_id and len(set(namegetter())) == len(namegetter()) \
             and model.hasObservableNames():
+        # use variable names
         return list(namegetter())
     elif model.hasObservableIds():
+        # use variable ids
         return list(idgetter())
     else:
+        # unable to create unique lables
         raise RuntimeError('Model Observable Names are not unique and '
```
FFroehlich: I think we should customize the Exception text according to the value of `by_id`

FFroehlich: Also just noticed that the function was generalized to different variable without updating the exception text, replace 
```suggestion
        raise RuntimeError(f'Model {variable} names are not unique and '
```

python/amici/pandas.py
yannikschaelte: ah, I thought the class instance has to be called as ``amici.amici.ExpData```. Just ``amici.ExpData`` had not worked in another project some weeks ago.

python/amici/pandas.py
yannikschaelte: in the method, also ``model.hasObservableIds()`` is called. That would also not fit for general variables, right?

FFroehlich: Yes! Needs to be replaced by `has{Variable}Names()`. See http://icb-dcm.github.io/AMICI/classamici_1_1_model.html

python/amici/pandas.py
FFroehlich: should work now, if not, please let me know.

yannikschaelte: Doesn't seem to work: https://travis-ci.org/ICB-DCM/AMICI/jobs/496114885#L7769

FFroehlich: Ah right, sorry! `amici.ExpData` only exposes the constructor, not the class itself.
lets do 
```suggestion
    if isinstance(edata_list, (amici.amici.ExpData, amici.amici.ExpDataPtr)):
```
then.

yannikschaelte:
does somebody know why travis fails? https://travis-ci.org/ICB-DCM/AMICI/jobs/496167567
I don't find the problem.

FFroehlich:
> does somebody know why travis fails? https://travis-ci.org/ICB-DCM/AMICI/jobs/496167567
> I don't find the problem.

Currently investigating with @dweindl. As this is happening in all PRs, we suppose it's not due to any changes in AMICI. We couldn't really identify what is causing it so far. Current attempt to fix it is in https://github.com/ICB-DCM/AMICI/pull/612/commits/5672cbe3a6d909846ecb132918a5d59190fb2234

dweindl:
> Current attempt to fix it is in [5672cbe](https://github.com/ICB-DCM/AMICI/commit/5672cbe3a6d909846ecb132918a5d59190fb2234)

Fixed by @FFroehlich and to be merged within the next hour.

PR #612, opened 2019-02-20
feature(python) Allow for light installation without building any ext…

### LINKED ISSUE #610: In SbmlImporter.sbml2amici, make model compilation optional
One might want to only use the generated source files from C++ or Matlab. So there is no need in waiting for compilation to finish or fail due to missing dependencies. Introduce `compile=True`.

dweindl:
…ensions. Allow for model generation without compilation (Closes #610)


python/amici/sbml_import.py
```diff
@@ -150,43 +152,48 @@ def loadSBMLFile(self, SBMLFile):
...
-                   modelName,
-                   output_dir=None,
-                   observables=None,
-                   constantParameters=None,
-                   sigmas=None,
-                   verbose=False,
-                   assume_pow_positivity=False,
-                   compiler=None,
-                   allow_reinit_fixpar_initcond = True
+                   modelName: str,
```
FFroehlich: Nice! What is doxygen doing with this type information? Can we remove @type instructions in the docstring? Would prefer to encode type info in code and not in docstrings that are prone to be stalled.

dweindl: Yeah, I found typehinting extremely useful. A proper IDE will tell you that you are passing the wrong arguments. Didn't check what doxygen makes of it yet, would be way more readable and up to date, yes.

dweindl: Doesn't work with doxygen :(. Maybe this can be changed somehow.

dweindl: https://github.com/UwePfautsch/doxyfilter_python should work with it, but there were other issues. Will need some further checking...

dweindl: Doxypy cannot identify arguments with typehints. We might want to fix the signature parsing ourselves.  

> /home/dweindl/src/AMICI-devel/python/amici/sbml_import.py:241: warning: argument 'observables' of command @param is not found in the argument list of SbmlImporter::sbml2amici(self, modelName)

PR #614, opened 2019-02-21
Fix #580

### LINKED ISSUE #580: Autogenerate matlab compilation script when importing models with Python
AMICI-code generated with the Python SBML-importer can be compiled for Matlab. An example is shown in the steadystate example notebook. Such a compile script should be created in the model folder automatically when writing the model code.

paulstapor:
Creates a little Matlab compile script from python code

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -2003,6 +2003,17 @@ class ODEExporter:
         # creating the code lines for the Matlab compile script
         lines = []
 
+        # Events are not yet implemented. Once this is done, the variable nz
+        # will have to be replaced by "self.model.nz()"
+        nz = 0
+
+        # Second order code is not yet implemented. Once this is done,
+        # those variables will have to be replaced by
+        # "self.model.<var>true()", or the corresponding "model.self.o2flag"
+        nxtrue_rdata = self.model.nx_rdata()
+        nytrue = self.model.ny()
+        o2flag = 0
+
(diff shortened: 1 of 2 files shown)


python/amici/ode_export.py
```diff
@@ -1989,6 +1990,56 @@ def _compileCCode(self, verbose=False, compiler=None):
...
+        lines.append('% This compile script was automatically created from Python SBML import.')
+        lines.append('% If mex compiler is set up within MATLAB, it can be run from MATLAB ')
+        lines.append('% in order to compile a mex-file from the Python generated C++ files.')
+        lines.append('')
+
+        # write the actual compiling code
+        lines.append('''modelName = '{model_name}';'''.format(
+            model_name=self.modelName))
+        lines.append('''modelDir = '{model_output_dir}';'''.format(
+            model_output_dir=os.path.abspath(self.modelPath)))
```
dweindl: Maybe it's better to use the relative path here? Let's assume the resulting matlab file always stays within the model directory. The matlab script should just use the mfilepath.

paulstapor: Hmmm...
Relative and absolute paths are reasonable to some degree: Personally, I never use the model folders for anything, but would rather move the compile model file. However, it would be easy for me to compile the model in place and then just move the mex file and the simulate file...
Probably you're right, since portability of the generated model is a very nice feature to have...
I'll change that.

PR #616, opened 2019-02-21
Faster conservation laws

FFroehlich:
Currently still working on improvements for flatten_conservation_laws

PR #623, opened 2019-02-23
0.9.5 Release

FFroehlich:


dweindl:
@FFroehlich : What else to be included here?

FFroehlich:
at least #590 and #616, but for #616 travis has decied to rather throw a tantrum instead of finishing the goddamn tests

PR #629, opened 2019-02-27
Refactor dwdp, fixes #537

### LINKED ISSUE #537: Speed up compilation
For me it seems as if compilation of `JSparse` (10 min) and `JSparseB` (13 min) can take excessively long compared to `J` (<1min) and `JB` (<1min) for large models (~15k nonzero entries). The main difference is `indexvals` and `indexptrs`, might be worthwhile to put them in a seperate function. Did anybody have a similar experience?

FFroehlich:


include/amici/abstract_model.h
```diff
@@ -666,7 +666,25 @@ class AbstractModel {
                        const realtype *stcl);
 
     /**
-     * Model specific implementation of dwdx
+     * Model specific sensitivity implementation of dwdp
```
dweindl: `@brief` please :)

include/amici/model.h
```diff
@@ -56,6 +56,7 @@ class Model : public AbstractModel {
      * repeating elements
      * @param ndwdp number of nonzero elements in the p derivative of the
      * repeating elements
+     * @param ndxdotdw number of nonzero elements in the w derivate of xdot
```
dweindl:
```suggestion
     * @param ndxdotdw number of nonzero elements in the w derivative of xdot
```

include/amici/model_ode.h
```diff
@@ -160,7 +164,48 @@ namespace amici {
         virtual void fJSparse(SUNMatrixContent_Sparse JSparse, const realtype t,
                               const realtype *x, const realtype *p,
                               const realtype *k, const realtype *h,
-                              const realtype *w, const realtype *dwdx) = 0;
+                              const realtype *w, const realtype *dwdx) {
+            throw std::invalid_argument("Not implement for this model!");
```
dweindl:
```suggestion
            throw std::invalid_argument("Not implemented for this model!");
```

include/amici/model_ode.h
```diff
@@ -160,7 +164,48 @@ namespace amici {
...
+            // don't use AmiException as the backtrace causes some hefty
```
dweindl: Is that really a concern when an non-recoverable exception is thrown?

Will this exception pass through runAmiciSimulation? Need to catch std::exception separately and print, not via catchall which only prints generic message.

include/amici/sundials_matrix_wrapper.h
```diff
@@ -131,6 +131,12 @@ class SUNMatrixWrapper {
      * @return index array
      */
     int sparsetype() const;
+    
+    /**
+     * @brief reset data to zeroes
+     * @return index array
```
dweindl:
```suggestion
```

Doesn't return anything

python/amici/ode_export.py
```diff
@@ -2103,6 +2109,48 @@ def _writeFunctionFile(self, function):
...
+        if indextype == 'colptrs':
+            values = self.model.colptrs(function)
+        elif indextype == 'rowvals':
+            values = self.model.rowvals(function)
+        else:
+            raise ValueError('Invalid value for type, must be colptr or '
+                             'rowval')
+
+        lines = list()
+        lines.append('#include "sundials/sundials_types.h"')
```
dweindl: Combining `append`s would avoid some reallocations. Not sure if measurable though.

src/model.cpp
```diff
@@ -1027,9 +1030,14 @@ void Model::fdydp(const realtype t, const AmiVector *x) {
 
     std::fill(dydp.begin(),dydp.end(),0.0);
     fw(t,x->data());
-    fdwdp(t,x->data());
+    auto dense = fdwdp(t,x->data());
+    // if dwdp is not dense, fdydp will expect the full sparse array
+    realtype *dwdp_tmp = dwdp.data();
     for(int ip = 0; ip < nplist(); ip++){
         // get dydp slice (ny) for current time and parameter
+        if (dense && nw > 0) // if dwdp is dense, fdydp will expect a slice
```
dweindl: Move if outside loop. Better 2 separate loops than conditional inside loop.

src/model.cpp
```diff
@@ -1428,23 +1436,59 @@ void Model::fw(const realtype t, const realtype *x) {
...
 }
 
-void Model::fdwdp(const realtype t, const realtype *x) {
-    fw(t,x);
-    std::fill(dwdp.begin(),dwdp.end(),0.0);
-    fdwdp(dwdp.data(), t, x, unscaledParameters.data(), fixedParameters.data(), h.data(), w.data(), total_cl.data(), stotal_cl.data());
+bool Model::fdwdp(const realtype t, const realtype *x) {
+    // if nw is zero, slicing dwdp throw an exception, we don't actually know
+    // whether we are in the matlab (sparse) or python (dense) case, but
+    // simulating the sparse case will tell the upstream function not to slice
```
dweindl: This seems like an abuse of exceptions :).

src/model_ode.cpp
```diff
@@ -42,9 +42,17 @@ namespace amici {
...
-        fJSparse(static_cast<SUNMatrixContent_Sparse>(SM_CONTENT_S(J)), t,
-                 N_VGetArrayPointer(x_pos), unscaledParameters.data(),
-                 fixedParameters.data(), h.data(), w.data(), dwdx.data());
+        try {
+            fJSparse(SM_DATA_S(J), t, N_VGetArrayPointer(x_pos),
+                     unscaledParameters.data(), fixedParameters.data(),
+                     h.data(), w.data(), dwdx.data());
+            fJSparse_colptrs(SM_INDEXPTRS_S(J));
+            fJSparse_rowvals(SM_INDEXVALS_S(J));
+        } catch (std::invalid_argument &) {
```
dweindl: Again somewhat abusive. Can't we check some type or add some model flag instead? Or is this planned to change sometime soon?

src/model.cpp
FFroehlich: cant move slicing out of loop :(

PR #632, opened 2019-02-28
Return root function value at end if Amici is called without edata

### LINKED ISSUE #620: sol.rz is always 0 even event doesn't triggered in caspase model
I just added the caspase model in AMICI example. It seems that the trigger function is returned in `example_event`, but it doesn't work for caspase model.

paulstapor:
So far, the root function value (which is used for penalization of non-observed events) was only returned correctly, if amidata (or edata) was passed. 
However, it is possible to specify a number of observed events using the options (and not using the data), so Amici knows which number of events is to be expected. In this case, root function values should be reported and not set to zero (which was the case so far).
This fixes #620 .


paulstapor:
ok, nice...
@FFroehlich : I'm not allowed to merge. SO either, I get allowance, or you merge. Both is fine for me... :)

FFroehlich:
okay, will merge as soon as CI reran, thanks for the fix!

PR #637, opened 2019-03-02
Fix pandas ids names

### LINKED ISSUE #636: Get variable names or ids if the list is (correctly) empty raises Error in pandas.py
Example: The Boehm model has no fixed parameters. Since we fixed an error which was only testing for ObservableIds but not {variable}Ids, this throws an error, in https://github.com/ICB-DCM/AMICI/blob/master/python/amici/pandas.py#L381.

yannikschaelte:
Fixes #636

### DIFF
--- a/python/amici/pandas.py
+++ b/python/amici/pandas.py
@@ -367,28 +367,27 @@ def _get_names_or_ids(model, variable, by_id):
     # check whether variable type permitted
     variable_options = ['Parameter', 'FixedParameter', 'Observable', 'State']
     if variable not in variable_options:
-        raise ValueError('variable must be in ' + str(variable_options))
+        raise ValueError('Variable must be in ' + str(variable_options))
 
-    # functions to extract attributes
-    namegetter = getattr(model, 'get' + variable + 'Names')
-    idgetter = getattr(model, 'get' + variable + 'Ids')
+    # extract attributes
+    names = list(getattr(model, f'get{variable}Names')())
+    ids = list(getattr(model, f'get{variable}Ids')())
 
     # extract labels
-    if not by_id and len(set(namegetter())) == len(namegetter()) \
-            and getattr(model, f"has{variable}Names")():
+    if not by_id and len(set(names)) == len(names):
         # use variable names
-        return list(namegetter())
-    elif getattr(model, f"has{variable}Ids")():
+        return names
+    elif len(ids) > 0 or len(ids) == 0 and len(names) == 0:
         # use variable ids
-        return list(idgetter())
+        return ids
     else:
         # unable to create unique labels
         if by_id:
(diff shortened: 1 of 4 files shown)


python/amici/pandas.py
```diff
@@ -367,28 +367,27 @@ def _get_names_or_ids(model, variable, by_id):
...
-    namegetter = getattr(model, 'get' + variable + 'Names')
-    idgetter = getattr(model, 'get' + variable + 'Ids')
+    # extract attributes
+    names = list(getattr(model, f'get{variable}Names')())
+    ids = list(getattr(model, f'get{variable}Ids')())
 
     # extract labels
-    if not by_id and len(set(namegetter())) == len(namegetter()) \
-            and getattr(model, f"has{variable}Names")():
+    if not by_id and len(set(names)) == len(names):
```
FFroehlich: should check for `has{variable}Names` here, otherwise names = [] will lead to false positivies here.

python/amici/pandas.py
```diff
@@ -367,28 +367,27 @@ def _get_names_or_ids(model, variable, by_id):
...
         # use variable names
-        return list(namegetter())
-    elif getattr(model, f"has{variable}Ids")():
+        return names
+    elif len(ids) > 0 or len(ids) == 0 and len(names) == 0:
```
FFroehlich: I believe this rather is a bug with has{Variable}Ids, I would rather change the implementations of has{Variable}Ids and has{Variable}Names unless @dweindl still remembers why they are implemented the way they currently are.

FFroehlich: more specifically change `&&` to `||` in the c++ code.

python/amici/pandas.py
yannikschaelte: I assumed that has{variable}Names() == len(get{variable}Names()). That is not true, then? What is the exact intended behavior? If that is not done yet, I would also document it.

FFroehlich: > boolean indicating whether parameter names were set

If len(getParameterNames) is zero, we do not know whether (i) the model was compiled without overriding getParameterNames (ii) np == 0. So if len(getParameterNames) == 0, you would have to explicitly check if np > 0 to see whether Names are _undefined_. Apparently we screwed up the implementation as the right thing to check is `np == 0 || getParameterNames.size()`

yannikschaelte: ok, now it makes sense, thanks! I will change that then.

yannikschaelte:
@FFroehlich I revised the changes now.

yannikschaelte:
Merging can only be done by authorized developers. @FFroehlich or @dweindl could you do that?

FFroehlich:
lifted that restriction for develop branch, can you merge now?

PR #639, opened 2019-03-02
Feature from sbml model

### LINKED ISSUE #638: SBML import from model, not from file
For use in pypesto+petab, it would be good to enable creating an amici model based on sbml model (+ document + reader currently), not on an underlying file.

yannikschaelte:
Allow to pass to the sbml import not only a file name, but also already an encoded sbml string. This enables not requiring to save a model written in petab back to file, before creating an amici model from it. Fixes #638 .

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -76,7 +76,11 @@ class SbmlImporter:
 
     """
 
-    def __init__(self, SBMLFile, show_sbml_warnings=False):
+    def __init__(
+            self,
+            sbml_source: str,
+            show_sbml_warnings: bool = False,
+            from_file: bool = True):
         """Create a new Model instance.
 
         Arguments:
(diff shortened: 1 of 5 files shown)


python/amici/sbml_import.py
```diff
@@ -76,58 +76,59 @@ class SbmlImporter:
 
     """
 
-    def __init__(self, SBMLFile, show_sbml_warnings=False):
-        """Create a new Model instance.
+    def __init__(
```
FFroehlich: can we please the doxygen docstring formating. I know this sucks but doxygen is the only thing that works reliably across multiple languages and none of the python filter that I know support sphinx style documentation (yes I know this sucks but theres not much we can do without investing a ridiculous amount of time)

yannikschaelte: yes, totally makes sense. hope it's doxygen compatible now.

FFroehlich:
```
DOXYGEN failed:
/Users/travis/build/ICB-DCM/AMICI/python/amici/sbml_import.py:156: warning: argument 'Either' of command @param is not found in the argument list of SbmlImporter::__init__(self, str sbml_source, bool show_sbml_warnings=False, bool from_file=True)
/Users/travis/build/ICB-DCM/AMICI/python/amici/sbml_import.py:156: warning: argument 'Or' of command @param is not found in the argument list of SbmlImporter::__init__(self, str sbml_source, bool show_sbml_warnings=False, bool from_file=True)
/Users/travis/build/ICB-DCM/AMICI/python/amici/sbml_import.py:155: warning: The following parameters of amici.sbml_import.SbmlImporter.__init__(self, str sbml_source, bool show_sbml_warnings=False, bool from_file=True) are not documented:
  parameter 'sbml_source'
  parameter 'show_sbml_warnings'
  parameter 'from_file'
/Users/travis/build/ICB-DCM/AMICI/python/amici/sbml_import.py:180: warning: return type of member amici.sbml_import.SbmlImporter.process_document is not documented
```

PR #643, opened 2019-03-07
add pscale and plist to ExpData

FFroehlich:
apparently we need a couple of more options in ExpData 😓

### DIFF
--- a/src/edata.cpp
+++ b/src/edata.cpp
@@ -404,7 +404,7 @@ void ConditionContext::restore()
 {
     // parameter list has to be set before initial state sensitivities
     model->setParameterList(originalParameterList);
-    // parameter scale has to be done befor initial state sensitivities
+    // parameter scale has to be set before initial state sensitivities
     model->setParameterScale(originalScaling);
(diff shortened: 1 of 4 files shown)


src/edata.cpp
```diff
@@ -343,6 +345,21 @@ void ConditionContext::applyCondition(const ExpData *edata)
     if(!edata)
         return;
     
+    // this needs to go first, otherwise nplist will not have the right the
```
dweindl:
```suggestion
    // this needs to go first, otherwise nplist will not have the```

File: src/edata.cpp [hunk 2]
```diff
@@ -385,11 +402,16 @@ void ConditionContext::applyCondition(const ExpData *edata)
 
 void ConditionContext::restore()
 {
+    // parameter list has to be set before initial state sensitivities
+    model->setParameterList(originalParameterList);
+    // parameter scale has to be done befor initial state sensitivities
```
dweindl:
```suggestion
    // parameter scale has to be set before initial state sensitivities
```

PR #648, opened 2019-03-10
Parallel compile, etc.

dweindl:


python/sdist/custom_commands.py
```diff
@@ -0,0 +1,223 @@
...
+            include_dirs=None, debug=0, extra_preargs=None,
+            extra_postargs=None, depends=None):
+    """Parallelized version of distutils.ccompiler.compile"""
+    macros, objects, extra_postargs, pp_opts, build = \
+        self._setup_compile(output_dir, macros, include_dirs, sources,
+                            depends, extra_postargs)
+    cc_args = self._get_cc_args(pp_opts, debug, extra_preargs)
+
+    # parallel compilation
+    num_threads = os.cpu_count()
```
FFroehlich: can we please have an option to limit the used cpus (env variable?).

dweindl: If we assume it may cause trouble, maybe we default to sequential and add `AMICI_PARALLEL_COMPILE=num_threads`

PR #649, opened 2019-03-10
fix critical bug for pysb models

FFroehlich:


FFroehlich:
No idea how this could pass unittests ...

PR #652, opened 2019-03-11
store copy of data pointer in SunMatrixWrapper

FFroehlich:


src/sundials_matrix_wrapper.cpp
```diff
@@ -50,13 +58,20 @@ SUNMatrixWrapper::SUNMatrixWrapper(const SUNMatrixWrapper &A, realtype droptol,
...
+    indexptrs_ptr = SM_INDEXPTRS_S(matrix);
+    indexvals_ptr = SM_INDEXVALS_S(matrix);
 }
 
 SUNMatrixWrapper::SUNMatrixWrapper(SUNMatrix mat) : matrix(mat) {}
 
 SUNMatrixWrapper::~SUNMatrixWrapper() {
     if (matrix)
         SUNMatDestroy(matrix);
+    data_ptr = nullptr;
```
dweindl: Those shouldn't be necessary.

PR #653, opened 2019-03-12
fix failfast

FFroehlich:


FFroehlich:
will add a test for this tomorrow ...

FFroehlich:
> Empty slots will cause trouble in Python, or why do we need that?

Correct.

dweindl:
Can you add a comment on that? Otherwise I might optimize it out soon ;)

FFroehlich:
> Can you add a comment on that? Otherwise I might optimize it out soon ;)

done, also added a testcase that will hopefull prevent you from ignoring the comment :D

PR #655, opened 2019-03-14
Speedup travis

FFroehlich:


FFroehlich:
I think pip installs are currently not cached as they are installed into a virtual environment and thus not detectable in the default directory. We could change that by moving the venv directory out of `build` and not clearing it every time. 

Alternative is to pre-install expensive packages such as `libsbml` at the system level and let the virtual-environment pick them up from there.

dweindl:
> Alternative is to pre-install expensive packages such as libsbml at the system level and let the virtual-environment pick them up from there.

Sounds good

PR #656, opened 2019-03-19
Feature custom llh

yannikschaelte:
Allow to specify different cost functions (not only a normal distribution, the former only thing), normal staying the default.

So far, 6 fixed distributions (normal, laplace, on lin/log/log10 scale) are supported.


python/amici/sbml_import.py
```diff
@@ -782,10 +796,22 @@ def processObservables(self, observables, sigmas):
...
+            if unknown_ids:
+                raise ValueError(
+                    f"Sigma provided for unknown observableIds: "
+                    f"{unknown_ids}.")
+        
+        if noise_distributions is None:
+            noise_distributions = {}
+        else:
+            # Ensure no non-existing observableIds have been specified
+            # (no problem here, bu usually an upstream bug)
```
dweindl:
```suggestion
            # (no problem here, but usually an upstream bug)
```

yannikschaelte:
@FFroehlich Can you have another look? I added a basic test now (can be improved if that is better). Indeed, testing proves useful, because the ``abs`` terms beforehand did not work. Now it works (tested also on a few own models). 
I had to tell sympy that the needed variables are all real-valued to get the correct derivative of abs -> sign. 
But I think that makes sense anyway?

FFroehlich:
Awesome, looks good!

PR #660, opened 2019-03-20
0.10.4 release

FFroehlich:


FFroehlich:
waiting for #656, will update version number after it is merged

PR #664, opened 2019-03-21
Fix missing *_colptrs.cpp and *_rowvals.cpp in CMakeLists.txt generated for new models (Fixes #658)

### LINKED ISSUE #658: *_colptrs.cpp and *_rowvals.cpp missing in CMakeLists.txt generated for new models
... leading to linker errors. I guess we should add a test case for CMake-based compilation of a newly created model.

dweindl:


python/amici/ode_export.py
```diff
@@ -2410,6 +2410,16 @@ def _writeCMakeFile(self):
         sources = [self.modelName + '_' + function + '.cpp '
                    for function in self.functions.keys()
                    if self.functions[function]['body'] is not None]
+
+        # add extra source files for sparse matrices
+        for function in self.functions.keys():
```
FFroehlich: `for function in sparse_functions`?

Issue #668, opened 2019-03-21
Incomplete Rdata


FFroehlich:
`rdata['sres']` is not available for `sensitvity_order=1` and `sensitivity_method=none` when only using newton solver equilibration. Probably needs some additional checks.

PR #671, opened 2019-03-26
only check sensi conv for SteadyStateSensitivityMode::simulationFSA

FFroehlich:


FFroehlich:
evaluating fsxdot many times for large models is completely unfeasible

paulstapor:
Do I get it correctly that `model->getSteadyStateSensitivityMode()` is there to differentiate between Newton and simulation? This is not exactly what is done here, right? The goal of this fix is omitting the convergence check for FSA simulation...
Does it have any obvious drawbacks to use this option, or are we save enough to use it?

FFroehlich:
> Do I get it correctly that `model->getSteadyStateSensitivityMode()` is there to differentiate between Newton and simulation? This is not exactly what is done here, right? The goal of this fix is omitting the convergence check for FSA simulation...
> Does it have any obvious drawbacks to use this option, or are we save enough to use it?

Yes, `model->getSteadyStateSensitivityMode()` changes between FSA (`simulationFSA`) and newton computation (`newtonOnly`) of the steady states. The goal of this commit is to only change the behavior for the `newtonOnly` case.

I may have to have a bit more closer look at this as `solver->getSensitivityMethod() == SensitivityMethod::forward ` should never be true for preequilibration anyways ...

paulstapor:
> I may have to have a bit more closer look at this as `solver->getSensitivityMethod() == SensitivityMethod::forward ` should never be true for preequilibration anyways ...

Why? Technically, it should just work as well...

> The goal of this commit is to only change the behavior for the newtonOnly case.

Ah, indeed, sorry, I overlooked that. And makes sense moreover...

FFroehlich:
I turned out not to be using preequilibration, but equilibration, so everything seems to be working as expected. what reallt needs to be fixed is #668

PR #686, opened 2019-04-12
Make dJydy sparse

### LINKED ISSUE #685: Non-finite states result in NaN gradients even if no measurement is concerned with this state
NaN in x propagates to dydx. Multiplication with 0 in djydy results in NaN in fdJydx.

-> Turn djydy into sparse matrix, so the respective entries of dydx will be ignored during multiplication (@FFroehlich )

dweindl:
Make dJydy sparse which should prevent NaN gradient in case NaN output parameters of unobserved observables (Closes #685)

- [x] Make dJydy sparse
- [x] Check dimensions for matrix vector multiplications (use gsl::span)
- [x] Regression test
- [x] Updaate fdJyd*
- [x] Handle matlab-generated models as before
- [x] Cleanup


FFroehlich:
👍

PR #687, opened 2019-04-15
Fix events for DEA models

FFroehlich:
Will add unittests.

Should probably add documentation that we assume events to be "impulse free" in a DAE context.


tests/cpputest/testfunctions.cpp
```diff
@@ -156,12 +155,21 @@ void checkEqualArrayStrided(const double *expected, const double *actual, int le
 
 void verifyReturnData(std::string const& hdffile, std::string const& resultPath,
                       const ReturnData *rdata, const Model *model, double atol, double rtol) {
+    
     CHECK_FALSE(rdata == nullptr);
+    
+    if(!hdf5::locationExists(hdffile, resultPath)) {
+        printf("WARNING: skipping checks for %s as no results data is "
```
dweindl: Potentially dangerous :)

PR #692, opened 2019-04-30
Fix(python) Fix unset noise distribution during SBML import for autom…

### LINKED ISSUE #691: Use amici without observables fails recently

Be able to compile and load a model without observables (i.e. just the system dynamics). For the exact details, please refer to @philippstaedter if unclear.

Fails if None passed.

image

**Amici version**
Error seems to be new in the latest amici version (0.10.6).

dweindl:
…atically generated observables if none are passed by the user (Fixes  #691)

- Add test for model wrapping
- Allow SBML import from libsbml.Model

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -86,9 +86,10 @@ class SbmlImporter:
 
         Arguments:
 
-            sbml_source:  Either a path to SBML file where the model is specified.
-            Or a model string as created by  sbml.sbmlWriter().writeSBMLToString().
-            @type string
+            sbml_source: Either a path to SBML file where the model is
+                specified, or a model string as created by
+                sbml.sbmlWriter().writeSBMLToString() or an instance of
+                libsbml.Model. @type str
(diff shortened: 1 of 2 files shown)


python/amici/sbml_import.py
```diff
@@ -86,9 +86,10 @@ def __init__(
 
         Arguments:
 
-            sbml_source:  Either a path to SBML file where the model is specified.
-            Or a model string as created by  sbml.sbmlWriter().writeSBMLToString().
-            @type string
+            sbml_source: Either a path to SBML file where the model is
```
yannikschaelte: on second thoughts, the argument ``from_file`` is not really necessary, because file path and sbml code are quite different in kind even though both are strings.

dweindl: Agreed. (Unless somebody saves his model with the name `<?xml`...) 
I didn't want to break anything unnecessarily though. Let's keep that for some future cleanup.

PR #699, opened 2019-05-08
Feature timing, Closes #524

### LINKED ISSUE #524: Get simulation time as output of solver
If we want to study the influence of the system size on computation time, it might be good to have computation time as an additional output of the C++ code of AMICI. This would be more rigorous for small-scale models than using tic, toc in Matlab or similar python functionality around solver calls. For the Newton solver, we already have that.

Should I introduce that for the ODE solve as well (it would hep me, at least)?
And if so, how to handle the preequilibration then? Get the time after preequilibration only, or also before?

paulstapor:
Added cputime for forward and backward solve to the output of Amici, to get a feeling for the overhead from python to C++, and to see how the pure computation time scales with the size of the model (for comparison: theory vs. practice).

### DIFF
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -43,8 +43,6 @@ Solver::Solver(const Solver &other) : Solver() {
     iter = other.iter;
     stldet = other.stldet;
     ordering = other.ordering;
-    cpu_time = other.cpu_time;
-    cpu_timeB = other.cpu_timeB;
 }
 
 int Solver::run(const realtype tout, ReturnData *rdata) const {
@@ -183,8 +181,6 @@ void Solver::getDiagnosis(const int it, ReturnData *rdata) const {
 
         getLastOrder(solverMemory.get(), &rdata->order[it]);
     }
-    
-    rdata->cpu_time = this->cpu_time;
 }
 
 void Solver::getDiagnosisB(const int it, ReturnData *rdata,
@@ -204,8 +200,6 @@ void Solver::getDiagnosisB(const int it, ReturnData *rdata,
         getNumNonlinSolvConvFails(solverMemoryB.at(which).get(), &number);
         rdata->numnonlinsolvconvfailsB[it] = number;
     }
-    
-    rdata->cpu_timeB = this->cpu_timeB;
 }
 
 void Solver::initializeLinearSolver(const Model *model) const {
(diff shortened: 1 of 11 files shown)


FFroehlich:
Had a brief look at the code. Please make the cpu times private `Solver` members and add respective getter functions. Then add extraction of time + writing to rdata to forward problem.

I am currently in the process of refactoring the whole handling of rdata and the goal is no never have solver/model interact with rdata, which increases modularity/interoperability of these classes and adds some additional flexibility in terms of when and whether data is written to rdata.

dweindl:
And please add `Closes #524` to the pull request description or commit message

src/solver.cpp
```diff
@@ -54,7 +54,7 @@ int Solver::run(const realtype tout, ReturnData *rdata) const {
     } else {
         status = solve(tout, AMICI_NORMAL);
     }
-    rdata->cpu_time +=(double)((clock() - starttime) * 1000) / CLOCKS_PER_SEC;
+    incrementCpuTime((realtype)((clock() - starttime) * 1000) / CLOCKS_PER_SEC);
```
FFroehlich: its a private member, you should still be able to modify it without a setter routine. I don't think we want a setter routine as the `Solver` class itself should be the only one that modifies those values.

paulstapor: okay, I'll do that. Thanks!
Other question: How about the way, it's currently implemented in forward or backwardproblem.cpp? Would you prefer incrementing the rdata-member after every run, or once at the very end of the routine (as it is now)?

FFroehlich: > Other question: How about the way, it's currently implemented in forward or backwardproblem.cpp? Would you prefer incrementing the rdata-member after every run, or once at the very end of the routine (as it is now)?

It's fine as it is now, as it is consistent with how other things are handled at the moment. In the long run I want to change this to be all done in a single function that can be called independently of forward/backwardproblem, but one thing at a time.

paulstapor:
@FFroehlich : As I'm (hopefully) done so far but have no permission for merging: Can you please merge, once the checks have passed?

FFroehlich:
> @FFroehlich : As I'm (hopefully) done so far but have no permission for merging: Can you please merge, once the checks have passed?

I don't see why you wouldn't have permission, we should fix that.

paulstapor:
> > @FFroehlich : As I'm (hopefully) done so far but have no permission for merging: Can you please merge, once the checks have passed?
> 
> I don't see why you wouldn't have permission, we should fix that.

At least, for the last PRs I didn't have it... So I guess I don't have it now...
So, alternatively, if I have permission, then I'd merge it...

PR #712, opened 2019-06-24
Fix initial amounts (fix #674)

### LINKED ISSUE #674: initialConcentration set to zero if not defined in SBML model
We had the case, where only the `initialAmount` was set in the SBML model but not `initialConcentration`. AMICI then set the `initialConcentration` to zero. Maybe we should give a warning in that case.

paulstapor:
I use a quick check whether or not only substance units are used for a given species. If this is the case, no initialConcentration will be defined, but an initialAmount. This means that the corresponding field has to be used.
This was already implemented but the if-clause which checks for that simply did not cover this case...

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -362,14 +362,15 @@ class SbmlImporter:
         concentrations = [spec.getInitialConcentration() for spec in species]
         amounts = [spec.getInitialAmount() for spec in species]
 
-        if all(self.speciesHasOnlySubstanceUnits):
-            warnings.warn('SBML file declares no concentrations, but only '
-                          'substance units. Using amounts instea of '
-                          'concentrations.')
-            concentrations = amounts.copy()
+#        if all(self.speciesHasOnlySubstanceUnits):
+#            warnings.warn('SBML file declares no concentrations, but only '
+ #                         'substance units. Using amounts instea of '
+ ##                         'concentrations.')
+ #           concentrations = amounts.copy()
 
         def getSpeciesInitial(index, conc):
-            if not math.isnan(conc):
+            if not self.speciesHasOnlySubstanceUnits[index] \
+                    or not math.isnan(conc):
                 return sp.sympify(conc)
             if not math.isnan(amounts[index]):
                 return \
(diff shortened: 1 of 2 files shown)


python/amici/sbml_import.py
```diff
@@ -363,7 +363,8 @@ def processSpecies(self):
         amounts = [spec.getInitialAmount() for spec in species]
 
         def getSpeciesInitial(index, conc):
-            if not math.isnan(conc):
+            if not (self.speciesHasOnlySubstanceUnits[index] or math.isnan(
```
dweindl: If speciesHasOnlySubstanceUnits is true, the initial amount should be set (and used), not the initial concentration. Is speciesHasOnlySubstanceUnits=True in combination with intialconcentration instead of initialamount valid SBML?

With change will use the nan, if the concentration is nan. I am not sure if this is what we want.

paulstapor: In valid SBML, you can not use 
`speciesHasOnlySubstanceUnits = "True"` and set `intialConcentration`.
Already used nan, if the concentration was nan. And, moreover, if my understanding of the code is correct, the initial concentration will never be set to nan but rather to 0.0 ...
Which is the main source of our problems...

dweindl: Sorry, missed the parentheses after `not`.

paulstapor: The main problem here is actually the libsbml-api, I think... the corresponding function getInitialConcentrations() is not doing the correct job... Yet, we have to handle it somehow...
This is the best quick-fix I could come up with that really helps in the situations we have, does not change the code for situations in which everything works and does not try to solve problems we currently don't have...

dweindl: > getInitialConcentrations() is not doing the correct job

What does it do and what should it do instead?

paulstapor: To my understanding, it should return `nan`, if no initialConcentration is defined, but it returns `0.0`. That's misleading...

dweindl: Should first check `species.isSetInitialConcentration()` then.

paulstapor: corrected it

PR #717, opened 2019-06-26
Allow setting model parameter from dict

### LINKED ISSUE #715: Set model parameters from dict
As a user i want to be able to set model parameters by passing a parameter_id => value dictionary. Ideally both in Python and C++ (std::map).

dweindl:
Closes #715

### DIFF
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -320,6 +320,19 @@ void Model::setParameters(const std::vector<realtype> &p) {
     unscaleParameters(originalParameters, pscale, unscaledParameters);
 }
 
+void Model::setParameters(const std::map<std::string, realtype> &p,
+                          bool ignoreErrors)
+{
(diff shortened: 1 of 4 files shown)


src/model.cpp
```diff
@@ -320,6 +320,19 @@ void Model::setParameters(const std::vector<realtype> &p) {
     unscaleParameters(originalParameters, pscale, unscaledParameters);
 }
 
+void Model::setParameters(const std::map<std::string, realtype> &p,
```
FFroehlich: `setParametersById`? Also implement `setParametersByName`?

dweindl: Not using it. But maybe good to add for symmetry...

dweindl:
> Can we have some testcases for the introduced functionality (both python+cpp)?

Added for Python. This implies that it is working for C++.

PR #725, opened 2019-07-03
Add AMICI interface overview / workflow figure and show in README

dweindl:


dweindl:
> Conversion from SBML to c++ is done via sympy, not numpy right?

Absolutely. Oops.

PR #740, opened 2019-07-26
update with publication from FOSBE2019

paulstapor:
Added the publications from this years FOSBE conferece


documentation/recreate_reference_list.py
```diff
@@ -52,7 +52,7 @@ def main():
     num_total = sum(map(len, by_year.values()))
     with open(outfile, 'w') as f:
         f.write('# References\n\n')
-        f.write('List of publications using AMICI. '
```
dweindl: I'd rather leave that out, since at times, we also have preprints in the list...

PR #741, opened 2019-07-26
Update BLAS and HDF5 detection from Python

### LINKED ISSUE #736: Smarter BLAS detection during python core/model build
~For finding a usable BLAS, check~
```
import numpy.distutils.system_info as sysinfo
sysinfo.get_info('blas')
```

Also consider ATLAS if no CBLAS is available

dweindl:


python/amici/setuptools.py
```diff
@@ -110,11 +120,15 @@ def getHdf5Config():
...
-            os.path.join(hdf5_library_dir_hint, 'libhdf5.a'))
+        # check for static or shared library
+        for lib_filename in ['libhdf5.a', 'libhdf5.so']:
+            hdf5_library_dir_found = os.path.isfile(
+                os.path.join(hdf5_library_dir_hint, lib_filename))
+            if hdf5_library_dir_found:
+                print(f'{lib_filename} found in {hdf5_library_dir_hint}')
+                h5pkgcfg['library_dirs'] = [hdf5_library_dir_hint]
+                break
         if hdf5_library_dir_found:
```
FFroehlich: This looks superfluous.

dweindl: The second break you mean? The first break only exists the inner loop. This may cause `hdf5_library_dir_found` (used below) to have the wrong value. Had the problem one commit before :).

FFroehlich: Oooh, I see. Might be worthwhile to add a comment then ;)

PR #777, opened 2019-08-01
Small docu fixes and assignmentRules2obervables() fix

MerktSimon:
1) Pythondoku (http://icb-dcm.github.io/AMICI/python_interface.html): "sigma" under "Model Compilation" is now fixed to be "sigmas" and under "Observables" now "filter" fixed to be "filter_function"
2) In function "assignmentRules2observables()" filter_function should be called with parameter_id instead of p


python/amici/sbml_import.py
```diff
@@ -1262,7 +1262,7 @@ def assignmentRules2observables(sbml_model,
     observables = {}
     for p in sbml_model.getListOfParameters():
         parameter_id = p.getId()
-        if filter_function(p):
+        if filter_function(parameter_id):
```
FFroehlich: I believe this was intentional and that the documentation was not updated. The idea is to enable filtering based on both name and id and that the user needs to include the call got `getId()` or `getName()`. @dweindl might remember more details.

dweindl: That is correct (or to filter by any other species attribute).

dweindl: This part of the documentation needs some update. Didn't remember that was there. Meanwhile PEtab should be used for that...

MerktSimon: I undid the changes to the function and corrected the use of it in the python docu.

PR #778, opened 2019-08-06
More informative output during swig detection, fix typos

dweindl:
More informative output during swig detection, fix typos

### DIFF
--- a/python/amici/setuptools.py
+++ b/python/amici/setuptools.py
@@ -203,23 +203,31 @@ def generateSwigInterfaceFiles():
     """Compile the swig python interface to amici
     """
     swig_outdir = '%s/amici' % os.path.abspath(os.getcwd())
-    swig_cmd = find_swig()
-    sp = subprocess.run([swig_cmd,
-                         '-c++',
-                         '-python',
-                         '-Iamici/swig', '-Iamici/include',
-                         '-DAMICI_SWIG_WITHOUT_HDF5',
-                         '-outdir', swig_outdir,
-                         '-o', 'amici/amici_wrap_without_hdf5.cxx',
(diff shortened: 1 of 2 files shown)


python/amici/setuptools.py
```diff
@@ -203,23 +203,31 @@ def generateSwigInterfaceFiles():
...
+
+    # Swig AMICI interface with HDF5 dependency
+    swig_cmd = [swig_exe,
+                '-c++',
+                '-python',
+                '-Iamici/swig', '-Iamici/include',
+                '-outdir', swig_outdir,
+                '-o', 'amici/amici_wrap.cxx',
+                'amici/swig/amici.i']
+    print(f"Running SWIG: {' '.join(swig_cmd)}")
```
FFroehlich: can we explicitetly state with/without HDF5 in the printed messages. Might avoid some confusion why swig is run twice.

dweindl: it's part of the output:
```
    Running SWIG: 'swig3.0 -c++ -python -Iamici/swig -Iamici/include -DAMICI_SWIG_WITHOUT_HDF5 -outdir /home/dweindl/src/AMICI-devel/python/sdist/amici -o amici/amici_wrap_without_hdf5.cxx amici/swig/amici.i'
    Running SWIG: 'swig3.0 -c++ -python -Iamici/swig -Iamici/include -outdir /home/dweindl/src/AMICI-devel/python/sdist/amici -o amici/amici_wrap.cxx amici/swig/amici.i'
``` 
Should be clear that it's doing different things, not?

FFroehlich: true, should be okay!

Issue #774, opened 2019-07-31
Incorrect results in SBML semantic test case testId=998


dweindl:
Needs further investigations. Maybe just about error tolerances.
```
======================================================================
FAIL: test_sbml_testsuite (__main__.TestAmiciSBMLTestSuite) (testId=998)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./tests/testSBMLSuite.py", line 34, in test_sbml_testsuite
    self.run_sbml_testsuite_case(get_test_str(testId))
  File "./tests/testSBMLSuite.py", line 95, in run_sbml_testsuite_case
    np.isclose(simulated_x, test_x, atol, rtol).all()
AssertionError: False is not true
```

paulstapor:
Unfortunately not a matter of tolerances: Amici Simulation fails for this model, since an `Inf` was encountered in the right hand side... Have to look into that more closely

dweindl:
I shortly checked yesterday, it's about initial amount and initial concentrations. Check the description of the test case before diving too deep into something else.

paulstapor:
yes, for some reason, it does not read the initial concentration correctly...
I don't really get it right now. But I will try to fix it tomorrow. Today ends now...

dweindl:
Closed in #789

Issue #773, opened 2019-07-31
Booleans in kinetic laws not handled correctly during SBML import


dweindl:
e.g.
```
======================================================================
ERROR: test_sbml_testsuite (__main__.TestAmiciSBMLTestSuite) (testId=1288)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./tests/testSBMLSuite.py", line 34, in test_sbml_testsuite
    self.run_sbml_testsuite_case(get_test_str(testId))
  File "./tests/testSBMLSuite.py", line 45, in run_sbml_testsuite_case
    model, solver, wrapper = compile_model(current_test_path, test_id)
  File "./tests/testSBMLSuite.py", line 147, in compile_model
    wrapper.sbml2amici(model_name, output_dir=model_dir)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/sbml_import.py", line 256, in sbml2amici
    exporter.generateModelCode()
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 1901, in generateModelCode
    self._generateCCode()
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 1942, in _generateCCode
    self._writeFunctionFile(function)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 2107, in _writeFunctionFile
    symbol = self.model.eq(function)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 1081, in eq
    self._compute_equation(name)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 1344, in _compute_equation
    self._total_derivative(**args)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 1592, in _total_derivative
    dxdz = self.sym_or_eq(name, dxdz_name)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 1620, in sym_or_eq
    return self.sym(varname)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 1037, in sym
    self._generateSymbol(name)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/ode_export.py", line 1216, in _generateSymbol
[...]

Issue #771, opened 2019-07-31
Error in handling of SBML piecewise(): Second argument must be a Boolean, not `xor`


dweindl:
```
======================================================================
ERROR: test_sbml_testsuite (__main__.TestAmiciSBMLTestSuite) (testId=201)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./tests/testSBMLSuite.py", line 34, in test_sbml_testsuite
    self.run_sbml_testsuite_case(get_test_str(testId))
  File "./tests/testSBMLSuite.py", line 45, in run_sbml_testsuite_case
    model, solver, wrapper = compile_model(current_test_path, test_id)
  File "./tests/testSBMLSuite.py", line 147, in compile_model
    wrapper.sbml2amici(model_name, output_dir=model_dir)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/sbml_import.py", line 242, in sbml2amici
    self.processSBML(constantParameters)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/sbml_import.py", line 284, in processSBML
    self.processReactions()
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/sbml_import.py", line 666, in processReactions
    symMath = _parse_special_functions(symMath)
  File "/home/icb/daniel.weindl/src/AMICI/python/sdist/amici/sbml_import.py", line 1218, in _parse_special_functions
    return sp.Piecewise(*grouper(args, 2, True))
  File "/home/icb/daniel.weindl/src/AMICI/build/venv/lib64/python3.6/site-packages/sympy/functions/elementary/piecewise.py", line 134, in __new__
    pair = ExprCondPair(*getattr(ec, 'args', ec))
  File "/home/icb/daniel.weindl/src/AMICI/build/venv/lib64/python3.6/site-packages/sympy/functions/elementary/piecewise.py", line 35, in __new__
    not `%s`''' % func_name(cond)))
TypeError:
Second argument must be a Boolean, not `xor`
```
and
```
======================================================================
ERROR: test_sbml_testsuite (__main__.TestAmiciSBMLTestSuite) (testId=279)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./tests/testSBMLSuite.py", line 34, in test_sbml_testsuite
    self.run_sbml_testsuite_case(get_test_str(testId))
  File "./tests/testSBMLSuite.py", line 45, in run_sbml_testsuite_case
[...]

PR #789, opened 2019-08-27
Run SBML test suite using github actions instead of travisCI

### LINKED ISSUE #571: SBML Testsuite is taking too long
Split testcases in 2-3 batches and execute in parallel jobs? Would be nice if we can skip them in regular builds and skip os x builds in the cron setting.

### LINKED ISSUE #731: Running SBML test suite as travis-ci cron job failing since months?
See https://travis-ci.com/ICB-DCM/AMICI/builds

To be checked if problem persists with updated `master`.

### LINKED ISSUE #775: Allow running individual tests from SBML test suite via CLI
Should allow for something like:
`testSBMLSuite.py 1,3-5,8`

dweindl:
* Run SBML test suite using github actions instead of travisCI (Closes #571)
* Allow running individual tests from SBML test suite via CLI (Closes #775)
* Fix SBML test suite issues (Closes #731)
* [x] On merge to master
* [x] Remove travis cron job?
* [x] #771 
* [x] #773 
* [x] #774 

Current run time for test suite is about 1h10


dweindl:
Looks quite usable. Currently test output is delayed by like 10min, not sure why.

dweindl:
Seems like the time limits are currently not enforced. One job running for >10h already.

dweindl:
Travis failing because http://www.valgrind.org/ seems to down so that downloading valgrind fails.

FFroehlich:
Ideally I would like to have this as required check for merging into master.

dweindl:
> Ideally I would like to have this as required check for merging into master.

For that to make sense, we should keep track of the tests we were able to pass. Only fail the github check upon regression, but not when known-to-fail tests fail.

FFroehlich:
There should not be any known-to-fail tests (which was true a couple of months ago and only changed with recent modifications).
The only known to fail case involves NaNs in the jacobian and is hardcoded in the testsuite script.

FFroehlich:
Is it possible to also submit results to the testsuite website on release?

dweindl:
> There should not be any known-to-fail tests [...]

Okay, that was imprecise. I meant known-to-not-pass. But true, skipped test will be ignored. The reason for failure were only the two issues i wrote in the top most. One of them being fixed now, the other one waiting for @paulstapor .

> Is it possible to also submit results to the testsuite website on release?

To where exactly? Can be done, sure. Maybe open a respective issue.

dweindl:
@paulstapor #774 should also be fixed here (you were assigned)

FFroehlich:
fixed #774, but looks like someone broke `_parse_logical_operators()`

PR #794, opened 2019-08-30
Fix piecewise

paulstapor:
Add support for logical operations in the boolean argument of "sympy.piecewise()"

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -8,11 +8,10 @@ import itertools as itt
 import warnings
 from typing import Dict, Union, List, Callable, Any, Iterable
 
+
 from .ode_export import ODEExporter, ODEModel
 from . import has_clibs
 
-from sympy.logic.boolalg import BooleanTrue as spTrue
-from sympy.logic.boolalg import BooleanFalse as spFalse
 
 class SBMLException(Exception):
     pass
@@ -402,8 +401,8 @@ class SbmlImporter:
                 index = species_ids.index(
                         initial_assignment.getId()
                     )
-                symMath = sp.sympify(
-                    sbml.formulaToL3String(initial_assignment.getMath()),
+                symMath = sp.sympify(_parse_logical_operators_in_math_string(
+                    sbml.formulaToL3String(initial_assignment.getMath())),
                     locals=self.local_symbols
                 )
                 if symMath is not None:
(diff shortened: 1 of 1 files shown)


python/amici/sbml_import.py
```diff
@@ -1188,41 +1192,53 @@ def _check_unsupported_functions(sym, expression_type, full_sym=None):
...
+        Arguments:
+            math_str: str with mathematical expression
+
+        Returns:
+            math_str: parsed math_str
+
+        Raises:
+    """
+    if ' xor(' in math_str or  'Xor(' in math_str:
+        raise SBMLException('Logical statements including Xor() are '
```
dweindl: Could have added as a single line...

python/amici/sbml_import.py
```diff
@@ -1188,41 +1192,53 @@ def _check_unsupported_functions(sym, expression_type, full_sym=None):
...
 
         Raises:
     """
-    args = tuple(_parse_special_functions(arg, False) for arg in sym._args)
+    args = tuple(_parse_special_functions(arg) for arg in sym._args)
 
+    # Do we have piecewise expressions?
     if sym.__class__.__name__ == 'abs':
         return sp.Abs(sym._args[0])
-    elif sym.__class__.__name__ == 'xor':
```
dweindl: ... here

dweindl: I'd propose to just add the exception line in a new commit on (an up-to-date) `test_action`. That will be the least painfull :)

paulstapor:
As mentioned: This one is branched from `test_action`, instead of develop, as discussed in #771. The actual changes I made should be only in sbml_import.py. Please review all other changes carefully with respect to that.

General question: Does this actually now resolve #771, or rather not?
In any case, it should at the same time fix the problem of having "and"s or "or"s in the piecewise function...

dweindl:
> Does this actually now resolve #771, or rather not?

Yes it does. But maybe you can add a new issue for re-enabling Xor once that is resolved in sympy (referencing your report there).

paulstapor:
> > Does this actually now resolve #771, or rather not?
> 
> Yes it does. But maybe you can add a new issue for re-enabling Xor once that is resolved in sympy (referencing your report there).

On which branch did you implement it? I branched away from test_action around noon today...

dweindl:
> On which branch did you implement it? I branched away from test_action around noon today...

If this refers to 
> What happened to my boolean-handling code?

Then here, and earlier:
https://github.com/ICB-DCM/AMICI/pull/789/commits/e618a6ad9d60c81be4fcd6ae88ca108bd8767198

paulstapor:
Damn, I know what probably happened. I will restore it...

paulstapor:
Now all should be fine again... Sorry for temporary messing up you previous code changes.

Issue #791, opened 2019-08-29
Submit testsuite results to testsuite db


FFroehlich:
http://sbml.org/Facilities/Database/Submission/Create

Might need to investigate whether there is a better API that does not require login

dweindl:
Ah, wasn't aware of that. 

If no better API, could start with automatically sending the test results for manual upload.

dweindl:
Some of the test results require additional parameters or compartment volumes not yet included. The required columns for test results need to be determined based on the expected result file of the respective test case.

dweindl:
Not sure if it is worth investing time here. The database is rather outdated. The most recent published uploads are from mid of 2018. Making AMICI results public does not seem to work. Too bad... liked the idea...

FFroehlich:
Probably worthwhile to post an issue about this on the testsuite git, mantainers were pretty positive about it at the last COMBINE.

dweindl:
Hm... maybe not... just [found](http://sbml.org/Facilities/Online_SBML_Test_Suite): 

> The Online SBML Test Suite has not been updated in some time, and will be retired  on September 1, 2016.  The rest of the SBML Test Suite, including the  test cases, the online database, and the standalone SBML Test Runner,  will continue to be maintained.

FFroehlich:
Managed to make current results (0.10.20) public http://sbml.org/Facilities/Database/Submission/Details/256

dweindl:
> Managed to make current results (0.10.20) public http://sbml.org/Facilities/Database/Submission/Details/256

1. How?

2. Really? :) 
> Server Error
404 - File or directory not found.
The resource you are looking for might have been removed, had its name changed, or is temporarily unavailable.

FFroehlich:
Ah looks like it didn't work and requires some kind of manual curation on their end, so lets close this for now.

dweindl:
yeah, i also clicked "make public" a couple of times, but nothing happened

FFroehlich:
oh i was able to make public and add email + description, but I suppose there is an additional curation step involved.

PR #798, opened 2019-08-31
Create SBML test suite result files for upload to http://sbml.org/Fac…

dweindl:
…ilities/Database/

... for manual upload (#791)


dweindl:
Some of the test results require additional parameters or species not yet included. The required columns for test results need to be determined based on the expected result file of the respective test case.

Issue #716, opened 2019-06-26
Python package installation from python/sdist/ broken


dweindl:
Only the development installation with `pip -e` or via sdist works. This is due to  the symlink to README.md. Needs to be cleaned up somehow.

dweindl:
Without this being addressed, Python package installations from the github repository won't work for Windows users. (Does not affect installation of source distributions created elsewhere.)

dweindl:
Fixing this also for Windows users is a bit tricky. Problem is that Python setuptools expects the same directory structure in the repository as we want to have in the installation directory. This is no problem for Python-only projects, but causes trouble here if we want to separate the different interfaces. Same problem with the current implementation of enscons.

Options:
1. keeping things as they are (symlinks to create required python package structure)
  * ++ separation of interfaces
  * ++ supports editable installation
  * -- won't work for windows
2. creating the required directory structure from cmake in a temporary directory
  * ++ separation of interfaces
  * ++ works for windows
  * -- requires cmake (minor)
  * -- breaks editable pip installation
3. rearrange directory structure for setuptools
  * ++ works on windows
  * ++ supports editable installation
  * -- is a big mess, no separation of interfaces
4. modifying setuptools or enscons behaviour to allow specifying destination of files in package
  * -- lot's of work
  * -- breaks editable pip installation

Any ideas who to handle that in a nicer way?

paulstapor:
What do you precisely mean by "separation of interfaces"?

dweindl:
> What do you precisely mean by "separation of interfaces"?

Keeping the C++/Matlab/Python/.. files in separate folders instead of pressing everything in a shape working for Python setuptools.

dweindl:
Update: The problem is not only the readme symlink which would be easy to fix, but it applies to many other files, especially everything below `ThirdParty/`. 

This is intended behavior of `pip`. So "wontfix". The way to go is via `sdist`.

dweindl:
Actually `cd python/sdist && python setup.py install` works, only `cd python/sdist && pip install .` fails.

dweindl:
> only `cd python/sdist && pip install .` fails.

Seems to work meanwhile. (Tested on Ubuntu 24.04, python 3.12, pip 24.2, setuptools-75.1.0).

PR #807, opened 2019-09-18
Fixes 803 805

FFroehlich:


scripts/run-cppcheck.sh
```diff
@@ -22,6 +22,9 @@ mv cppcheck_tmp.txt cppcheck.txt
 grep -v "'AmiVectorArray' does not have a operator=" cppcheck.txt > cppcheck_tmp.txt
 mv cppcheck_tmp.txt cppcheck.txt
 
+grep -v "Obsolete function 'alloca' called." cppcheck.txt > cppcheck_tmp.txt
```
dweindl: Shall we just remove the `alloca`s?

dweindl: Alternatively use `--exitcode-suppressions=` (or less preferred, `--suppress=`), which is cleaner and probably more stable

FFroehlich:
> Is there a timeline for committing this fix to master? That would be very useful for me, because the I could install PyAMICI.

@thomassligon I guess it will take a little bit until it makes its way into master. Can you try whether direct installation from github works for you?

```
pip3 install git+https://github.com/ICB-DCM/AMICI@fix_803_805#egg=amici\&subdirectory=python/sdist
```

scripts/run-cppcheck.sh
dweindl: @FFroehlich Do we have any evidence that using std::vector will have a noticeable performance impact? Otherwise I'd just change it to that. With clang, g++, intel compilers we could use variable length arrays, but this will cause trouble with msvc++...

thomassligon:
Could it be that the command is trying to find setup.py in the root directory of AMICI? (Maybe it's just the way it copies files to a temp directory.)
After removing the backslash from directly before the ampersand in the command, I got this far:

  C:\data\Tom\Research\AMICI\python\sdist>pip3 install git+https://github.com/ICB-    DCM/AMICI@fix_803_805#egg=amici&subdirectory=python/sdist
Collecting amici from git+https://github.com/ICB-DCM/AMICI@fix_803_805#egg=amici
  Cloning https://github.com/ICB-DCM/AMICI (to revision fix_803_805) to c:\users\thoma\appdata\local\temp\pip-install-jml89dq5\amici
  Running command git clone -q https://github.com/ICB-DCM/AMICI 'C:\Users\thoma\AppData\Local\Temp\pip-install-jml89dq5\amici'
  Running command git checkout -b fix_803_805 --track origin/fix_803_805
  Branch 'fix_803_805' set up to track remote branch 'fix_803_805' from 'origin'.
  Switched to a new branch 'fix_803_805'
    ERROR: Command errored out with exit status 1:
     command: 'c:\program files (x86)\microsoft visual studio\shared\python37_64\python.exe' -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\thoma\\AppData\\Local\\Temp\\pip-install-jml89dq5\\amici\\setup.py'"'"'; __file__='"'"'C:\\Users\\thoma\\AppData\\Local\\Temp\\pip-install-jml89dq5\\amici\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base pip-egg-info
         cwd: C:\Users\thoma\AppData\Local\Temp\pip-install-jml89dq5\amici\
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "c:\program files (x86)\microsoft visual studio\shared\python37_64\lib\tokenize.py", line 447, in open
        buffer = _builtin_open(filename, 'rb')
    FileNotFoundError: [Errno 2] No such file or directory:
  'C:\\Users\\thoma\\AppData\\Local\\Temp\\pip-install-jml89dq5\\amici\\setup.py'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
'subdirectory' is not recognized as an internal or external command,
operable program or batch file.

dweindl:
> Could it be that the command is trying to find setup.py in the root directory of AMICI? [...]

Can you please rerun it with `-v` for more detailed information?
First guess is that this is related to #716.

thomassligon:
@dweindl OK, I ran it with -v, but the output is exactly the same.

dweindl:
> @dweindl OK, I ran it with -v, but the output is exactly the same.

Okay. That was unexpected.

I created a source distribution package from this branch. You can try installing that with `pip3 install amici-0.10.11-dirty.tar.gz`. 
[amici-0.10.11-dirty.tar.gz](https://github.com/ICB-DCM/AMICI/files/3658000/amici-0.10.11-dirty.tar.gz)

Quite sure #716 will have to be addressed until you are able to install AMICI from the repository on Windows. Put as soon as these changes here are on pypi, this shouldn't be a critical limitation.

thomassligon:
This looks much better. I downloaded the tar and put it in my local AMICI at AMICI/python/sdist, then ran the command
pip3 install amici-0.10.11-dirty.tar.gz
from that directory. 21 files are compiled without error, including solver_cvodes.cpp. Then I get new problems in vector.cpp.
    amici\src\vector.cpp(27): error C2039: 'negate': is not a member of 'std'
    amici/ThirdParty/gsl/gsl/gsl-lite.hpp(2837): note: see declaration of 'std'
    amici\src\vector.cpp(27): error C2065: 'negate': undeclared identifier
    amici\src\vector.cpp(27): error C2275: 'amici::realtype': illegal use of this type as an expression
    amici/include\amici/defines.h(54): note: see declaration of 'amici::realtype'
    amici\src\vector.cpp(27): error C2059: syntax error: ')'
    amici\src\vector.cpp(120): warning C4267: 'initializing': conversion from 'size_t' to 'int', possible loss of data
    error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Community\\VC\\Tools\\MSVC\\14.22.27905\\bin\\HostX86\\x64\\cl.exe' failed with exit status 2
Here are lines 25-28 of AMICI/src/vector.cpp:
void AmiVector::minus() {
    std::transform(vec.begin(), vec.end(),
                   vec.begin(), std::negate<realtype>());
}
Could it be that I am making a mistake by running this from my AMICI directory? Maybe I have the wrong include file somewhere.
For example, AMICI/include/amici has a vector.h that manipulates std, but I don’t see a negate…

Tom
(Dr. Thomas S. Ligon)
Tom@thomassligon.info<mailto:Tom@thomassligon.info>
Frohnloher Str. 6a
81475 Muenchen
Germany
Tel. +49(89)74575075

dweindl:
> This looks much better.

Great. Can you please retry with [amici-0.10.11.tar.gz](https://github.com/ICB-DCM/AMICI/files/3658474/amici-0.10.11.tar.gz). I hope I fixed that.

Shouldn't matter from where you run that or where you put that file.

scripts/run-cppcheck.sh
FFroehlich: I believe there was a noticable performance impact when considering second order forward sensitivities for the Swameye model.

dweindl: Then we stay with `alloca`, I guess

thomassligon:
With this fix, all AMICI cpp files compile without error (but with some warnings).
Now there is a problem building sundials:
    cl : Command line error D8021 : invalid numeric argument '/Wno-misleading-indentation'
Unfortunately, this one is an error, not just a warning. The basic problem is that compiler options (command-line arguments) are not defined in the ISO C++ standard. They are always compiler specific.
We need to make one change to fix the error, and it would be advisable to make a second change to fix a warning.

First, replace -Wno-misleading-indentation by nothing. It seems to be wrong on GCC and, if anything, just suppresses a warning. Alternatively, it could be used on GCC, but replaced by nothing on MSVC (Windows).
GCC compiler options
https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html
This defines -Wmisleading-indentation but not -Wno-misleading-indentation.

Second, continue to use -std=c++11 (or a later version) on GCC, and use /std:c++14 or /std:c++latest on MSVC (Windows).
https://docs.microsoft.com/en-us/cpp/build/reference/compiler-options?view=vs-2019
https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=vs-2019

dweindl:
> With this fix, all AMICI cpp files compile without error (but with some warnings).

Great.

> Now there is a problem building sundials:
> cl : Command line error D8021 : invalid numeric argument '/Wno-misleading-indentation'
> Unfortunately, this one is an error, not just a warning. The basic problem is that compiler options (command-line arguments) are not defined in the ISO C++ standard. They are always compiler specific.

Yes... :-/

> We need to make one change to fix the error, and it would be advisable to make a second change to fix a warning.
> 
> First, replace -Wno-misleading-indentation by nothing. It seems to be wrong on GCC and, if anything, just suppresses a warning. Alternatively, it could be used on GCC, but replaced by nothing on MSVC (Windows).
> GCC compiler options
> https://gcc.gnu.org/onlinedocs/gcc/Option-Summary.html
> This defines -Wmisleading-indentation but not -Wno-misleading-indentation.

This is not true. See: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Warning-Options:
> You can request many specific warnings with options beginning with ‘-W’, for example -Wimplicit to request warnings on implicit declarations. Each of these specific warning options also has a negative form beginning ‘-Wno-’ to turn off warnings; for example, -Wno-implicit.

> Second, continue to use -std=c++11 (or a later version) on GCC, and use /std:c++14 or /std:c++latest on MSVC (Windows).
> https://docs.microsoft.com/en-us/cpp/build/reference/compiler-options?view=vs-2019
> https://docs.microsoft.com/en-us/cpp/build/reference/std-specify-language-standard-version?view=vs-2019

This is some pain in the ass. Where exactly does this error occur? Sundials is all C, there should not be any `-std=c++11`.

thomassligon:
The warning
    cl : Command line warning D9002 : ignoring unknown option '-std=c++11'
only occurs while compiling AMICI code, maybe because sundials compilation never gets that far. Fixing it is not necessary and not important. I just like to avoid warnings because it is a good idea in general.
Avoiding warnings due to misleading indentation is not necessary either, but incorrect indentation is not good for readability. How important is the option when using GCC for sundials? I don't remember having this problem when I compiled sundials via CMAKE in Visual Studio, in order to use it for a problem in celestial mechanics, where I was using all C (not C++).

dweindl:
Added this as new issue #812. This will involve a couple of more lines. I won't be able to do that within the next few days.

dweindl:
Maybe you want to use mingw for a start. This is known to work on windows. The required steps are described in the [installation instructions](https://github.com/ICB-DCM/AMICI/blob/master/INSTALL.md#windows).

src/solver_cvodes.cpp
```diff
@@ -65,7 +65,7 @@ static int fJBandB(realtype t, N_Vector x, N_Vector xB, N_Vector xBdot,
                    N_Vector tmp2B, N_Vector tmp3B);
 
 static int fJDiag(realtype t, N_Vector JDiag, N_Vector x, void *user_data)
-__attribute__((unused));
+[[gnu::unused]];
```
dweindl: @FFroehlich :
```
/home/travis/build/ICB-DCM/AMICI/src/solver_cvodes.cpp:68:15: warning: attribute ignored [-Wattributes]

 [[gnu::unused]];

               ^

/home/travis/build/ICB-DCM/AMICI/src/solver_cvodes.cpp:68:15: note: an attribute that appertains to a type-specifier is ignored

/home/travis/build/ICB-DCM/AMICI/src/solver_cvodes.cpp:849:5: warning: ‘int amici::fJDiag(amici::realtype, N_Vector, N_Vector, void*)’ defined but not used [-Wunused-function]

 int fJDiag(realtype t, N_Vector JDiag, N_Vector x, void *user_data) {

     ^~~~~~
```
Anything against simply removing this function?

dweindl:
@FFroehlich : Can be merged from my point of view.

FFroehlich:
> @FFroehlich : Can be merged from my point of view.

Great, will do as soon as tests have passed

PR #813, opened 2019-09-27
Fix handling of species specified in PEtab condition table

dweindl:
which currently have to be converted to parameters


python/bin/amici_import_petab.py
```diff
@@ -118,10 +118,16 @@ def get_fixed_parameters(condition_file_name, sbml_model,
     # remove overridden parameters
     fixed_parameters = [p for p in fixed_parameters
                         if condition_df[p].dtype != 'O']
-
     # must be unique
     assert(len(fixed_parameters) == len(set(fixed_parameters)))
 
+    # States occurring as columns names of the condition table need to be
```
yannikschaelte:
```suggestion
    # States occurring as column names of the condition table need to be
```

python/bin/amici_import_petab.py
```diff
@@ -152,44 +158,41 @@ def get_fixed_parameters(condition_file_name, sbml_model,
...
-    constant species for PEtab, since there it is not possible to specify
-    constant species in the condition table.
+def species_to_parameters(species_ids: list[str],
+                          sbml_model: 'libsbml.Model') -> None:
+    """Turn a SBML species into parameters and replace species references
+    inside the model instance.
 
     Arguments:
-        sbml_model: libsbml model instance
+        species_ids: List of SBML species ID to convert to parameter and ID
```
yannikschaelte:
```suggestion
        species_ids: List of SBML species IDs to convert to parameter and ID
```
don't understand this sentence

python/bin/amici_import_petab.py
```diff
@@ -199,18 +202,44 @@ def constant_species_to_parameters(sbml_model):
...
+    Arguments:
+        sbml_model: libsbml model instance
+
+    Returns:
+        List of IDs of SBML species that have been turned into constants
+
+    Raises:
+
+    """
+    transformables = []
```
yannikschaelte: ok I guess you just wrote a function for converting any species to parameters, and extracted the checking of which species are constant to here. seems fine.

python/bin/amici_import_petab.py
dweindl: Continued on the next line. A bit unfortunate how it is presented here ;)

yannikschaelte: I got that ;) what is "ID of the resulting parameter"? Means the parameter gets the same ID? (and: I assume you check that such a parameter does not yet exist?)

dweindl: Yes, species is removed and parameter is created with the same ID.
There can't be a species and parameter of the same name, so no need to check.

PR #820, opened 2019-10-04
Compute initial state for Model::getInitialStates if not already set (Fixes #818)

### LINKED ISSUE #818: Model::getInitialStates() return empty list
When instantiating `Model` and calling `getInitialStates()`, this returns an empty vector. Not sure whether this is a bug, but I'd consider it at least unexpected behaviour. If initial states have not been set explicitly, this should in my opinion call `Model::initializeStates()` automatically. Other opinions?

dweindl:
Requires returning initial state vector by value, as we don't want to cache the computed ones.


dweindl:
This does not work as expected. Problem is, that in `amici::ConditionContext` we use `(get|set)InitialStates()` to reset the Model after simulating different conditions. Therefore, x0 will be computed before simulating for condition 0, and 'restored' to this x0 after simulation. For condition 1, x0 will be set and thus fx0() not be called resulting in potentially wrong x0 being used for simulation.

Not sure how to best deal with that...

FFroehlich:
Introduce function
```
bool hasCustomInitialStates() {return x0data.empty()}
```
store that value in condition context and only restore if it was set, clear otherwise?
Same for `sx0`.

src/edata.cpp
```diff
@@ -404,12 +408,16 @@ void ConditionContext::restore()
     model->setParameterList(originalParameterList);
     // parameter scale has to be set before initial state sensitivities
     model->setParameterScale(originalScaling);
-    model->setInitialStates(originalx0);
-    model->setUnscaledInitialStateSensitivities(originalsx0);
+
+    if(!originalx0.empty())
+        model->setInitialStates(originalx0);
```
FFroehlich: Do we really need to check whether `originalx0` is empty here? I think we do want to reset Initial states if they were empty originally.

dweindl: I think if they were empty originally, them Model::x0 will still be empty after simulation. I think fx0 is not stored there if not manually set (in which case originalx0 here would be non-empty). With the current implementation it is not strictly required (we would just replace an empty vector by another empty vector), but I think it doesn't hurt. Maybe I am overlooking something.

FFroehlich: I think you are right, fine as is.

PR #824, opened 2019-10-04
Change SBML test suite to pytest to allow for parallel test execution…

### LINKED ISSUE #755: Allow running SBML test cases in parallel
... using python multiprocessing

dweindl:
… (Closes #755)

Pytest allows for cleaner test selection via command line interface
and parameterized tests. Can be run in parallel using pytest-xdist
e.g. via `pytest -n 4`

When using multiple threads, this does not show the current test ID
anymore. However, the test suite now completes in 40min instead 
of 1h20 on github actions.


FFroehlich:
👍  nice

PR #829, opened 2019-10-08
feature(cmake) Add option to disable compiler optimizations ...

### LINKED ISSUE #828: Add CMake option to disable compiler optimization for wrapfunctions.cpp
For large models, certain compilers (Intel) require vast amounts of memory optimizing `wrapfunctions.cpp`, since this compilation unit seems to include the code from `$modelname.h`. Add an option to disable all optimizations for this specific file. By default optimizations should be enabled.

dweindl:
... for wrapfunctions.cpp (Fixes #828)

### DIFF
--- a/src/CMakeLists.template.cmake
+++ b/src/CMakeLists.template.cmake
@@ -35,6 +35,13 @@ ${MODEL_DIR}/wrapfunctions.cpp
 add_library(${PROJECT_NAME} ${SRC_LIST_LIB})
 add_library(model ALIAS ${PROJECT_NAME})
 
+# This option can be helpful when using the Intel compiler and compilation of
+# wrapfunctions.cpp fails due to insufficient memory.
+option(ENABLE_WRAPFUNCTIONS_OPTIMIZATIONS "Enable compiler optimizations for wrapfunctions.cpp?" ON)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.template.cmake
```diff
@@ -35,6 +35,13 @@ ${MODEL_DIR}/wrapfunctions.cpp
 add_library(${PROJECT_NAME} ${SRC_LIST_LIB})
 add_library(model ALIAS ${PROJECT_NAME})
 
+# This option can be helpful when using the Intel compiler and compilation of
+# wrapfunctions.cpp fails due to insufficient memory.
+option(ENABLE_WRAPFUNCTIONS_OPTIMIZATIONS "Disable compiler optimizations for wrapfunctions.cpp" ON)
+if(NOT ENABLE_WRAPFUNCTIONS_OPTIMIZATIONS)
```
paulstapor: default is false or true here?

dweindl: default is at the very right: ENABLE_WRAPFUNCTIONS_OPTIMIZATIONS -> ON

Text is misleading though. Changed name without text. Fixing.

paulstapor: ah, sorry!

PR #838, opened 2019-10-19
Simplify symbolic sensitivity expressions during Python SBML import

dweindl:
Fixes Infs in the Jacobian when using Hill-functions with states of 0.0.

Mark all sympy symbols as reals, which sometimes allows for additional simplifications.

### DIFF
--- a/documentation/development.md
+++ b/documentation/development.md
@@ -98,8 +98,10 @@ described below:
   case for all existing code, any new contributions should do so. 
 
 * We use Python [type hints](https://docs.python.org/3/library/typing.html)
-  for all functions. In Python code type hints should be used instead of
-  doxygen `@type`. (All legacy `@type` attributes are to be removed.)
+  for all functions (but not for class attributes, since they are not supported
+  by the current Python doxygen filter). In Python code type hints should be
+  used instead of doxygen `@type`. (All legacy `@type` attributes are to be
+  removed.)
     
   For function docstrings, follow this format:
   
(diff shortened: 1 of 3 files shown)


dweindl:
@FFroehlich : This works for the SBML import but not yet for PySB. Most likely due to `pysb.core.Parameter`s being present in some expressions (In `ODEModel._derivative` called for `x`, `dxdotdx`). Was there a reason to not apply `strip_pysb` everywhere?

FFroehlich:
Not sure about the exact details anymore, but I believe applying strip_pysb everywhere is computationally expensive. Have to discuss this with pysb developers, but should probably be reasonable to assume all variables are real directly in pysb.

FFroehlich:
Is there any explanation why this works?

dweindl:
> Not sure about the exact details anymore, but I believe applying strip_pysb everywhere is computationally expensive. Have to discuss this with pysb developers, but should probably be reasonable to assume all variables are real directly in pysb.

The pysb problem is not so much whether they are reals, but that sympy.simplify cannot be applied to them (e.g. https://travis-ci.com/ICB-DCM/AMICI/jobs/247389091#L7209).

For now I will make simplification optional in `ODEModel` and enable only for SBML import.

For longer term, it would be cleaner if we could strip the pysb symbols already when assembling `ODEModel` to keep that part rather generic. Creating the `sympy.Symbol` shouldn't be crazily expensive compared to the computations they are used in afterwards.

dweindl:
> Is there any explanation why this works?

2 things I did here: 
* using `simplify`: this gets rid of states as multiplicands in the denominator of the respective entries of the Jacobian and is explainable 
* declare all sympy symbols as reals: not completely sure how much it matters in this context, but when testing some other stuff, it had some influence on how things were simplified. would have to search for an example.

PR #845, opened 2019-10-22
Simplify symbolic sensitivity expressions during Python SBML import

dweindl:
* Simplify symbolic sensitivity expressions during Python SBML import

Fixes Infs in the Jacobian when using Hill-functions with states of 0.0.

Mark all sympy symbols as reals, which sometimes allows for additional simplifications.

* Simplify all equations, but only if asked to do so

* doxygen cannot handle attribute typehints

Note that sympy substitutions only work for if IDs *and* predicates agree


dweindl:
Tested with a couple of SBML test suite cases. Works now as expected.

dweindl:
Finally... :expressionless:

dweindl:
@FFroehlich : Please review when convenient (travis failure is still due to ghostscript).

PR #848, opened 2019-10-24
Newton solver extensions

### LINKED ISSUE #842: Add Newton solver extensions
Merge the Newton solver improvements from

G. T. Lines, @paszkow, L. Schmiester, D. Weindl, P. Stapor, and J. Hasenauer. Efficient computation of steady states in large-scale ODE models of biochemical reaction networks. accepted for Proceedings of the 8th IFAC Conference on Foundations of Systems Biology in Engineering (FOSBE), Valencia, Spain, October 2019.

paszkow:
The changes that allow performing Newton tests from the paper:

G. T. Lines, Ł. Paszkowski, L. Schmiester, D. Weindl, P. Stapor, and J. Hasenauer. Efficient computation of steady states in large-scale ODE models of biochemical reaction networks. accepted for Proceedings of the 8th IFAC Conference on Foundations of Systems Biology in Engineering (FOSBE), Valencia, Spain, October 2019.

Closes #842

### DIFF
--- a/include/amici/defines.h
+++ b/include/amici/defines.h
@@ -148,6 +148,12 @@ enum class NewtonStatus {
     newt_sim_newt=3,
 };
 
+/** Damping factor flag for the Newton method */
+enum class NewtonDampingFactorMode {
+    off = 0,
+    on = 1
+};
+
 /**
  * @brief msgIdAndTxtFp
  * @param identifier string with error message identifier
--- a/include/amici/newton_solver.h
+++ b/include/amici/newton_solver.h
@@ -47,6 +47,7 @@ class NewtonSolver {
      * @param atol absolute tolerance
      * @param rtol relative tolerance
      * @param dampingFactorMode to switch on/off a use of a damping factor
+     * @param dampingFactorLowerBound a lower bound for the damping factor
      * @return solver NewtonSolver according to the specified linsolType
      */
     static std::unique_ptr<NewtonSolver> getSolver(realtype *t, AmiVector *x,
@@ -56,7 +57,8 @@ class NewtonSolver {
                                                    int maxlinsteps,
                                                    int maxsteps,
                                                    double atol, double rtol,
-                                                   NewtonDampingFactorMode dampingFactorMode);
+                                                   NewtonDampingFactorMode dampingFactorMode,
+                                                   double dampingFactorLowerBound);
 
     /**
      * Computes the solution of one Newton iteration
@@ -107,6 +109,8 @@ class NewtonSolver {
     double rtol = 1e-8;
     /** damping factor flag */
     NewtonDampingFactorMode dampingFactorMode = NewtonDampingFactorMode::on;
+    /** damping factor lower bound */
+    double dampingFactorLowerBound = 1e-8;
 
   protected:
     /** time variable */
--- a/include/amici/newton_solver.h
+++ b/include/amici/newton_solver.h
@@ -46,6 +46,7 @@ class NewtonSolver {
      * @param maxsteps maximum number of allowed Newton steps for steady state computation
      * @param atol absolute tolerance
      * @param rtol relative tolerance
+     * @param dampingFactorMode to switch on/off a use of a damping factor
      * @return solver NewtonSolver according to the specified linsolType
      */
     static std::unique_ptr<NewtonSolver> getSolver(realtype *t, AmiVector *x,
@@ -54,7 +55,8 @@ class NewtonSolver {
                                                    ReturnData *rdata,
                                                    int maxlinsteps,
                                                    int maxsteps,
-                                                   double atol, double rtol);
+                                                   double atol, double rtol,
+                                                   NewtonDampingFactorMode dampingFactorMode);
 
     /**
      * Computes the solution of one Newton iteration
@@ -103,6 +105,8 @@ class NewtonSolver {
     double atol = 1e-16;
     /** relative tolerance */
     double rtol = 1e-8;
+    /** damping factor flag */
+    NewtonDampingFactorMode dampingFactorMode = NewtonDampingFactorMode::on;
 
   protected:
     /** time variable */
--- a/src/hdf5.cpp
+++ b/src/hdf5.cpp
@@ -636,7 +636,7 @@ void readSolverSettingsFromHDF5(H5::H5File const& file, Solver &solver,
 
     if(attributeExists(file, datasetPath, "newton_damping_factor_lower_bound")) {
         solver.setNewtonDampingFactorLowerBound(
-                    getDoubleScalarAttribute(file, datasetPath, "newton_damping_factor_lower_bound")));
+                    getDoubleScalarAttribute(file, datasetPath, "newton_damping_factor_lower_bound"));
     }
 
     if(attributeExists(file, datasetPath, "newton_maxlinsteps")) {
--- a/src/hdf5.cpp
+++ b/src/hdf5.cpp
@@ -628,6 +628,12 @@ void readSolverSettingsFromHDF5(H5::H5File const& file, Solver &solver,
                     getIntScalarAttribute(file, datasetPath, "newton_preeq"));
     }
 
+    if(attributeExists(file, datasetPath, "newton_damping_factor_mode")) {
+        solver.setNewtonDampingFactorMode(
+                    static_cast<NewtonDampingFactorMode>(
+                        getIntScalarAttribute(file, datasetPath, "newton_damping_factor_mode")));
+    }
+
     if(attributeExists(file, datasetPath, "newton_maxlinsteps")) {
         solver.setNewtonMaxLinearSteps(
                     getIntScalarAttribute(file, datasetPath,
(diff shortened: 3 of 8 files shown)


dweindl:
Thanks @paszkow . @paulstapor , can you please look into that when time permits?

paszkow:
@dweindl @paulstapor You might want to hold on for a while. Turn out I cannot reproduce all of the tests from the paper. There might be something missing here.

paulstapor:
Thanks for your work @paszkow , and thanks for your message, good to know.
We are also busy with fixing #849, just poke me as soon as you think I should do a code review.

include/amici/defines.h
```diff
@@ -148,6 +148,12 @@ enum class NewtonStatus {
     newt_sim_newt=3,
 };
 
+/** Damping factor flag for the Newton method */
+enum class NewtonDampingFactorMode {
+    off = 0,
+    on = 1
```
paulstapor: Just to make sure: Default should then be `on`, right?

include/amici/newton_solver.h
```diff
@@ -103,6 +105,8 @@ class NewtonSolver {
     double atol = 1e-16;
     /** relative tolerance */
     double rtol = 1e-8;
+    /** damping factor flag */
+    NewtonDampingFactorMode dampingFactorMode = NewtonDampingFactorMode::on;
```
paulstapor: There we go, yes... Ignore the comment above...

paszkow:
I think I have pushed all relevant changes. After a proper re-generating the model and tweaking the test code I can reproduce the simulations.

src/hdf5.cpp
```diff
@@ -628,6 +628,12 @@ void readSolverSettingsFromHDF5(H5::H5File const& file, Solver &solver,
                     getIntScalarAttribute(file, datasetPath, "newton_preeq"));
     }
 
+    if(attributeExists(file, datasetPath, "newton_damping_factor_mode")) {
+        solver.setNewtonDampingFactorMode(
+                    static_cast<NewtonDampingFactorMode>(
+                        getIntScalarAttribute(file, datasetPath, "newton_damping_factor_mode")));
+    }
+
```
paulstapor: Yep, makes sense that this had an impact on reproducing the results... ;)

src/steadystateproblem.cpp
```diff
@@ -206,21 +207,16 @@ void SteadystateProblem::applyNewtonsMethod(ReturnData *rdata, Model *model,
...
             converged = wrms < RCONST(1.0);
 
-            if (converged) {
-                /* Ensure positivity of the found state */
-                for (ix = 0; ix < model->nx_solver; ix++) {
-                    if (x[ix] < 0.0) {
-                        x[ix] = 0.0;
-                        converged = FALSE;
-                    }
-                }
```
paulstapor: What is the reason to kick out this part? Usually, we want positive states...
Sure, this part may screw up convergence, if we converged to a steady state with negative entries... But usually, we don't want to keep this steady state with negative entries anyway...

src/steadystateproblem.cpp
```diff
@@ -206,21 +207,16 @@ void SteadystateProblem::applyNewtonsMethod(ReturnData *rdata, Model *model,
...
+            if (!converged && newtonSolver->dampingFactorMode==NewtonDampingFactorMode::on) {
+              /* increase dampening factor (superfluous, if converged) */
+              gamma = fmin(1.0, 2.0 * gamma);
             }
-        } else {
-            /* Reduce dampening factor */
+        } else if (newtonSolver->dampingFactorMode==NewtonDampingFactorMode::on) {
+            /* Reduce dampening factor and raise an error when becomes too small */
             gamma = gamma / 4.0;
+            if (gamma < 1e-6)
```
paulstapor: How sure are we about this number?
Might it be good to make this an user-adjustable parameter?

src/steadystateproblem.cpp
paszkow: Makes problems when you have a zero steady-state (or at least zero at one index). Then you will never converge due to floating-point operations. The number -1e-32 is so close to zero that you do not want to force it to be 0.0. We had such issues.

On the other hand, for strictly positive steady-states this part of the code was never invoked.

src/steadystateproblem.cpp
paszkow: Up to you. It is just a reasonable lower bound. Usually, you do not need to go lower. If you need, then most likely you will not converge at all.

src/steadystateproblem.cpp
paulstapor: I see your point... but:
This part of Amici is sometimes used to preequilibrate systems and start a simulation after prequilibration, i.e., sometimes, the Ewton-solver is used to find inital values of ODEs. Negative states can then cause integration failure, when a simulation is started from them... We also experienced cases like this...

So, we must find a solution to this... 
I don't yet really understand what goes wrong if we just set those `x` to `0.0`. Can you re-explain, please?

About strict positivity: At least to my understanding, this is not what one is looking for. It is important to be `>= 0.0`. Please correct me here, if I'm wrong. Maybe @FFroehlich knows more about these cases than I do...

How about checking whether we are in between of `-absolute_tolerance` and `0.0`, an just flip sign in that case?

src/steadystateproblem.cpp
paulstapor: I see the point, but I like avoiding magic numbers which are hard-coded in algorithms...
As the damping factor is something like a step size, I'm wondering whether linking it to the error tolerances might be a good idea...

src/steadystateproblem.cpp
paszkow: I see the issues. Maybe a reasonable solution would be to have `if (x[i] < -eps)` for some epsilon.

In our case, with this part of the code and zero steady-states, the Newton solver never converged. Always reached the maximum number of iterations. In the debugging, I noticed the solution was forced to be zero and in the next step, it again went to negative (-1e-32). The process repeated all the time.

src/steadystateproblem.cpp
dweindl: Agreed with the magic numbers. Maybe for start just add them as function arguments with the given default values. Linking to error tolerances seems like a start of a new project. Sounds good, but maybe too much here.

paszkow: Totally up to you.

src/steadystateproblem.cpp
paulstapor: Ah, okay, now I see what's wrong:
You're totally right about taking out the `converged = FALSE;`, indeed!
One thing: Can we agree on setting stuff to 0.0, and then just rechecking for convergence?
If convergence persists, we are fine and keep `converged = TRUE;` and hence drop out of the loop. Only if we have no convergence, we set  `converged = FALSE;` after reapplying the positivity constraint?

src/steadystateproblem.cpp
paulstapor: Let's make it a user option for the moment.

src/steadystateproblem.cpp
paszkow: Understand. I will check it out. Could be enough indeed.

src/steadystateproblem.cpp
paszkow: OK

paszkow: Is 1e-6 a reasonable default value or you prefer a different number?

paulstapor: Seems reasonable to me. Either that or 1e-8, which is our default for relative tolerances... Both are fine for me.

src/steadystateproblem.cpp
paulstapor: Does your code example still work with this adaptation now? If yes, I'd do a review for merging.

paszkow: I wouldn't have done this change without running my test first. It works indeed.

paulstapor: Okay, point for you. ;)

paulstapor:
Little error persisting on the documentation:
```
DOXYGEN failed:
3559/Users/travis/build/ICB-DCM/AMICI/include/amici/solver.h:228: warning: argument 'damping' of command @param is not found in the argument list of Solver::setNewtonDampingFactorMode(NewtonDampingFactorMode dampingFactorMode)
3560/Users/travis/build/ICB-DCM/AMICI/include/amici/solver.h:232: warning: The following parameters of amici::Solver::setNewtonDampingFactorMode(NewtonDampingFactorMode dampingFactorMode) are not documented:
3561  parameter 'dampingFactorMode'
3562/Users/travis/build/ICB-DCM/AMICI/include/amici/solver.h:240: warning: argument 'damping' of command @param is not found in the argument list of Solver::setNewtonDampingFactorLowerBound(double dampingFactorLowerBound)
3563/Users/travis/build/ICB-DCM/AMICI/include/amici/solver.h:244: warning: The following parameters of amici::Solver::setNewtonDampingFactorLowerBound(double dampingFactorLowerBound) are not documented:
3564  parameter 'dampingFactorLowerBound'
```

dweindl:
CI failure is due to an unrelated doxygen problem. Consider that passed.

dweindl:
@paszkow : Can you please add `Model::newton_damping_factor_mode` and `Model::newton_damping_factor_lower_bound` to the respective function in `include/amici/serialization.h`, then we are done, I think.

paszkow:
Sorry. I missed the serialization part.

PR #853, opened 2019-10-25
Setup performance test on github actions

### LINKED ISSUE #846: Add performance test to CI
For merges to master, it would be great to import and simulate a large model (Fröhlich et al., Cell Systems, 2018; Schmiester at al, Bioinformatics, 2019) and ensure that simulation and import times did not degrade significantly.

dweindl:
Setup performance test on github actions (Closes #846)

* Setup github action to import and run model from Fröhlich et al., Cell Systems, 2018 (forward simulation, forward+adjoint sensitivities) 
* Don't require use of ModelPtr.get to call ExpData(Model)
* Fix import in generated model Python package
* Setup AMICI standalone scripts as setuptools entrypoints
* Add reference values for wall times and check we are within the limits


python/amici/__init__.py
```diff
@@ -157,6 +157,8 @@ def ExpData(*args):
         # the constructor signature may have changed and you are glad this
         # wrapper did not break.
         return amici.ExpData(args[0].get(), *args[:1])
+    elif isinstance(args[0], ModelPtr):
+        return amici.ExpData(args[0].get())
```
FFroehlich:
```suggestion
        return amici.ExpData(args[0].get(), *args[:1])
```

FFroehlich:
👍

python/amici/__init__.py
dweindl: For reasons I don't quite understand, this didn't work in 
4a13511. Keeping as was.

PR #855, opened 2019-10-28
[WIP] Support of RateRules and import of raw ODEs

jvanhoefer:
Current version of txt import + rate rules (still work in progress!)


jvanhoefer:
Not everything regarding the rate rules is working yet. 

I just wanted to move my work from my own AMICI fork to a feature branch and will continue to work on/test the rate rules there.

jvanhoefer:
Still: The SBML generation from a .txt-file should work.

python/amici/sbml_import.py
```diff
@@ -522,6 +519,29 @@ def processCompartments(self):
...
+
+        for rule in self.sbml.getModel().getListOfRules():
+            if rule.isRate():
+
+                # set constant input reactions with kinetic law
+                # equal to the right hand site of the Rate Rule
+                r = self.sbml.createReaction()
+                r.setId('d_dt_' + rule.getVariable())
+                r.setName('d_dt_' + rule.getVariable())
+                r.createProduct().setSpecies(rule.getVariable())
```
dweindl: The rule variable can also be a parameter, this would have to be treated differently (see SBML specs 4.9.4).

Please run the SBML test suite with these changes (`scripts/run-SBMLTestsuite.sh`). This should include a couple of cases for RateRules.

I am not sure if we want to alter the SBML object here, or not rather just add these reactions to the symbols of SBMLImporter (@FFroehlich ?)

jvanhoefer: Thank you for your pointing this out!

If it is to a parameter, should we create a new species, that is "equal to the parameter", and apply the rate rule to the new species? (And also later, what if we introduce variable compartment size and the rate rule acts on the compartment? That we need to keep in mind for later :) )

FFroehlich:
Do we really want to deal with plaintext specified models? Just eyeballing `parse_txt` I bet you that I can come up with over 10 examples of small errors that will break the parser. We would have to develop testcases to anticipate and catch all of these issues and make sure that meaningful error messages are displayed for each issue (i.e., major pain in the ass).

Having the user specify the model in some standardized sympy would shift the responsibility of specifying something interpretable to the user (which shouldnt be too difficult and serves as an error control to him) and makes our life way easier. If it really has to be a plain text file I would prefer something a bit more structured like yaml which would provide us with some basic input verification. Also makes programmatic specification of models much easier.

JanHasenauer:
We would like to have something which can be used from Matlab and python, hence sympy is not a good choice. yaml would be fine for me but would not address the first point you raised.

FFroehlich:
yaml should already solve most of the issue related to plain text parsing and significantly reduce the amount of sanity checks we have to perform.

jvanhoefer:
The thing is that we will need a txt import for several projects relatively soon and the current version fo the SBML generation is working for all examples from those projects. 

My suggestion would be to take the (already implemented) txt->SBML translation part and replace it by another import format later.

jvanhoefer:
Alternatively we can not include the txt->SBML translation in AMICI and I only use this on my own and only focus on implementing Rate Rules in the SBML import.

FFroehlich:
I am strongly opposed to knowingly creating technical debt (just browse through the issues and count the issues regarding things we wanted to implement "later"). Converting to a more structured format such as yaml really is minimal effort (you effectively just add a couple of `-` and minimally change the syntax) but is likely going to prevent a lot of headaches/troubleshooting down the road.

plaintext is pretty permissive in terms of what you can specify (similar to sbml, which is why they need a gazillion testcases to ensure everything is always handled correctly). If you add functionality to AMICI I expect you also provide respective tests to a reasonable degree (which for plaintext would be a phenomenal amount of work). 

I am pretty pedantic about theses this issue because I know that these things usually cause trouble when you are already facing 5 other deadlines and don't really have the time to deal with new issues (ask @paulstapor how much he appreciates working through the symbolic processing right now). If you have the chance to fix things early on, do it, because it usually goes a long way. 

It generally also makes things more reusable. Imagine you want to come back to your txt models in 3 years. You will probably either have to go back to AMICI version X which doesnt have feature Y (again, ask @paulstapor how pleasant this experience is) which you really need so you have to rewrite all your models AND make sure that they still do the same thing as their txt implementation.
If you dont want anybody ever to reuse your code, why would you publish it in the first place?

Just to make it clear, @paulstapor is not directly responsible for the technical debt that is causing issues, but currently has to deal with it.

python/amici/sbml_import.py
FFroehlich: We do not want to ever allow rate rules for parameters as respective sensitivities are not well defined.

I would actually say that flattening rate rules to reactions is a good thing as it simplifies downstream processing (it will give rise to a stoichiometric matrix + flux vector etc that we can use later on)

jvanhoefer: @FFroehlich @dweindl  Is it the right way to go, to modify the libsbml model? Or do we want to modify the sympy-part of SBML_import? (In that case I would have to take a deeper dive into the code's structure and would for sure need some more week(s?) more...)

FFroehlich: I think its fine to modify the SBML, however I think this functionality should not be part of `sbml_import.py`. I would rather have this implemented as general purpose 'flattener' that acts on an sbml object, which is passed to import_sbml as preprocessing, similar to what we have for observables/constants etc.

dweindl: Okay then. As Fabi said, then this should be a free function.

dweindl:
Also in favor of something that is simpler to generate programmatically. Agreed that it would be good to have something readable from both matlab and python. However, if it is supposed to be usable from both of them directly, then the same parsing and SBMLification needs to be implemented in Matlab as well. This creates more Python/Matlab redundancies which are an annoying maintainance burden and which we were aiming to reduce.
(If matlab users should create the sbml model using python, then there is no good reason to not use sympy directly)

dweindl:
Closing due to inactivity. Reopen if appropriate.

PR #871, opened 2019-12-01
Speed up sympy simplification

### LINKED ISSUE #862: Speed-up sympy simplification
#845 fixed some issues regarding NaNs in the Jacobian. However, simplification doubles import time for larger models. Check if a more targeted simplification does the job faster. And/or make simplification option.

(afterwards, adapt time limit in `tests/performance/reference.yaml`)

dweindl:
Closes #862

* Add test case
* Use powsimp instead of simplify

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -18,7 +18,7 @@ except ImportError:
     pysb = None
 
 
-from typing import Callable
+from typing import Callable, Optional
 from string import Template
 import sympy.printing.ccode as ccode
 from sympy.matrices.immutable import ImmutableDenseMatrix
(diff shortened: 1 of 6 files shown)


python/amici/ode_export.py
```diff
@@ -714,11 +715,12 @@ class ODEModel:
...
 
-        _simplify: if set to true, symbolic derivative expressions will be
-        simplified. NOTE: This does currently not work with PySB symbols.
+        _simplify: if set, this function will be used to simplify symbolic
+         derivative expressions.
+         NOTE: This does currently not work with PySB symbols.
     """
 
-    def __init__(self, simplify: bool = False):
+    def __init__(self, simplify: Callable = sp.powsimp):
```
FFroehlich: can we make simplify an iterable of Callables that are to be applied sequentially? Otherwise its not straightforward to me what to pass if no simplifications are to be applied.

dweindl: Can do that, or change typehint to `Optional[Callable]` which implies None as a valid option. Do you see clear applications for an Iterable? Can also use None as default here.

FFroehlich: What if I want to apply both `ratsimp` and `powsimp`? For the models we look at we actually apply multiple rounds of simplifications step to specific expressions in the model (this is done before model import, so it doesnt need to be applied here, but should illustrate that this is sometimes reasonable).

dweindl: The you pass a lambda doing that ;)

FFroehlich: Oh passing lambda is probably the better approach as its overall more flexible. Fine with me to just mention that in the documentation.

Issue #866, opened 2019-11-19
Several Matlab examples broken


dweindl:
```
Unexpected Standard exception from MEX file.
What() is:vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
..

Error in simulate_model_dirac_adjoint (line 227)
sol = ami_model_dirac_adjoint(tout,theta(1:4),kappa(1:0),options_ami,plist,xscale,init,data);

Error in example_dirac_adjoint (line 33)
sol = simulate_model_dirac_adjoint(tout,log10(p),k,D,options);
```

dweindl:
Also
```
Unexpected Standard exception from MEX file.
What() is:vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
..

Error in simulate_model_adjoint (line 227)
sol = ami_model_adjoint(tout,theta(1:3),kappa(1:0),options_ami,plist,xscale,init,data);

Error in example_adjoint (line 35)
sol = simulate_model_adjoint(t,log10(p),k,D,options);

Error in amiExamples (line 7)
example_adjoint
```

dweindl:
```
Unexpected Standard exception from MEX file.
What() is:vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
..

Error in simulate_model_adjoint_hessian (line 232)
    sol = ami_model_adjoint_hessian(tout,theta(1:3),kappa(1:0),options_ami,plist,xscale,init,data);

Error in example_adjoint_hessian (line 34)
sol = simulate_model_adjoint_hessian(t,log10(p),k,D,options);

Error in amiExamples (line 11)
example_adjoint_hessian
```

FFroehlich:
💩

dweindl:
and `example_dae_events` referenced in `amiExamples.m` is missing

paulstapor:
on which branch?

FFroehlich:
I guess thats on master 🤷‍♂

dweindl:
`example_neuron` crashes (my) matlab on `develop`. 

> on which branch?

Tested on #576, but most likely master as well, as the relevant parts haven't been touched in a while.

dweindl:
That's what you get from not being able to run your CI workflow on the Matlab parts...

dweindl:
Fixed:
example_dirac_adjoint
example_adjoint
example_adjoint_hessian
example_dirac_adjoint_hessVecProd

[...]

PR #875, opened 2019-12-04
Fix accessing empty vector during adjoint state event update (Closes …

dweindl:
…#866)


FFroehlich:
guess thats on me, good that it only took us 6 months to notice 🙄

dweindl:
Yeah, seems like Matlab+Events is not overly used :)

PR #885, opened 2019-12-15
CI: Ignore doxygen ghostscript error until doxygen is fixed

dweindl:


dweindl:
Finally some green again :D

PR #888, opened 2019-12-16
Add option for PEtab-YAML import in amici_import_petab

dweindl:


python/amici/petab_import.py
```diff
@@ -401,12 +405,11 @@ def parse_cli_args():
             args.parameter_file_name = petab.get_default_parameter_file_name(
                 args.model_name)
 
-    if not args.model_name and \
-            (not args.sbml_file_name
-             or not args.condition_file_name
-             or not args.measurement_file_name):
-        parser.error('When not specifying a model name, sbml, '
-                     'condition and measurement file must be specified')
+    if not args.model_name and not args.yaml_file_name \
```
yannikschaelte: being picky today ;) can be simplified as
```python
if not (args.model_name or args.yaml_file_name
        or all(args.sbml_file_name, args.condition_file_name,
               args.measurement_file_name)):
```
just kidding, fine as it is

dweindl: I shortly considered an outer `any`, but then it felt like too many parentheses... ;)

PR #898, opened 2020-01-09
Python 3.8, fixes #897

### LINKED ISSUE #897: python 3.8
Python 3.8 was released October last year but tests are still running on python 3.7. 
We are currently claiming support for >=3.6 but don't actually check that in CI. Did we ever decide on how many minor python versions we want to support?

FFroehlich:


FFroehlich:
not sure why codecov claims coverage is going down, its actually going up by 3.8%

.travis.yml
```diff
@@ -101,30 +130,31 @@ install:
...
-    - if [[ "$CI_MODE" == "deploy" ]]; then pip3 install --user --upgrade pip==9.0.3 doxypypy; fi # pinning pip because of https://github.com/pypa/pip/issues/5240
-    - if [[ "$CI_MODE" == "deploy" ]]; then export PATH="/usr/local/opt/bison/bin:$PATH"; LDFLAGS="-L/usr/local/opt/bison/lib" scripts/downloadAndBuildDoxygen.sh; fi
-    - if [[ "$CI_MODE" == "test" ]]; then ./scripts/buildSuiteSparse.sh; fi
-    - if [[ "$CI_MODE" == "test" ]]; then ./scripts/buildSundials.sh; fi
-    - if [[ "$CI_MODE" == "test" ]]; then ./scripts/buildCpputest.sh; fi
-    - if [[ "$CI_MODE" == "test" ]]; then ./scripts/buildBNGL.sh; fi
-    - if [[ "$CI_MODE" == "test" ]]; then ./scripts/buildAmici.sh; fi
-    - if [[ "$CI_MODE" == "test" ]]; then ./scripts/installAmiciArchive.sh; fi
-    - if [[ "$CI_MODE" == "test" ]]; then ./scripts/installAmiciSource.sh; fi
+    - if [[ "$TRAVIS_OS_NAME" == "linux" ]] && [[ "$CI_DOC" != "TRUE" ]]; then pip3 install --upgrade pip==9.0.3 setuptools wheel pkgconfig scipy; fi
```
dweindl: `"$CI_DOC" != "TRUE"` will do the job here for this moment, but I would an *equals* more intuitive and future-safe. same for the *not-equals* below.

.travis.yml
```diff
@@ -101,30 +130,31 @@ install:
...
-    - if [[ "$TRAVIS_OS_NAME" != "linux" ]] && [[ "$CI_MODE" == "test" ]]; then $FOLD python-tests ./scripts/run-python-tests.sh; fi
-    - if [[ "$TRAVIS_OS_NAME" != "linux" ]] && [[ "$CI_MODE" == "test" ]]; then $FOLD cmake ./tests/testCMakeCompilation.sh; fi
-    - if [[ "$CI_MODE" == "deploy" ]]; then $FOLD doxygen ./scripts/run-doxygen.sh; fi
-    - if [[ "$TRAVIS_OS_NAME" != "linux" ]] && [[ "$CI_MODE" == "test" ]]; then $FOLD cppcheck ./scripts/run-cppcheck.sh; fi
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]] && [[ "$CI_MODE" == "test" ]]; then $FOLD valgrind ./scripts/run-valgrind.sh; fi
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]] && [[ "$CI_MODE" == "test" ]]; then $FOLD codecov ./scripts/run-codecov.sh; fi
+    - if [[ "$CI_DOC" != "TRUE" ]]; then $FOLD notebooks "cd $BASE_DIR && scripts/runNotebook.sh python/examples/example_*/"; fi
+    - if [[ "$CI_CPPUTEST" == "TRUE" ]] && [[ "$CI_VALGRIND" == "TRUE" ]]; then $FOLD valgrind ./scripts/run-valgrind.sh; fi
+    - if [[ "$CI_CPPUTEST" == "TRUE" ]] && [[ "$CI_VALGRIND" != "TRUE" ]]; then $FOLD cpputest ./scripts/run-cpputest.sh; fi
```
dweindl: the cpputest one we could drop, right? they are also executed with run-valgrind.sh. would safe some time.

.travis.yml
```diff
@@ -101,30 +130,31 @@ install:
...
+    - if [[ "$CI_CPPCHECK" == "TRUE" ]]; then $FOLD cppcheck ./scripts/run-cppcheck.sh; fi
```
dweindl: shall we move the fast steps to the beginning?

.travis.yml
FFroehlich: as far as I remember valgrind doesnt work for osx so we have to keep the pure cpputest, made this more explicit.

.travis.yml
FFroehlich: done

FFroehlich:
> Absolutely makes sense to organize things this way. Was wondering if we should move some of the tests to github actions (not in this PR). Works really well, the nodes seem to be faster and offer 2 CPUs afaik. The queues on travis are often annoying...

Not sure, havent looked at gh actions in too much detail. Does it offer the same kind of job matrices? I suppose for the doc-build/check it probably would make sense as the overlap with other ci-jobs is really minimal, but for everything else the build matrix really does minimize code duplication.

PR #916, opened 2020-01-16
C++11 code printing

FFroehlich:
use more modern code printer in sympy.


FFroehlich:
> Didn't look at the result, but sounds good. Any hint on whether that affects compilation time in any way?

I don't think this should have any effect on compilation time. I think this primarily increases supported+reserved functions and thus permits/prevents compilation where appropriate. Stumbled across this in a stale branch where I was trying out additional features that didn't turn out to be relevant. Still thought this is a reasonable thing to do.

PR #932, opened 2020-02-03
Model name, fixes #930

### LINKED ISSUE #930: Include model name in generated C++ code
Would sometimes be convenient when passing around Model instances without other context to have a `model.getName()`.

FFroehlich:


src/model_header.ODE_template.h
```diff
@@ -733,6 +733,10 @@ class Model_TPL_MODELNAME : public amici::Model_ODE {
     TPL_X_SOLVER_IMPL
 
     TPL_TOTAL_CL_IMPL
+    
+    virtual std::string getName() const override {
```
dweindl:
```suggestion
    std::string getName() const override {
```

Redundant for overrides.

PR #939, opened 2020-02-07
add simplified petab import function

yannikschaelte:
This is basically the pypesto import function which there serves to allow to define an objective from yaml without having to define any model names etc.. makes sense to move here since can be used also in other tools?


python/amici/petab_import.py
```diff
@@ -185,6 +188,129 @@ def constant_species_to_parameters(sbml_model: 'libsbml.Model') -> List[str]:
     return species_to_parameters(transformables, sbml_model)
 
 
+def import_petab_problem(
+        petab_problem: petab.Problem,
+        folder: str = None,
```
dweindl: stick to `model_output_dir` used below? unnecessarily confusing if the same argument has different names

python/amici/petab_import.py
```diff
@@ -185,6 +188,129 @@ def constant_species_to_parameters(sbml_model: 'libsbml.Model') -> List[str]:
...
+
+        # compile the model
+        import_model(sbml_model=petab_problem.sbml_model,
+                     condition_table=petab_problem.condition_df,
+                     observable_table=petab_problem.observable_df,
+                     model_name=model_name,
+                     model_output_dir=model_output_dir,
+                     **kwargs)
+
+    # load moduule
```
dweindl:
```suggestion
    # load module
```

PR #947, opened 2020-02-10
PEtab: fix parameter mapping errors in presence of LocalParameters

dweindl:


dweindl:
@yannikschaelte : Same problem currently occurs in pypesto and needs fixing there.

PR #921, opened 2020-01-24
Extend PEtab support

dweindl:
Closes #890 

* Update PEtab model import (now using YAML files and observable table)
* Add `simulate_petab(...)` to compute cost function / simulate model based on PEtab data
* Add tests for PEtab models based on benchmark collection to run as github action

PR #961, opened 2020-02-18
fix processing of assignment rule variables in petab import

FFroehlich:


FFroehlich:
already fixed in #921

PR #964, opened 2020-02-19
Correct Borghans model, apply parameter scaling for fixed parameters

yannikschaelte:


tests/benchmark-models/benchmark_models.yaml
```diff
@@ -12,8 +12,8 @@ Boehm_JProteomeRes2014:
   note: benchmark collection reference ignores factor 1/2
 
 Borghans_BiophysChem1997:
-  llh: -3471.740659965799
-  note: benchmark collection reference value does not match, but model outputs do. maybe due to D2D data normalization
+  llh: 175.90130682476047
+  note: benchmark collection reference value does not match, but model outputs do. due to D2D data normalization
```
dweindl:
```suggestion
  note: matches benchmark collection, except for sign
```
right?

yannikschaelte: benchmark says -192, so we need to add the log10-normal correction sum(log10(sim)), then yes we have minus the benchmark value ...

dweindl: okay, please update accordingly

yannikschaelte: done

PR #948, opened 2020-02-10
AMICI dockerfile and deployment to dockerhub 

dweindl:
#909

PR #966, opened 2020-02-20
add llh for crauste

yannikschaelte:
add llh for crauste


yannikschaelte:
> There seems to be something wrong with the branch name...

ups. how could that happen? two typos in one word.

so the tests do not run through due to libcombine. what can we do about that?

dweindl:
> so the tests do not run through due to libcombine. what can we do about that?

Working on it in #948

PR #970, opened 2020-02-21
Fix(python) import_petab_problem: Add absolute paths to python path, invalidate caches and reload

### LINKED ISSUE #965: petab_import fails due to filesystem/import delay?
Not sure about the exact reason, but I find that https://github.com/ICB-DCM/AMICI/blob/16e3d762a94a35d21825f904d9d4381f82d117ab/python/amici/petab_import.py#L288 is sometimes arbitrarily failing. Might be due to filesystem delay?

dweindl:
…Closes #965)


FFroehlich:
👍

PR #984, opened 2020-03-02
fix publishing on RTD

FFroehlich:


FFroehlich:
see https://amici.readthedocs.io/en/fix_rtd/generated/amici.amici.html#module-amici.amici

PR #992, opened 2020-03-06
Make w available in ReturnData (Closes #990)

### LINKED ISSUE #990: Make flux vector available in ReturnData
Would be helpful. Is already computed, but so far not exposed.

Will again increase memory requirements. If that is a concern, it could be addressed by introducing an option to specify which quantities are to be retained.

Anything against adding that to ReturnData?

dweindl:


src/model.cpp
```diff
@@ -742,6 +742,12 @@ void Model::requireSensitivitiesForAllParameters() {
     initializeVectors();
 }
 
+void Model::getW(gsl::span<realtype> w, const realtype t, const AmiVector &x)
```
FFroehlich:
```suggestion
void Model::getExpression(gsl::span<realtype> w, const realtype t, const AmiVector &x)
```

swig/model.i
```diff
@@ -31,6 +31,7 @@ using namespace amici;
 %ignore getEventTimeSensitivity;
 %ignore getObservable;
 %ignore getObservableSensitivity;
+%ignore getW;
```
FFroehlich:
```suggestion
%ignore getExpression;
```

src/forwardproblem.cpp
```diff
@@ -471,7 +471,8 @@ void ForwardProblem::getEventSensisFSA(int ie) {
 void ForwardProblem::handleDataPoint(int it) {
     model->fx_rdata(x_rdata, x);
     std::copy_n(x_rdata.data(), rdata->nx, &rdata->x.at(it * rdata->nx));
-
+    model->getW(slice(rdata->w, it, model->nw),
```
FFroehlich:
```suggestion
    model->getExpression(slice(rdata->w, it, model->nw),
```

include/amici/model.h
```diff
@@ -675,6 +675,14 @@ class Model : public AbstractModel {
      */
     void requireSensitivitiesForAllParameters();
 
+    /**
+     * @brief Time-resolved w,
+     * @param w buffer (dimension: nw)
+     * @param t current timepoint
+     * @param x current state
+     */
+    void getW(gsl::span<realtype> w, const realtype t, const AmiVector &x);
```
FFroehlich:
```suggestion
    void getExpression(gsl::span<realtype> w, const realtype t, const AmiVector &x);
```

PR #996, opened 2020-03-07
Python model import: Verify that chosen model name is a valid identif…

### LINKED ISSUE #928: Python SBML import: Verify that chosen model name is a valid identifier
The model name passed to (or deduced by) `sbml2amici` is used as python and swig module name, and there can't be an arbitrary string. Using invalid identifiers results in rather cryptic error messages (https://github.com/ICB-DCM/AMICI/issues/923#issuecomment-579348130). Therefore, check the model name early.

dweindl:
…ier (Closes #928)

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -943,7 +943,7 @@ class ODEModel:
     def ncl(self) -> int:
         """
         Number of conservation laws.
-        
+
         :return:
             number of conservation laws
         """
@@ -1026,7 +1026,7 @@ class ODEModel:
         Returns (and constructs if necessary) the formulas for a symbolic
         entity.
 
-        :param name: 
+        :param name:
             name of the symbolic variable
 
         :return:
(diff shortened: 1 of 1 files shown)


python/amici/ode_export.py
```diff
@@ -2803,3 +2809,20 @@ def csc_matrix(matrix: sp.Matrix,
...
+    digits and underscores, and must not start with a digit.
+
+    Arguments:
+        x: string to check
+
+    Returns:
+        ``True`` if valid, ``False`` otherwise
+    """
+
+    return re.match(r'^[a-zA-Z_]\w*$', x) is not None
```
FFroehlich: `re.match` only searches at beginning of string
```suggestion
    return re.search(r'^[a-zA-Z_]\w*$', x) is not None
```

dweindl: But we explicitly match beginning and end (^, $)

FFroehlich: ah, right :)

PR #1006, opened 2020-03-12
Fix coverage report from github actions

FFroehlich:


FFroehlich:
Currently fails due to https://github.com/codecov/codecov-action/issues/61

FFroehlich:
@dweindl okay coverage is working now (https://codecov.io/github/ICB-DCM/AMICI/commit/787373a4b50f8e083d4bc51ab5d0e01d6bb36343), its only covering the test-cases though. Thats probably not what you intended to do right? Probably needs to switch to in source install and change coverage source?

dweindl:
> @dweindl okay coverage is working now (https://codecov.io/github/ICB-DCM/AMICI/commit/787373a4b50f8e083d4bc51ab5d0e01d6bb36343),

:+1: Thanks for fixing

> its only covering the test-cases though. Thats probably not what you intended to do right? 

No, it isn't.

> Probably needs to switch to in source install and change coverage source?

Yeah, didn't think that far. Should also add lcov...

PR #1018, opened 2020-03-24
fix_typehints petab_objective

FFroehlich:


FFroehlich:
> what is actually the difference between Model and ModelPtr?

ModelPtr is the swig binding for a smart pointer `std::unique_ptr< amici::Model >`. Should have no effect at all, which means that we have to always account for the fact it could be either of both.

PR #1023, opened 2020-03-28
Feature solverhdf5

FFroehlich:


python/amici/__init__.py
```diff
@@ -223,3 +226,28 @@ def runAmiciSimulations(model, solver, edata_list, failfast=True,
                                                    failfast,
                                                    num_threads)
     return [numpy.ReturnDataView(r) for r in rdata_ptr_list]
+
+
+def readSolverSettingsFromHDF5(*args) -> None:
```
yannikschaelte: why not explicit arguments? + docstring what is to be passed?

FFroehlich: Hmm, the function is overloaded (allows passing file either as `str` or `H5::H5File`). In the RTD documentation this is properly processed but looks like tools such as PyCharm do indeed not have proper typhints+docstrings. Will fix as the arguments are the same across overloads.

tests/testHDF5.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
+
+"""Miscellaneous AMICI Python interface tests"""
```
dweindl:
```suggestion
"""AMICI HDF5 I/O tests"""
```

tests/testHDF5.py
```diff
@@ -0,0 +1,97 @@
...
+import amici
+import sys
+import copy
+import os
+import unittest
+import importlib
+
+
+class TestAmiciHDF5(unittest.TestCase):
+    """TestCase class various AMICI Python interface functions"""
```
dweindl:
```suggestion
    """TestCase class for AMICI HDF5 I/O"""
```

src/hdf5.cpp
```diff
@@ -513,6 +513,136 @@ bool attributeExists(H5::H5Object const& object,
...
+    H5LTset_attribute_double(file.getId(), hdf5Location.c_str(),
+                             "atol_fsa", &dbuffer, 1);
+    
+    dbuffer = solver.getRelativeToleranceFSA();
+    H5LTset_attribute_double(file.getId(), hdf5Location.c_str(),
+                             "rtol_fsa", &dbuffer, 1);
+    
+    dbuffer = solver.getAbsoluteToleranceB();
+    H5LTset_attribute_double(file.getId(), hdf5Location.c_str(),
+                             "rtolB", &dbuffer, 1);
```
dweindl:
```suggestion
                             "atolB", &dbuffer, 1);
```

src/hdf5.cpp
```diff
@@ -513,6 +513,136 @@ bool attributeExists(H5::H5Object const& object,
...
+    
+    dbuffer = solver.getRelativeToleranceB();
+    H5LTset_attribute_double(file.getId(), hdf5Location.c_str(),
+                             "atolB", &dbuffer, 1);
```
dweindl:
```suggestion
                             "rtolB", &dbuffer, 1);
```

src/hdf5.cpp
FFroehlich: oops

Issue #763, opened 2019-07-31
Python SBML import: add support for compartment assignment rules


dweindl:

cemale:
This addition would definitely be very useful, biologically relevant, and necessary. I would very much appreciate making the compartment size variable (i.e. dependent on a specific species amount).

FFroehlich:
Hmm, I believe @matthiaskoenig was also interested in this feature. Shouldn't be too difficult to implement this, so I will give this a try.

matthiaskoenig:
Yes, definitely interested. I have a few models with variable compartment sizes (i.e. `constant=false` compartments and assignment rules which calculate the compartment sizes. Only normal assignment rules no rate rules).
This should not be too complicated to implement. Depending on if volumes are set via
- rate rule (`dv/dt=rule`), a new state variable must be tracked
- assignment rule (`v=rule` , depending on state variables), a derived quantity must be calculated from state variables, parameters, ...

If you have any questions about the SBML part let me know.

dweindl:
Support for compartment assignment rules has been added a while ago in #1036. Closing.

PR #1002, opened 2020-03-09
refactor rdata

FFroehlich:
Goal is to completely remove any access to rdata in forward and backward problem. This will allow us to

- reduce memory impact of rdata since we don't need to prefill any results but can compute everything ad hoc (given a model and solver instance)
- implement hierarchical optimization in runAmiciSimulations with adjoints since we can seperately loop over forward problems, then compute noise parameters and then loop over backward problems.
- better debugging of integration failures as we can implement a more fine-grained access to the solution data in solvers after integration failure.

Both topics should and will be adressed in later PRs. May also split the rdata refactoring into multiple PRs.

PR #1036, opened 2020-04-06
Variable compartments

dilpath:
Related issue: #763 

This isn't ready for review yet. It's uploaded so @jvanhoefer can compare it with his rate rules implementation. Any comments on the design/implementation are very welcome!

- Assignment and rate rules are implemented for compartments.
- Compartments with rate rules are implemented as fake species, which includes modifications to the stoichiometric matrix/flux vector.
- Compartments can be used in other assignment and rate rules, and reactions.

The SBML test suite may not test this pull request well, as many of the tests related to variable compartments are skipped for other reasons (e.g. species rate rules in the SBML). I'll need to do some additional testing. I also raise an `SBMLException` for SBML files that have no reactions, assignment rules, or rate rules (as there is no source of time-dependence), as models without these dynamics caused some issues for this PR. However, AMICI seemed to "simulate" these models before, so I'll have another look.

Currently, the following SBML test suite tests fail: 140, 525-527. Test 140 might fail in the concentration to amount conversion in `testSBMLSuite.py` (the simulated values are multiples of the expected values).

Tests 525-527 fail with the following error:
```
SBMLTest00527_xdot.cpp: In function ‘void xdot_SBMLTest00527(realtype*, realtype, const realtype*, const realtype*, const realtype*, const realtype*, const realtype*)’:
SBMLTest00527_xdot.cpp:13:28: error: invalid operands of types ‘double’ and ‘const realtype*’ {aka ‘const double*’} to binary ‘operator/’
     xdot[0] = -50.0*flux_r0/k;
               ~~~~~~~~~~~~~^~
```


FFroehlich:
> Tests 525-527 fail with the following error:
> 
> ```
> SBMLTest00527_xdot.cpp: In function ‘void xdot_SBMLTest00527(realtype*, realtype, const realtype*, const realtype*, const realtype*, const realtype*, const realtype*)’:
> SBMLTest00527_xdot.cpp:13:28: error: invalid operands of types ‘double’ and ‘const realtype*’ {aka ‘const double*’} to binary ‘operator/’
>      xdot[0] = -50.0*flux_r0/k;
>                ~~~~~~~~~~~~~^~
> ```

This issue sounds like sbml is using `k` as parameter name (this is mapped via macros in `p.h`), which cause trouble in the generated c++ code as the pointer to fixedParameter values is also called `k`. `SbmlImporter._clean_reserved_symbols()` is supposed to take care of this. My guess is that `SbmlImporter._replace_in_all_expressions()` needs to be updated to also apply to newly added member variables in the importer class.

python/amici/sbml_import.py
```diff
@@ -742,16 +763,52 @@ def _process_rules(self) -> None:
...
+                    assignments[str(variable)] = formula
+                    #continue # avoid simplifying substitution of the compartment expression, to avoid possibly circular substitution
+                elif rule.getTypeCode() == sbml.SBML_RATE_RULE:
+                    self.compartment_rules[variable] = {
+                            'type_code': sbml.SBML_RATE_RULE,
+                            'formula': formula,
+                            'v0': self.compartment_volume[list(self.compartment_symbols).index(variable)]
+                        }
+
+                    self.symbols['species']['identifier'] = self.symbols['species']['identifier'].col_join(sp.Matrix([variable]))
```
FFroehlich, line 798: Would it be possible to directly implement this in `_process_species`? Would have the benefit that we won't have to worry about downstream effects later on.

python/amici/sbml_import.py
```diff
@@ -742,16 +763,52 @@ def _process_rules(self) -> None:
...
+                    self.symbols['species']['name'].append('')
+                    self.symbols['species']['value'] = self.symbols['species']['value'].col_join(sp.Matrix([variable]))
+                    #self.symbols['species']['value'] = self.symbols['species']['value'].col_join(sp.Matrix([self.compartment_rules[variable]['v0']]))
+                    self.species_index.update({sp.sstr(variable): len(self.species_index)}) #necessary?
+                    self.stoichiometric_matrix = self.stoichiometric_matrix.col_join(
```
FFroehlich, line 803: Extending like this is likely to be inefficient for larger matrices and the stoichiometric matrix can easily have multiple million entries. By moving the above code to `_process_species` this allocation is already partially adressed. I think it would be great to split this code in two, one that precomputes most of what is done here, but doesn't update `stoichiometric_matrix` and `flux_vector` yet. `_process_reactions()` could then correctly initialize the stoichiometric matrix based on the precomputed variables and this function would just update `stoichiometric_matrix` and `flux_vector` accordingly.

python/amici/sbml_import.py
```diff
@@ -1013,9 +1081,17 @@ def _replace_in_all_expressions(self,
             'species', 'observables',
         ]
         for symbol in symbols:
-            if symbol in self.symbols:
+            if symbol in self.symbols and old not in self.compartment_rules:
                 self.symbols[symbol]['value'] = \
                     self.symbols[symbol]['value'].subs(old, new)
+        if 'compartment_rules' in dir(self):
```
FFroehlich, line 1087: add emtpy initialization to constructor?

python/amici/ode_export.py
```diff
@@ -843,9 +847,65 @@ def import_from_sbml_importer(self,
...
+            elif x_id in si.compartment_symbols:
+                raise KeyError('This should not happen. There exists an ODE for an compartment that does not have an associated rate rule.')
+            v_name = si.species_compartment[x_index]
+            if v_name in si.compartment_rules:
+                if (si.compartment_rules[v_name]['type_code']
+                        == libsbml.SBML_ASSIGNMENT_RULE):
+                    v = si.compartment_rules[v_name]['formula']
+                    dv_dt = v.diff(si.amici_time_symbol)
+                    dv_dx = v.diff(x_id)
+                    # not sure if parentheses here are necessary...
```
FFroehlich, line 879: not necessary :)

python/amici/ode_export.py
```diff
@@ -843,9 +847,65 @@ def import_from_sbml_importer(self,
...
+                    return ((x_Sw) - (dv_dt)*(x_id))/((dv_dx)*(x_id) + (v))
+                elif (si.compartment_rules[v_name]['type_code']
+                          == libsbml.SBML_RATE_RULE):
+                    t = si.amici_time_symbol
```
FFroehlich, line 883: this looks complicated, would be nice to have a short reminder (as comment in the code) why we need to do this and, if necessary, how we are doing it.

python/amici/ode_export.py
```diff
@@ -1636,6 +1696,9 @@ def _equation_from_component(self, name: str, component: str) -> None:
         self._eqs[name] = sp.Matrix(
             [comp.get_val() for comp in getattr(self, component)]
         )
+
+        if name == 'x0':
```
FFroehlich, line 1700: I'm not sure I understand the full logic yet how this is implemented, but this shouldn't be handled here. After initialization, the ODEExporter should be completely independent of the SbmlImporter, since it can also handle other model types (currently only PYSB models).

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

python/amici/ode_export.py
dilpath: I've added a description in the code. I can add the derivation of the equations to the AMICI documentation, is there a suitable page already?

python/amici/sbml_import.py
dilpath: The implementation of rate rules no longer involves `stoichiometric_matrix`/`flux_vector`.

dilpath:
The latest commit includes implementations of assignment and rate rules for species ( @jvanhoefer ). Briefly, rate rules are implemented as ODEs, and assignment rules are implemented as observables.

Tests 525-527 (and 140) now pass, after re-arranging the previous implementation according to Fabian's feedback.

Test 1498 now fails. I've added an [exception](https://github.com/ICB-DCM/AMICI/pull/1036/files#diff-d16483a28b0305562f35450153750757R840) (currently commented out) to handle it, but that also results in ~60 tests, that the develop branch previously passed, now being skipped.

Only test 1223 appears to involve a variable compartment with an assignment rule (so additional testing would be good).

Any feedback, especially about whether the current implementation is reasonable, would be appreciated!

python/amici/sbml_import.py
```diff
@@ -492,6 +557,141 @@ def get_species_initial(index, conc):
...
+        # compartments with rules are replaced with constants in the relevant
+        # equations during the _replace_in_all_expressions call inside
+        # _process_rules
+        for rule in rules:
+            ### Taken from _process_rules
+            if rule.getFormula() == '':
+                continue
+            variable = sp.sympify(rule.getVariable(),
+                                  locals=self.local_symbols)
+            # avoid incorrect parsing of pow(x, -1) in symengine
```
FFroehlich, line 577:
```suggestion
```

The code below fixes logical expressions in sbml and makes them parseable by symy. Looks like that comment went stale

python/amici/sbml_import.py
```diff
@@ -492,6 +557,141 @@ def get_species_initial(index, conc):
...
+            ###
+
+            # Species rules are processed first, to avoid processing
+            # compartments twice (as compartments with rate rules are
+            # implemented as species). Could also be avoided with a
+            # `not in self.compartment_rate_rules` condition.
+            if variable in self.symbols['species']['identifier']:
+                if rule.getTypeCode() == sbml.SBML_ASSIGNMENT_RULE:
+                    # Handled in _process_rules and _process_observables.
+                    # TODO: set a species initial value to its assignment rule?
```
FFroehlich, line 592: setting initial value shouldn't be necessary here, as its defined by the assignment rule.

python/amici/sbml_import.py
```diff
@@ -492,6 +557,141 @@ def get_species_initial(index, conc):
...
+                    # values (see SBML L3V2 manual, Section 3.4.8).
+                    # Priority appears to be above InitialAssignment.
+                    self.compartment_volume[list(
+                        self.compartment_symbols).index(variable)] = formula
+                elif rule.getTypeCode() == sbml.SBML_RATE_RULE:
+                    self.add_d_dt(
+                        formula,
+                        variable,
+                        self.compartment_volume[list(
+                            self.compartment_symbols).index(variable)],
```
FFroehlich, line 617: Personally not a big fan of this indentation style, makes it difficult to visually keep track of where brackets start/end. 
```suggestion
                            self.compartment_symbols
                         ).index(variable)],
```

python/amici/sbml_import.py
```diff
@@ -1001,21 +1226,62 @@ def _replace_in_all_expressions(self,
...
-                self.symbols[symbol]['value'] = \
-                    self.symbols[symbol]['value'].subs(old, new)
+                # Initial species values that are specified as amounts need to
+                # be divided by their compartment volume to obtain
+                # concentration. The condition below ensures that species
+                # initial amount is divided by the initial compartment size,
+                # and not the expression for a compartment assignment rule.
+                if symbol == 'species' and (
+                        old in self.compartment_assignment_rules):
+                    self.symbols[symbol]['value'] = \
```
FFroehlich, line 1259: this is pretty nasty to read. introduce intermediate variables?

FFroehlich:
the sphinx error is already fixed in #1002, which will hopefully be merged soon

FFroehlich:
> Any feedback, especially about whether the current implementation is reasonable, would be appreciated!

Overall pretty happy with the implementation! :)

FFroehlich:
> Test [1498](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml) now fails. I've added an [exception](https://github.com/ICB-DCM/AMICI/pull/1036/files#diff-d16483a28b0305562f35450153750757R840) (currently commented out) to handle it, but that also results in ~60 tests, that the develop branch previously passed, now being skipped. The reason why Test 1498 fails is related to its use of the `S2_sr` species ID, which is not defined in its `listOfSpecies`.

How does 1498 fail? Is it a problem with about the correctness of the solution or is it an issue with matching simulation results agains expected results?

dilpath:
> How does 1498 fail? Is it a problem with about the correctness of the solution or is it an issue with matching simulation results against expected results?

Below is the error. I've had another look at the [SBML](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L42), the issue might be that `S2_sr` is used in the rate rule for the species `S1`, but `S2_sr` is not declared as a species/parameter. Instead, its existence is declared in an initial assignment rule.

```
SBMLTest01498_xdot.cpp: In function ‘void xdot_SBMLTest01498(realtype*, realtype, const realtype*, const realtype*, const realtype*, const realtype*, const realtype*)’:
SBMLTest01498_xdot.cpp:13:15: error: ‘S2_sr’ was not declared in this scope
     xdot[0] = S2_sr;
               ^~~~~
```

FFroehlich:
> > How does 1498 fail? Is it a problem with about the correctness of the solution or is it an issue with matching simulation results against expected results?
> 
> Below is the error. I've had another look at the [SBML](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L42), the issue might be that `S2_sr` is used in the rate rule for the species `S1`, but `S2_sr` is not declared as a species/parameter. Instead, its existence is declared in an initial assignment rule.
> 
> ```
> SBMLTest01498_xdot.cpp: In function ‘void xdot_SBMLTest01498(realtype*, realtype, const realtype*, const realtype*, const realtype*, const realtype*, const realtype*)’:
> SBMLTest01498_xdot.cpp:13:15: error: ‘S2_sr’ was not declared in this scope
>      xdot[0] = S2_sr;
>                ^~~~~
> ```

Oh yeah, that makes sense. Sounds like an edge case I'm not too enthusiastic to support in AMICI. Will propose a way to catch this more specifically.

FFroehlich:
> > > How does 1498 fail? Is it a problem with about the correctness of the solution or is it an issue with matching simulation results against expected results?
> > 
> > 
> > Below is the error. I've had another look at the [SBML](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L42), the issue might be that `S2_sr` is used in the rate rule for the species `S1`, but `S2_sr` is not declared as a species/parameter. Instead, its existence is declared in an initial assignment rule.
> > ```
> > SBMLTest01498_xdot.cpp: In function ‘void xdot_SBMLTest01498(realtype*, realtype, const realtype*, const realtype*, const realtype*, const realtype*, const realtype*)’:
> > SBMLTest01498_xdot.cpp:13:15: error: ‘S2_sr’ was not declared in this scope
> >      xdot[0] = S2_sr;
> >                ^~~~~
> > ```
> 
> Oh yeah, that makes sense. Sounds like an edge case I'm not too enthusiastic to support in AMICI. Will propose a way to catch this more specifically.

Just checked, and we should actually already be able to handle this situation. The respective code is at `https://github.com/ICB-DCM/AMICI/blob/96df8c473db8ac53c7db3099d78f6948ea5048f7/python/amici/sbml_import.py#L808`. In this case "id" is used to specify the stoichiometry through an initial assignment. The way this should be resolved is that the reference is supposed to be replaced by the numerical value of the initial assignment. However it looks like`get_element_from_assignment` is not able to do that and instead of the (here undefined) symbol id is used (which is sometimes necessary as this might also be defined through other means).

dilpath:
> > > > How does 1498 fail? Is it a problem with about the correctness of the solution or is it an issue with matching simulation results against expected results?
> > > 
> > > 
> > > Below is the error. I've had another look at the [SBML](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L42), the issue might be that `S2_sr` is used in the rate rule for the species `S1`, but `S2_sr` is not declared as a species/parameter. Instead, its existence is declared in an initial assignment rule.
> > > ```
> > > SBMLTest01498_xdot.cpp: In function ‘void xdot_SBMLTest01498(realtype*, realtype, const realtype*, const realtype*, const realtype*, const realtype*, const realtype*)’:
> > > SBMLTest01498_xdot.cpp:13:15: error: ‘S2_sr’ was not declared in this scope
> > >      xdot[0] = S2_sr;
> > >                ^~~~~
> > > ```
> > 
> > 
> > Oh yeah, that makes sense. Sounds like an edge case I'm not too enthusiastic to support in AMICI. Will propose a way to catch this more specifically.
> 
> Just checked, and we should actually already be able to handle this situation. The respective code is at `https://github.com/ICB-DCM/AMICI/blob/96df8c473db8ac53c7db3099d78f6948ea5048f7/python/amici/sbml_import.py#L808`. In this case "id" is used to specify the stoichiometry through an initial assignment. The way this should be resolved is that the reference is supposed to be replaced by the numerical value of the initial assignment. However it looks like`get_element_from_assignment` is not able to do that and instead of the (here undefined) symbol id is used (which is sometimes necessary as this might also be defined through other means).

Are we talking about the same thing? My first guess at the source of the issue was the [reaction](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L42), but i now think it's the [rate rule](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L25) (the implementation of rate rules here does not involve `_process_reactions`). If this is the case, then the simplest solution might be to add code to the new `add_d_dt` method to explicitly substitute initial assignment values, if the initial assignment symbol is not a time-dependent component (a compartment or species).

FFroehlich:
> > > > > How does 1498 fail? Is it a problem with about the correctness of the solution or is it an issue with matching simulation results against expected results?
> > > > 
> > > > 
> > > > Below is the error. I've had another look at the [SBML](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L42), the issue might be that `S2_sr` is used in the rate rule for the species `S1`, but `S2_sr` is not declared as a species/parameter. Instead, its existence is declared in an initial assignment rule.
> > > > ```
> > > > SBMLTest01498_xdot.cpp: In function ‘void xdot_SBMLTest01498(realtype*, realtype, const realtype*, const realtype*, const realtype*, const realtype*, const realtype*)’:
> > > > SBMLTest01498_xdot.cpp:13:15: error: ‘S2_sr’ was not declared in this scope
> > > >      xdot[0] = S2_sr;
> > > >                ^~~~~
> > > > ```
> > > 
> > > 
> > > Oh yeah, that makes sense. Sounds like an edge case I'm not too enthusiastic to support in AMICI. Will propose a way to catch this more specifically.
> > 
> > 
> > Just checked, and we should actually already be able to handle this situation. The respective code is at `https://github.com/ICB-DCM/AMICI/blob/96df8c473db8ac53c7db3099d78f6948ea5048f7/python/amici/sbml_import.py#L808`. In this case "id" is used to specify the stoichiometry through an initial assignment. The way this should be resolved is that the reference is supposed to be replaced by the numerical value of the initial assignment. However it looks like`get_element_from_assignment` is not able to do that and instead of the (here undefined) symbol id is used (which is sometimes necessary as this might also be defined through other means).
> 
> Are we talking about the same thing? My first guess at the source of the issue was the [reaction](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L42), but i now think it's the [rate rule](https://github.com/sbmlteam/sbml-test-suite/blob/master/cases/semantic/01498/01498-sbml-l3v2.xml#L25) (the implementation of rate rules here does not involve `_process_reactions`). If this is the case, then the simplest solution might be to add code to the new `add_d_dt` method to explicitly substitute initial assignment values, if the initial assignment symbol is not a (possibly) time-dependent component (a compartment or species).

Oh sorry didn't catch that. I agree, explicitly substituting the initial assignment value is the right way to go.

FFroehlich:
The _Initial_Assignment is, by definition, time independent. If there are any time-dependent quantities in the formula, those should be replaced with their respective initial values

dilpath:
> The **Initial**Assignment is, by definition, time independent. If there are any time-dependent quantities in the formula, those should be replaced with their respective initial values

Makes sense, the reason why I wrote "if the initial assignment symbol is not a (possibly) time-dependent component (a compartment or species)" is that, if the symbol of the initial assignment matches a time-dependent component then, if that symbol is used in a rate rule, the symbol in the rate rule should not be subtituted.

Example: if `S1(t)` and `S2(t)` are species, and there is a rate rule `dS1(t)/dt = S2(t)`, and an initial assignment that specifies `S2(0) = 5`, then `dS1(t)/dt = 5` is not true, unless S2 is not time-dependent. So in this case the symbol for the initial assignment should not be substituted.

So my proposed solution would be to substitute the initial assignment, if its symbol is not a species or compartment (a sufficient condition, but not a necessary one... easier to code up but I can be very specific if you prefer and check whether the compartment/species is actually time-dependent).

FFroehlich:
> > The **Initial**Assignment is, by definition, time independent. If there are any time-dependent quantities in the formula, those should be replaced with their respective initial values
> 
> Makes sense, the reason why I wrote "if the initial assignment symbol is not a (possibly) time-dependent component (a compartment or species)" is that, if the symbol of the initial assignment matches a time-dependent component then, if that symbol is used in a rate rule, the symbol in the rate rule should not be subtituted.
> 
> Example: if `S1(t)` and `S2(t)` are species, and there is a rate rule `dS1(t)/dt = S2(t)`, and an initial assignment that specifies `S2(0) = 5`, then `dS1(t)/dt = 5` is not true, unless S2 is not time-dependent. So in this case the symbol for the initial assignment should not be substituted.
> 
> So my proposed solution would be to substitute the initial assignment, if its symbol is not a species or compartment (a sufficient condition, but not a necessary one... easier to code up but I can be very specific if you prefer and check whether the compartment/species is actually time-dependent).

Correct, but the replacement should only be performed if the respective symbol was not defined in the first place and not generally for all assignment rules. The crucial thing I want to point out is that if you have `dS1(t)/dt = S2_sr(t)` and your initial assignment is `S2_sr = S1`, you end up with `dS1(t)/dt = S1(0)` and not  `dS1(t)/dt = S1(t)`.

dilpath:
> > > The **Initial**Assignment is, by definition, time independent. If there are any time-dependent quantities in the formula, those should be replaced with their respective initial values
> > 
> > 
> > Makes sense, the reason why I wrote "if the initial assignment symbol is not a (possibly) time-dependent component (a compartment or species)" is that, if the symbol of the initial assignment matches a time-dependent component then, if that symbol is used in a rate rule, the symbol in the rate rule should not be subtituted.
> > Example: if `S1(t)` and `S2(t)` are species, and there is a rate rule `dS1(t)/dt = S2(t)`, and an initial assignment that specifies `S2(0) = 5`, then `dS1(t)/dt = 5` is not true, unless S2 is not time-dependent. So in this case the symbol for the initial assignment should not be substituted.
> > So my proposed solution would be to substitute the initial assignment, if its symbol is not a species or compartment (a sufficient condition, but not a necessary one... easier to code up but I can be very specific if you prefer and check whether the compartment/species is actually time-dependent).
> 
> Correct, but the replacement should only be performed if the respective symbol was not defined in the first place and not generally for all assignment rules. The crucial thing I want to point out is that if you have `dS1(t)/dt = S2_sr(t)` and your initial assignment is `S2_sr = S1`, you end up with `dS1(t)/dt = S1(0)` and not `dS1(t)/dt = S1(t)`.

Ah right, I misunderstood you earlier. Yes, the initial assignment will need to be evaluated at `t=0`. Thanks for pointing that out!

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

python/amici/sbml_import.py
```diff
@@ -492,6 +557,141 @@ def get_species_initial(index, conc):
...
+            variable0: Union[float, sp.Expr],
+            component: int,
+            name: str = None
+    ) -> None:
+        '''
+        Creates or modifies species, to implement rate rules for compartments
+        and species, respectively.
+
+        Arguments
+        ---------
```
FFroehlich, line 636: Sphinx can't deal with numpy style annotations since swig outputs standard python docstrings.

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

dilpath:
The branch now contains a very simple implementation of the solution to the issue described in the previous comments (relating to SBML test suite case 1498). If there is a species reference that is defined by an initial assignment and then used in a rate rule, then an `SBMLException` is raised if the initial assignment contains symbols that may require evaluation at `t=0`. If the initial assignment is some real number, then the substitution is made.

This is sufficient to pass test 1498. The simple implementation was chosen because I'm not sure if the more complicated implementation should be considered an expected use case for AMICI. Let me know if you disagree. In any case, I have included a description for what the more complicated implementation might look like, as a comment in the code.

Now that the SBML test suite cases do not produce errors, an additional 196 cases pass, compared to the `develop` branch. I removed the `SBMLException` for SBML files that have no reactions, assignment rules, or rate rules, which is described in the first post (I'm not sure what originally caused, or resolved this issue).

FFroehlich:
> The branch now contains a very simple implementation of the solution to the issue described in the previous comments (relating to SBML test suite case 1498). If there is a species reference that is defined by an initial assignment and then used in a rate rule, then an `SBMLException` is raised if the initial assignment contains symbols that may require evaluation at `t=0`. If the initial assignment is some real number, then the substitution is made.
> 
> This is sufficient to pass test 1498. The simple implementation was chosen because I'm not sure if the more complicated implementation should be considered an expected use case for AMICI. Let me know if you disagree. In any case, I have included a description for what the more complicated implementation might look like, as a comment in the code.
> 
> Now that the SBML test suite cases do not produce errors, an additional 196 cases pass, compared to the `develop` branch. I removed the `SBMLException` for SBML files that have no reactions, assignment rules, or rate rules, which is described in the first post (I'm not sure what originally caused, or resolved this issue).

Great! feel free to merge this PR with "squash and merge".

dweindl:
> Now that the SBML test suite cases do not produce errors, an additional 196 cases pass, compared to the develop branch.

Great! Maybe you can directly update https://github.com/ICB-DCM/AMICI/blob/master/README.md#status-of-sbml-support-in-python-amici accordingly.

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

dilpath:
> > Now that the SBML test suite cases do not produce errors, an additional 196 cases pass, compared to the develop branch.
> 
> Great! Maybe you can directly update https://github.com/ICB-DCM/AMICI/blob/master/README.md#status-of-sbml-support-in-python-amici accordingly.

Done in the latest commit (small note: when I run the SBML test suite, I only see 1779 tests, not 1780).

Now the PEtab test suite fails in the [performance test](https://github.com/ICB-DCM/AMICI/blob/develop/tests/performance/check_time.sh) with these [reference values](https://github.com/ICB-DCM/AMICI/blob/develop/tests/performance/reference.yml): `TOO LONG: 140.40 > 140`. Running the test again might result in a pass.. should the reference value for `install_sdist` be increased?

FFroehlich:
> > > Now that the SBML test suite cases do not produce errors, an additional 196 cases pass, compared to the develop branch.
> > 
> > 
> > Great! Maybe you can directly update https://github.com/ICB-DCM/AMICI/blob/master/README.md#status-of-sbml-support-in-python-amici accordingly.
> 
> Done in the latest commit (small note: when I run the SBML test suite, I only see 1779 tests, not 1780).
> 
> Now the PEtab test suite fails in the [performance test](https://github.com/ICB-DCM/AMICI/blob/develop/tests/performance/check_time.sh) with these [reference values](https://github.com/ICB-DCM/AMICI/blob/develop/tests/performance/reference.yml): `TOO LONG: 140.40 > 140`. Running the test again might result in a pass.. should the reference value for `install_sdist` be increased?

Yeah `install_sdist` has been failing repeatedly over last couple of weeks, feel free to increase this to 150.

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/amici/sbml_import.py  8
         

Clones added
============
- python/amici/sbml_import.py  4
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5316682&bid=17399336)

PR #1039, opened 2020-04-16
update to sundials 5.2.0, fixes #679

### LINKED ISSUE #679: Update to sundials 5.2.0
See https://computation.llnl.gov/sites/default/files/public/cv_guide-dev.pdf

FFroehlich:


FFroehlich:
only minor changes to a couple header files, we are now hopefully also able to directly use sundials without any changes.

python/sdist/setup_clibs.py
```diff
@@ -43,23 +43,17 @@ def get_sundials_sources() -> List[str]:
...
         os.path.join('src', 'sundials', 'sundials_dense.c'),
         os.path.join('src', 'sundials', 'sundials_nvector.c'),
         os.path.join('src', 'sundials', 'sundials_version.c'),
-        os.path.join('src', 'sundials', 'sundials_spfgmr.c'),
-        os.path.join('src', 'sundials', 'sundials_sparse.c'),
         os.path.join('src', 'sundials', 'sundials_iterative.c'),
         os.path.join('src', 'sundials', 'sundials_nonlinearsolver.c'),
         os.path.join('src', 'sundials', 'sundials_linearsolver.c'),
         os.path.join('src', 'sundials', 'sundials_band.c'),
+        os.path.join('src', 'sundials', 'sundials_futils.c'),
```
dweindl, line 56: I guess, changes from python/sdist/setup_clibs.py will need to be applied to matlab compilation functions as well. Maybe at some point introduce a list of source files that can be used from both python and matlab...

FFroehlich: good point!

PR #1043, opened 2020-04-18
add test for data replicates, fixes #433

### LINKED ISSUE #433: Confirm that sensitivities are computed correctly for replicate measurements
Confirm that sensitivities are computed correctly for replicate measurements -> See discussion in #262 

Ideally add a test case to one of the existing models.

FFroehlich:


FFroehlich:
> Somewhat redundant with `tests/cpputest/jakstat_adjoint/tests1.cpp`, but maybe better to have one test case more than less...

Ah, wasn't aware of that test. This one also inclues inf datapoints from postequilibration (probably why coverage is slightly increased).

PR #1047, opened 2020-04-19
Constant species to conservation laws

paulstapor:
This PR should implement constant species in SBML models by using the conservation law interface. A unit test is added.


python/amici/sbml_import.py
```diff
@@ -240,6 +243,11 @@ def sbml2amici(self,
             If True, compile the generated Python package,
             if False, just generate code.
 
+        :param compute_conservation_laws:
+            if set to true, conservation laws are automatically computed and
+            applied such that the state-jacobian of the ODE right-hand-side has
+            full rank. This option should be set to True when using the newton
+            algorithm to compute steadystates
```
FFroehlich, line 250:
```suggestion
            algorithm to compute steadystate sensititivities.
```

python/amici/sbml_import.py
```diff
@@ -989,6 +998,67 @@ def replace_assignments(formula: str) -> sp.Basic:
...
+            if ode_model.state_is_constant(ix):
+                target_state = ode_model._states[ix]._identifier
+                total_abundance = ode_model._states[ix].get_val()
+                conservation_laws.append({
+                    'state': target_state,
+                    'total_abundance': total_abundance,
+                    'state_expr': total_abundance,
+                    'abundance_expr': target_state,
+                })
+                self._replace_in_all_expressions(target_state, total_abundance)
```
FFroehlich, line 1039: Isn't this automatically handled by the ConservationLaw implementation?

python/tests/test_compare_conservation_laws_sbml.py
```diff
@@ -0,0 +1,141 @@
...
+    model_without_cl.setSteadyStateSensitivityMode(
+        amici.SteadyStateSensitivityMode_simulationFSA
+    )
+
+    # run simulations
+    rdata_cl = get_results(model_with_cl, edata=edata_fixture)
+    rdata = get_results(model_without_cl, edata=edata_fixture)
+
+    # compare preequilibrated states
+    assert np.isclose(rdata['x'], rdata_cl['x']).all()
```
FFroehlich, line 110:
```suggestion
    for field in ['x', 'x_ss', 'llh']
        assert np.isclose(rdata[field], rdata_cl[field]).all(), field
```

python/tests/test_compare_conservation_laws_sbml.py
```diff
@@ -0,0 +1,141 @@
...
+    assert np.isclose(rdata['llh'], rdata_cl['llh']).all()
+
+    # ----- compare simulations wo edata, sensi = 1, states and sensis -------
+
+    # run simulations
+    rdata_cl = get_results(model_with_cl, edata=edata_fixture, sensi_order=1)
+    rdata = get_results(model_without_cl, edata=edata_fixture, sensi_order=1)
+
+    # compare state sensitivities with edata and preequilibration
+    assert np.isclose(rdata['x'], rdata_cl['x']).all()
```
FFroehlich, line 121: see above

python/tests/test_compare_conservation_laws_sbml.py
```diff
@@ -0,0 +1,141 @@
...
+        'observable_product': {'name': '', 'formula': 'product'},
+        'observable_substrate': {'name': '', 'formula': 'substrate'},
+    }
+    sigmas = {'observable_product': 1.0, 'observable_substrate': 1.0}
+
+    # wrap models with and without conservations laws
+    sbml_importer.sbml2amici(model_name_cl,
+                             model_output_dir_cl,
+                             observables=observables,
+                             constantParameters=constant_parameters,
```
FFroehlich, line 46: please use `constant_parameters` instead of `constantParameters` (will be deprecated at some point)

python/amici/sbml_import.py
paulstapor: Not 100% sure, but I think not. It didn't work for me without this line... But I can retry...

python/tests/test_compare_conservation_laws_sbml.py
paulstapor: done already ;)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  15
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

python/amici/sbml_import.py
FFroehlich: Looks like it wasn't necessary in https://github.com/ICB-DCM/AMICI/blob/390305f98c96ea505cf52884991774e7dfefdeff/python/amici/pysb_import.py#L359

python/tests/test_compare_conservation_laws_sbml.py
FFroehlich: Oh I mean also as argument name for `sbml2amici`, `constantParameters` still exists for backwards compatibility.

paulstapor: Got it a minute later... :D
Done locally, need to push

python/amici/sbml_import.py
paulstapor: douvble-checked, it is necessary.
pysb_import is constructed somewhat differently, but also calls a similar functionality in _flatten_conservation_laws. There, also substitutions are called...
So, I could either use replace_in_all_expressions, or copy-paste flatten_conservation_laws and adapt it, if necessary...
If you prefer the latter, I can doe the latter. But the current solution does the job without code duplication, so I used this one.

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  11
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

python/amici/sbml_import.py
FFroehlich: but `_flatten_conservation_laws` only operates on the conservation laws to remove interdependencies and doesn't modify any other expressions.

I think this implementation is conceptually not correct. The right way to go is to use `target_state` as `state_expr`. What is currently being done is hardcopying the total abundance everywhere where the state occurs. You may get away with that, but it's not in the spirit of the conservation law implementation and might prevent conservation laws from correctly updating after reinitialization with updated fixedParameters (may be worthwhile to add a corresponding test).

FFroehlich: Although I have to double check since its implemented the same way in `_add_conservation_for_constant_species` 🤔

FFroehlich: Okay I think the issue is more about the way total abundance is specified here. I don't think you want to pass the value here, but rather a new symbolic variable such as `total_abundance = sp.Symbol(f'tcl_{target_state}')

paulstapor: The point is also a bit that I didn't totally get where to put "target_state" and where "state_expression". I tried different things, none of which worked without the "replace in all expression". I feel like state_expression should actually be something totally different, neither target_state nor state_expression, but rather an independent symbolic...

I felt like it's kinda difficult to get those four lines correct because I don't fully understand what is finally replaced where in add_conservation_law. It's pretty hard to develop an intuition for that, as the overall idea, how the code should be affected, is somewhat missing for me... :/

paulstapor: > Okay I think the issue is more about the way total abundance is specified here. I don't think you want to pass the value here, but rather a new symbolic variable such as `total_abundance = sp.Symbol(f'tcl_{target_state}')`

I think I even tried that one without success... 😅 
I'd be really thankful for some help on those lines, as I feel a little lost there...
I can try your suggestion, but I think something would still be missing...

FFroehlich: Sounds like I need to add more documentation to `add_conservation_law`, will do.

paulstapor: yep... The problem is that the old state-expression is *not* replaced in the flux vector by the current code. This causes that the old state expression still shows up in cpp-files, which don't know the identifier any more, resulting finally in a compiler error.
I somehow need to replace the old state expression in the flux vector, otherwise it seems like it won't work.
I tried quite some permutations and adding other symbolic expressions now...

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  11
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

python/amici/sbml_import.py
FFroehlich: Added documentation. I think the key issue that currently isn't addressed is that `add_conservation_law` adds an Expression to the model to compute the value of the conserved quantity. This means that `ode_model._eqs['w'] = self.flux_vector` no longer works but that flux_vector should be translated into expressions.

paulstapor: Okay, the doc helps, that's nice!
However, the problem persists, that the flux_vector needs to be changed...
Would a simple `subs(state, state_expr)` thrown on the flux vector do the trick? Or would something be missing?

paulstapor: And... we shouldn't forget: Things may look slightly different for real conservation laws, but this is actually only the interface for constant species. This code isn't even meant to capture actual conserved quantities across states...

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  11
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

python/amici/sbml_import.py
paulstapor: Ok, I changed the substitution. This is now the minimal version it needs to work.
Ich you want, I can still add a symbolic idnetifier in here, but this should anyway be a symbolic substitution...

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  11
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

python/amici/sbml_import.py
FFroehlich: I don't think the current implementation works with reinitialization of states/fixedParameters. I think the right way to go is to is

- [ ] pass the values as described above to `add_conservation_law`
- [ ] remove assignment to `ode_model._eqs['w']`, instead use `add_component` to add flux vector entries to the ode_model
- [ ] store the symbols used in `add_component` and reconsitute the flux vector (with symbolic identifiers), compute `xdot` based on those and update `dxdotdw` (add a couple zero rows in the right place).

paulstapor: > I don't think the current implementation works with reinitialization of states/fixedParameters. I think the right way to go is to is

But my unit test uses this reinitialization and works... (?)

FFroehlich: Okay then it probably works in most cases. Looking at the generated code, the part that is circumvented is there to avoid issues with larger expressions for conservation laws that get repeatedly differentiated and thereby cause and explosion of the size of derivatives.

Still not happy with the current implementation as it explicitly doesn't use the API as intended and will inevitably confuse the next person looking at that code.

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  18
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  18
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

python/amici/sbml_import.py
paulstapor: okay, then I'll change it. Code is supposed to stay maintainable... 👍

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  18
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

python/amici/sbml_import.py
paulstapor: Update on this one:
I will need to include the constants k into total_cl.cpp, as the conservation laws now need to be informed about constant parameters. Otherwise, conservation_laws depending on constant parameters won't work...

paulstapor: Some more questions to this...

> * [ ]  remove assignment to `ode_model._eqs['w']`, instead use `add_component` to add flux vector entries to the ode_model

Shall I add them as generic expressions, or introduce a new class "flux"?

> * [ ]  store the symbols used in `add_component` and reconsitute the flux vector (with symbolic identifiers), compute `xdot` based on those and update `dxdotdw` (add a couple zero rows in the right place).

Wait... This means that xdot is not S * w any more, and that would mean that I have to alter the stoichiometric matrix S, beyond just deleting the zero-rows, right?
I totally see the purpose: It would introduce the framework for implementing real conservation laws directly and I would already do half the job which is necessary to implement those (except computing the left kernel of S and identifying the most promising species to be replaced).

Two things on this:
Firstly: this is substantially more changing than I was planning to do right now. Not crucially problematic, but wanted to point it out. This was never the aim of this pull-request.
Secondly, and substantially more important: I really think this isn't the best way to go for species, which are just initialized with super-simple rules. It adds overhead to the symbolic processing (don't know how much, but definitely not few), by refactoring the whole stoichiometry, although this will never be necessary for this very specific type of conservation law. For hits reason, I really think it is better to keep simple things simple and not refactor fluxes and the stoichiometry, if we know that we don't need to...

FFroehlich: No, just use the `add_component(Expression(...))`, keep track of the symbolic identifiers for those expression and add them to a vector `r`. You can then calculate `xdot` as `S*r` instead of `S*w`, only `dxdotdw` needs additional rows for non-flux Expressions (all of them come from the conservation laws. Shouldn't be too much changes.

paulstapor: i.e.: `r` is then basically `w`, just with the additional expressions, one per CL?
`xdot = S * r` then?

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  18
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  18
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  18
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

paulstapor:
Okay, addressed your PR review and comment @FFroehlich 
Tell me, if you're fie with the implementation, then I will merge things.
Would be nice to bring things into master soon, to implement it into parPE then

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  18
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

FFroehlich:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- python/tests/test_compare_conservation_laws_sbml.py  18
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/ICB-DCM/AMICI/pullRequest?prid=5383001&bid=17649846)

python/amici/sbml_import.py
```diff
@@ -1269,7 +1269,7 @@ def replace_assignments(formula: str) -> sp.Basic:
         self.symbols['llhy']['name'] = l2s(llh_y_syms)
         self.symbols['llhy']['identifier'] = llh_y_syms
 
-    def _process_conservation_laws(self, ode_model) -> None:
+    def process_conservation_laws(self, ode_model) -> None:
```
paulstapor, line 1272: Maybe stupid, but: Just want to understand, why not starting with an underscore here...

python/amici/ode_export.py
```diff
@@ -833,17 +834,24 @@ def import_from_sbml_importer(self,
...
-        # setting these equations prevents native equation generation
-        self._eqs['dxdotdw'] = si.stoichiometric_matrix
-        self._eqs['w'] = si.flux_vector
-        self._syms['w'] = sp.Matrix(
-            [sp.Symbol(f'flux_r{idx}', real=True)
-             for idx in range(len(si.flux_vector))]
-        )
-        self._eqs['dxdotdx'] = sp.zeros(si.stoichiometric_matrix.shape[0])
-
+        # assemble fluxes and add them as expressions to the model
```
paulstapor, line 840: I added a comment here, that this generates a preliminary dxdt.
Seems like that was erased. Please leave it in. It not really understandable otherwise why we generate the dynamics twice (and overwrite what we just created). That caused me some headache to undertsand and it will cause headaches to the next person working on it, if not documented

python/amici/ode_export.py
```diff
@@ -833,17 +834,24 @@ def import_from_sbml_importer(self,
...
+        fluxes = []
+        for ir, flux in enumerate(si.flux_vector):
```
paulstapor, line 842: we don't necessarily need that, but could use the code lines from line 916 (`for symbol in [s for s in symbols if s != 'my']:`) and following (as I suggested). What's the purpose of changing that to this version?

python/amici/ode_export.py
```diff
@@ -897,19 +905,43 @@ def dx_dt(x_index, x_Sw):
...
+        self._eqs['dxdotdx'] = sp.zeros(nx_solver)
+
+        nw = len(self._expressions)
+        # append zero rows for conservation law `w`s, note that
+        # _process_conservation_laws is called after the fluxes are added as
+        # expressions, if this ordering needs to be changed, this will have
+        # to be adapted.
+        self._eqs['dxdotdw'] = si.stoichiometric_matrix.row_join(
+            sp.zeros(nx_solver, nw-nr)
+        )
```
paulstapor, line 940: We can either shorten the expression or increase dxdotdw.
I suggested to do the former. What's the advantage of doing the latter?

python/amici/sbml_import.py
FFroehlich: underscore mark private methods/members that are not part of the public API. Generally shouldn't be used outside the scope of the respective functions and may change without backwards-compatibility.

python/amici/ode_export.py
FFroehlich: But we do not generate a preliminary `dxdt` here? I am not sure what you think is generated twice.

python/amici/sbml_import.py
paulstapor: okay, and that's because it used across 2 files, you wouldn't underscore it?

python/amici/ode_export.py
paulstapor: We use the fluxes to create ` symbols['species']['dt']`.
1. If you're not familiar with the routine, you would expect that this should basically be teh symbolic expression for dxdot.
2. Based on ` symbols['species']['dt']`, model quantities are created. Based on those quantities, it's decied, whether a state is marked as constant or not, which later on influences CL computation. In this sense, we indeed create information on xdot already here. If you're not really, really familiar with ode_export.py, there is no way that would understand how these code parts are intertwined...

paulstapor: Maybe to make it more precise: Based on what is created here, ode_model._states[*index of state*]['dt'] is written.
However, ode_model._sym['x'] is not created from it. Before, those two qunatities (ode_model._states[*index of state*]['dt'] and  ode_model._sym['x']) were in line, now they aren't any more (and can't be due to CLs).
That's really not intuitive and took me quite a while. We need to document that somehow, e.g., by a comment here.

python/amici/ode_export.py
FFroehlich: The idea here is to directly loop over `si.flux_vector` and only do that once. Directly looping over the vector entries is generally more pythonic, easier to mantain and less error-prone. Keeping everything in the same place is in the same spirit. Making changes is way easier as, you don't have to find all the different loops that may have drifted apart due to additional code changes (also less likely to miss one of them).

Initialization of expressions could indeed be done via the symbols variable, no strong preference on that.

python/amici/ode_export.py
FFroehlich: It doesn't make sense to add the conservation laws as expressions and then remove them from the expressions again. For this very specific setup of the sbml_import it currently doesn't matter (because those species are not used outside the flux vector, where they are currently flattened out).

Adding stuff and removing it again is generally very bad practice as it increases the total amount of code and number of functions/files you have to look at in order to figure out what is actually happening. If you don't want to add them to the expressions, implement that in add_conservation_laws.

paulstapor: This is why I was considering to remove the adding of CLs to the expressions. I just didn't do this because I was sure it was there for good purpose and probably needed for pySB. The SBML-Importer doesn't need it, and I also don't see, how it would need it, if real CLs were implemented...

From that perspective, I was left with: either (1) removing them right now directly or (2) adding zeros to dxdotdw and remove them by multiplying them with zeros.
(2) doesn't seem more elegant/any better practice to me than (1), as it also adds stuff just to remove (indirectly)later and moreover increases the size of variables used in C++(such as w) by entries, which are never used and hence adds (even if only minimal) computational overhead.

FFroehlich: In other words, what is done with `self._eqs['dxdotdx']` and `self._eqs['dxdotdw']` is absolutely not how things are supposed to be done. But it's one way to do it as efficiently as it is currently done. 

If there already is a function that does something similar to what you want, change that function and don't try to wiggle your away around outside of it.

python/amici/ode_export.py
FFroehlich: the values in `symbols['species']['dt']` use the public API of `State`. There is nothing that is created twice. Maybe its confusing to have `import_from_sbml_importer` as member of `SbmlImporter` as thats how it is intended to operate wrt `ODE_model`.

Don't forget that `ode_model._states` is a dict, i.e. dont assume anything about order. `import_from_sbml_importer` is super hacky and makes a shitton of assumptions, such as the ordering and number of elements in `ode_model._sym['x']` and `ode_model._states[index of state]`. If you want the ordering of two things to be the same order, put them in a list of dicts/tuples/whatever. If they are in two seperate lists, assume thats intential that they may not have the the same order. just because they the same order now, doesn't mean they will have it in the future. This may not always be the case everywhere as some amici code is pretty old by now, but thats how things are supposed to be (not just in amici, but generally in python code)

In this part of the code, you shouldn't even think or crae about `ode_model._states`. It doesn't appear anywhere in this function, and there shouldn't be any mention/documentation about it. 

In terms of the circular dependency that you are referring to, the important thing is that `process_conservation_laws(self)` requires that species are already added to the model, so it needs to be done after those are added to the model.

python/amici/sbml_import.py
FFroehlich: it's not so much about the files. Its called from an ODE_model method, which is by definition not a `SbmlImporter` method.

python/amici/ode_export.py
FFroehlich: > This is why I was considering to remove the adding of CLs to the expressions. I just didn't do this because I was sure it was there for good purpose and probably needed for pySB. The SBML-Importer doesn't need it, and I also don't see, how it would need it, if real CLs were implemented...
> 
> From that perspective, I was left with: either (1) removing them right now directly or (2) adding zeros to dxdotdw and remove them by multiplying them with zeros.
> (2) doesn't seem more elegant/any better practice to me than (1), as it also adds stuff just to remove (indirectly)later and moreover increases the size of variables used in C++(such as w) by entries, which are never used and hence adds (even if only minimal) computational overhead.

The fact that all fluxes are captured in `w` and that state variables **only** occur in `w` is absolutely unique to the python-sbml importer (and bound not to be true at some point as we add features to the sbml import). For matlab models as well as pysb models states can and do occur outside of `w`, which means that having them defined as expression is necessary (although conservation laws don't exist in matlab models).

Its fine to change this implementation, but you have to do it in the right place. So where are those Expressions added (this is the place you will want to edit)? `ODE_model.add_conservation_law()`. Turns out that is a general function that is implemented independent of SBML/PySB import. If you change something there, this also changes stuff for the pysb import. This is intended like that. This function is not taylored to the pysb import, but to what we can generally assume about models defined in ODE_model. Changing that function will mess up a lot of stuff you don't want to deal with, so its way easier to just add the couple of entries to `w` and live with the minimal overhead, having general purpose implementation means that we have to do that from time to time.

PR #1049, opened 2020-04-23
Bugfix in processing of SBML rules

lcontento:
I found a couple of bugs in the processing of SBML assignment rules.
The first is related to the missing translation from the Pythonic `**` exponentiation to the `^` operator supported by `libsbml`.
The second regards some code that looks like it is doing nothing. I changed to what I think it should be the correct version, but please check it.


python/amici/sbml_import.py
```diff
@@ -1029,11 +1029,12 @@ def _process_rules(self) -> None:
...
                     nested_formula = \
-                        nested_formula.subs(variable, formula)
-                    nested_rule.setFormula(str(nested_formula))
+                        str(nested_formula.subs(variable, formula)).replace('**', '^')
+                    if nested_rule.setFormula(nested_formula) != sbml.LIBSBML_OPERATION_SUCCESS:
+                        raise SBMLException(f'Formula {nested_formula} cannot be parsed by libSBML!')
 
-                for variable in assignments:
-                    assignments[variable].subs(variable, formula)
+                for variable2 in assignments:
```
FFroehlich, line 1036:
```suggestion
                for assignment in assignments:
```

python/amici/sbml_import.py
```diff
@@ -1029,11 +1029,12 @@ def _process_rules(self) -> None:
...
+                    assignments[variable2] = assignments[variable2].subs(variable, formula)
```
FFroehlich, line 1037:
```suggestion
                    assignments[assignment] = assignments[assignment].subs(variable, formula)
```

FFroehlich:
How did you find this? Did it occur in the testsuite?

lcontento:
No, it was with a SBML model generated by me. Whether the code generation failed actually depended on the order of the rules inside the xml file.

PR #1050, opened 2020-04-23
account for compartment volumes in dxdotdw

paulstapor:
Little fix, as currently, dxdotdw was missing compartment volumes.

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -1269,29 +1269,42 @@ class SbmlImporter:
         self.symbols['llhy']['name'] = l2s(llh_y_syms)
         self.symbols['llhy']['identifier'] = llh_y_syms
 
-    def process_conservation_laws(self, ode_model) -> None:
+    def process_conservation_laws(self, ode_model, volume_updates) -> List:
         """
         Find conservation laws in reactions and species.
 
         :param ode_model:
             ODEModel object with basic definitions
 
+        :param volume_updates:
+            List with updates for the stoichiometrix matrix accounting for
+            compartment volumes
+
+        :returns volume_updates_solver:
+            List (according to reduced stoichiometry) with updates for the
+            stoichiometrix matrix accounting for compartment volumes
         """
         conservation_laws = []
 
         # So far, only conservation laws for constant species are supported
-        self._add_conservation_for_constant_species(ode_model,
-                                                    conservation_laws)
+        species_solver = self._add_conservation_for_constant_species(
+            ode_model, conservation_laws)
+
+        # prune out species from stoichiometry and
+        volume_updates_solver = self._reduce_stoichiometry(species_solver,
+                                                           volume_updates)
 
         # add the found CLs to the ode_model
         for cl in conservation_laws:
             ode_model.add_conservation_law(**cl)
 
+        return volume_updates_solver
+
     def _add_conservation_for_constant_species(
             self,
             ode_model: ODEModel,
             conservation_laws: List[ConservationLaw]
-    ) -> None:
+    ) -> List[int]:
         """
         Adds constant species to conservations laws
 
@@ -1300,6 +1313,9 @@ class SbmlImporter:
 
         :param conservation_laws:
             List of already known conservation laws
+
+        :returns species_solver:
+            List of species indices which remain later in the ODE solver
         """
 
         # decide which species to keep in stoichiometry
@@ -1321,9 +1337,39 @@ class SbmlImporter:
                 # mark species to delete from stoichiometrix matrix
                 species_solver.pop(ix)
 
+        return species_solver
+
+    def _reduce_stoichiometry(self, species_solver, volume_updates) -> List:
+        """
+        Reduces the stoichiometry with respect to conserved quantities
+
+        :param species_solver:
+            List of species indices which remain later in the ODE solver
+
+        :param volume_updates:
+            List with updates for the stoichiometrix matrix accounting for
+            compartment volumes
+
+        :returns volume_updates_solver:
+            List (according to reduced stoichiometry) with updates for the
+            stoichiometrix matrix accounting for compartment volumes
+        """
+
         # prune out constant species from stoichiometric matrix
         self.stoichiometric_matrix = \
             self.stoichiometric_matrix[species_solver, :]
+
+        # updates of stoichiometry (later dxdotdw in ode_exporter) must be
+        # corrected for conserved quantities:
+        volume_updates_solver = []
+        for update in volume_updates:
+            x_index = update[0]
+            if x_index in species_solver:
+                x_index = species_solver.index(x_index)
+                volume_updates_solver.append((x_index, update[1], update[2]))
+
+        return volume_updates_solver
+
 
     def _replace_in_all_expressions(self,
                                     old: sp.Symbol,
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -899,10 +899,16 @@ class ODEModel:
                 v = si.compartment_assignment_rules[v_name]
                 dv_dt = v.diff(si.amici_time_symbol)
                 dv_dx = v.diff(x_id)
-                return (x_Sw - dv_dt*x_id)/(dv_dx*x_id + v)
+                xdot = (x_Sw - dv_dt*x_id)/(dv_dx*x_id + v)
+                for w_index, flux in enumerate(fluxes):
+                    dxdotdw_updates.append((x_index, w_index, xdot.diff(flux)))
+                return xdot
             else:
                 v = si.compartment_volume[list(si.compartment_symbols).index(
                     si.species_compartment[x_index])]
+                for w_index, flux in enumerate(fluxes):
+                    dxdotdw_updates.append((x_index, w_index,
+                        si.stoichiometric_matrix[x_index, w_index] / v))
                 return x_Sw/v
 
         # create dynmics without respecting conservation laws first
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -891,7 +891,6 @@ class ODEModel:
             if v_name in si.compartment_rate_rules:
                 dv_dt = si.compartment_rate_rules[v_name]
                 xdot = (x_Sw - dv_dt*x_id)/v_name
-                # might need to do this for assignment rules as well
                 for w_index, flux in enumerate(fluxes):
                     dxdotdw_updates.append((x_index, w_index, xdot.diff(flux)))
                 return xdot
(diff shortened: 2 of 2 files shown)


paulstapor:
Anything to consider here, or am I free to merge?

dilpath:
So just to clarify, I mean if an assignment rule or rate rule contains an expression equivalent to `flux_r0`, e.g. `kS^2`, should `kS^2` be swapped for `flux_r0` before differentiating with respect to `flux_r0`?

Also, species with rate rules might not have correct `dxdotdw` entries either, as they are returned immediately, in the first few lines of `dx_dt` (similarly, compartments with rate rules).

https://github.com/ICB-DCM/AMICI/blob/develop/python/amici/ode_export.py#L876

paulstapor:
> So just to clarify, I mean if an assignment rule or rate rule contains an expression equivalent to `flux_r0`, e.g. `kS^2`, should `kS^2` be swapped for `flux_r0` before differentiating with respect to `flux_r0`?
> 
> Also, species with rate rules might not have correct `dxdotdw` entries either, as they are returned immediately, in the first few lines of `dx_dt` (similarly, compartments with rate rules).
> 
> https://github.com/ICB-DCM/AMICI/blob/develop/python/amici/ode_export.py#L876

If the dxdt term is computed correctly in there atm but has no fluxes, this should still be fine, I think...
I am just worried about the Jacobian in those cases...

FFroehlich:
> > So just to clarify, I mean if an assignment rule or rate rule contains an expression equivalent to `flux_r0`, e.g. `kS^2`, should `kS^2` be swapped for `flux_r0` before differentiating with respect to `flux_r0`?
> > Also, species with rate rules might not have correct `dxdotdw` entries either, as they are returned immediately, in the first few lines of `dx_dt` (similarly, compartments with rate rules).
> > https://github.com/ICB-DCM/AMICI/blob/develop/python/amici/ode_export.py#L876
> 
> If the dxdt term is computed correctly in there atm but has no fluxes, this should still be fine, I think...
> I am just worried about the Jacobian in those cases...

FD implementations and checks for the jacobian and others are pretty high up in my priority list atm :) https://github.com/ICB-DCM/AMICI/milestone/9

paulstapor:
> FD implementations and checks for the jacobian and others are pretty high up in my priority list atm :) https://github.com/ICB-DCM/AMICI/milestone/9

Ok, that's very nice...
I would currently guess, that for rate rules, we have explicit x-dependencies in xdot, hence need to compute `dxdotdx` via symbolic differentiation.
For assignment rules to species, I really don't know, how it's handeled and whether they are actually treated as states or rather as observables?

paulstapor:
okay, let's make it like that:
I'll update parPE just to this feature branch of Amici for the moment and not to develop or master. This is maybe not the nicest practice, but helps us with two things:
 - This PR will not be time-critical any more, and we have time to really think about how to best deal with stuff.
 - People who need this fix in parPE right now can still work with it.

FFroehlich:
> I would currently guess, that for rate rules, we have explicit x-dependencies in xdot, hence need to compute `dxdotdx` via symbolic differentiation.

Lets just remove the manual assignment of dxdotdx and check whether compute time for the large scale model blows up?

paulstapor:
> > I would currently guess, that for rate rules, we have explicit x-dependencies in xdot, hence need to compute `dxdotdx` via symbolic differentiation.
> 
> Lets just remove the manual assignment of dxdotdx and check whether compute time for the large scale model blows up?

we can do that, yep... For CanPathPro models, this should actually anyhow just give zeros...

FFroehlich:
> okay, let's make it like that:
> I'll update parPE just to this feature branch of Amici for the moment and not to develop or master. This is maybe not the nicest practice, but helps us with two things:
> 
> * This PR will not be time-critical any more, and we have time to really think about how to best deal with stuff.
> * People who need this fix in parPE right now can still work with it.

Shouldn't this branch be less wrong as the current develop branch? From my side it's fine to merge this and fix remaining errors in a different PR.

paulstapor:
It is already less wrong, you're right. 😁 
It might however still be incorrect to some degree...

paulstapor:
I'm also fine with merging

paulstapor:
Nice that we have a unit test on conservation laws... 👍 
That showed that also the dxdotdw_updates need to be changed according to conservation laws now. I implemented that.
Had to restore the function reduce_stoichiometry, however, as stoichiometry reduction goes beyond a one-line-thing now and should hence get a function of its own...

Concerning dxdotdx:
How about doing this line-by-line in dx_dt, only for those species, which actually have rate rules? I don't know, how much symbolic differentiation will affect computation time, but if we don't need to do it in 99% of the cases (especially in large models), I feel like we should try to avoid it, if its not much overhead and can be done without completely violating the intended code structure...

I feel like a line-by-line thing can be implemented in pretty much analogously to dxdotdw with 3-4 additional lines maybe. Hence, it would not violate the idea of ode_export more than this is done already...

Shall I make a quick suggestion?
Opinions?

python/amici/sbml_import.py
```diff
@@ -1321,10 +1337,40 @@ def _add_conservation_for_constant_species(
...
         self.stoichiometric_matrix = \
             self.stoichiometric_matrix[species_solver, :]
 
+        # updates of stoichiometry (later dxdotdw in ode_exporter) must be
+        # corrected for conserved quantities:
+        volume_updates_solver = []
+        for update in volume_updates:
+            x_index = update[0]
+            if x_index in species_solver:
+                x_index = species_solver.index(x_index)
```
FFroehlich, line 1368: overwriting the irator in a loop is a recipe for confusion.

paulstapor: I had that changed at some point. You're definitely right!
Will change that...
How/when best? We will have another PR very soon anyhow...

FFroehlich: especially since thats also the iterator of the outer loop, maybe I'm missing some wild recursion here, but I don't think this function is doing what its supposed to do.

python/amici/sbml_import.py
```diff
@@ -1321,10 +1337,40 @@ def _add_conservation_for_constant_species(
...
+        """
+
         # prune out constant species from stoichiometric matrix
         self.stoichiometric_matrix = \
             self.stoichiometric_matrix[species_solver, :]
 
+        # updates of stoichiometry (later dxdotdw in ode_exporter) must be
+        # corrected for conserved quantities:
+        volume_updates_solver = []
+        for update in volume_updates:
```
FFroehlich, line 1365: x_index, var1, var2, makes it explicit what we loop over and more evident that we have two loops with the same iterator

python/amici/sbml_import.py
paulstapor: Ah, right. No, that's actually fine. `update` is immutable and a tuple...
Hence, line 1366 creates a new obejct and doesn't overwrite ;)

python/amici/sbml_import.py
paulstapor: had that one at some point, needed to enumerate over update (hence changed to update) at some point and left it then...
Agree with you!

python/amici/sbml_import.py
paulstapor: This makes it safe, but still bad practice. Will change that.

FFroehlich: the logic in this function is so convoluted, I can't really say what its supposed to do. According to the docstring, x_index is already a list of indices, so I don't understand why you have to index over that again.

From a design perspective, its probably more robust to generally work with identifiers here instead of indices as we can reconstruct the index ad hoc when the update is applied.

FFroehlich: > Ah, right. No, that's actually fine. `update` is immutable and a tuple...
> Hence, line 1366 creates a new obejct and doesn't overwrite ;)

Yeah, overwriting the original value would be even worse, but currently the value of update[0] goes completely unused. If that is intended, then assign update[0] to `_` in the first place. Like this it looks like something is fishy in the implementation.

paulstapor: Need better doc then...
No, x_index is an int. Actually, this thing is supposed to be rather simple. It's one loop and one if, in which the first index is "corrected"...

FFroehlich: So you are trying to do this:
```
for _, x, y in volume_updates:
    volume_updates_solver.extend([(ix, x, y) for ix in range(len(species_solver))])
```
?

Its not prohibited to iterate over indices in python, I would actually prefer that over iterating over elements and then looking up the indices.

paulstapor: ah, wait, now I see more doc is needed...
This indexing is not just for looking up! It provides the mapping between x_rdata and x_solver.

paulstapor: > So you are trying to do this:
> 
> ```
> for _, x, y in volume_updates:
>     volume_updates_solver.extend([(ix, x, y) for ix in range(len(species_solver))])
> ```
> 
> ?
> 
> Its not prohibited to iterate over indices in python, I would actually prefer that over iterating over elements and then looking up the indices.

Yes, this would work, but assume that updates are always passed in the right order...
If there is no way whatsoever that the state ordering is changed by conservation laws, then it's fine...

FFroehlich: happy to be proven wrong, but I am pretty sure that this function just reimplements what is currently being done. I agree that using indices is not the best way to go about this.

paulstapor: you mean the whole reduce_stoichiometry function?
If there is already a function with an API for this, it wasn't used in SBML import so far. SBML import was lacking this reduction until now. But there may be some functionality I am not aware of...

FFroehlich:
```
for update in volume_updates:
    x_index = update[0]
    if x_index in species_solver:
        x_index = species_solver.index(x_index)
        volume_updates_solver.append((x_index, update[1], update[2]))
```
is equivalent to
```
for _, x, y in volume_updates:
    volume_updates_solver.extend([(ix, x, y) for ix in range(len(species_solver))])
```

intermediate steps:
```
for x_index, x, y in volume_updates:
    if x_index in species_solver:
        x_index = species_solver.index(x_index)
        volume_updates_solver.append((x_index, x, y))
```

```
for _, x, y in volume_updates:
    if x_index in species_solver:
        x_index = species_solver.index(x_index)
        volume_updates_solver.append((x_index, x, y))
```

```
for _, x, y in volume_updates:
    if x_index in range(len(species_solver)):
        volume_updates_solver.append((x_index, x, y))
```

not a big fan of having hour long conversations about four lines of code ...

Issue #1054, opened 2020-04-24
SBMLTestSuite failing


FFroehlich:
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[9]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[13]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[30]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[29]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[118]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[213]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[579]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[1245]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[1422]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[1423]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[1431]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[1638]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[1341]
FAILED tests/testSBMLSuite.py::test_sbml_testsuite_case[1342]

Looks like there are currently 14 failing testcases. Issue was introduced somewhere in #1036, #1047 or #1050. Currently trying to figure out where exactly. In case you are intending to use this code anywhere near production, which I believe is the case here, I would recommend to get this fixed ASAP.

FFroehlich:
Also, if you are touching the SBML import in any way, run the testsuite before merging anything. The testuite runs for close to 1.5h which I would like to avoid having as mandatory requirement for all PRs.

dweindl:
> Also, if you are touching the SBML import in any way, run the testsuite before merging anything. The testuite runs for close to 1.5h which I would like to avoid having as mandatory requirement for all PRs.

Fully agreed. Also, reviewers should remind contributors of that.

If somebody feels bored: Why not check the git diff during each CI run, and run the sbml test suite as soon as sbml import code was modified...

dilpath:
Hm I don't think I changed `sbml_import.py` after my last run of the test suite. I'm on `develop`, commit 0b4f025 (the merging of #1036), so I'll double check and run the test suite now.

paulstapor:
[...]

PR #1057, opened 2020-04-25
CI: Run SBML test suite on pull request only if AMICI SBML import rou…

dweindl:
…tines are changed (#1054)


dweindl:
@FFroehlich : Should now trigger sbml test suite everytime SBML import gets modified. However, not automatically on PR to master (not supported). Merge/push to develop will still trigger always, so that shouldn't be too much of a problem.

paulstapor:
Nice, thanks!

.github/workflows/sbml-semantic-test-suite.yml
```diff
@@ -5,8 +5,9 @@ on:
       - develop
       - master
   pull_request:
-    branches:
-      - master
+    paths:
+      - python/amici/sbml_import.py
+      - python/amici/ode_export.py
```
FFroehlich, line 10: add `scripts/run-SBMLTestsuite.sh` and `tests/testSBMLSuite.py`?

dweindl: Sounds reasonable

PR #1060, opened 2020-04-27
Fix sbml import

paulstapor:
Sbml import should be fixed by now.
Moreover, import time should be mostly back to normal, for large-scale models with constant species.

The trick for the SBML-Import: Models without species are not supported. If all species are conserved, no species would be left after applying CLs. Hence, in this case, CLs are neglected. I think this is reasonable, as models without dynamics are not supposed to be our main problem, even if we neglect conservation laws...

The trick for high import times:
The symbols for the sensitivities of conservation laws are not crated for constant species, as there, the sensitivities will always be zero. The corresponding equations are set to zero. Thus, the ode_exporter skips a (potentially huge) matrix-matrix multiplication, which blew up the computation time.


paulstapor:
SBML test suite was passing, import time for the large-scale test model was fine again (slightly higher than before, due to CLs, but not much).
I however changed slight things in ode_exporter (standard loops to list comprehensions), which should do exactly the same. However, it makes definitely sense to wait for all checks...

paulstapor:
I'll do my best to get the pySB tests running again. It seems like the change in the symbols of dtcldp messed up things...
That's actually the only part which could have impacted pySB tests

python/amici/ode_export.py
```diff
@@ -1494,8 +1496,20 @@ def _compute_equation(self, name: str) -> None:
...
-            self._eqs[name] = self.sym(name)
+            # check, whether the CL consists of only one state. Then,
+            # sensitivities drop out, syms can be deleted
+            state_set = {state.get_id() for state in self._states}
+            self._eqs[name] = sp.Matrix([
+                [0] * self.np()
+                if len(state_set.intersection(
+                    tcl.get_val().free_symbols)) == 1
+                else [
+                    sp.Symbol(f's{strip_pysb(tcl.get_id())}__'
```
FFroehlich, line 1507: this should ideally reuse the symbols from the syms matrix, prone to diverge like that. Alternatively implement a function that generates symbol names and use it in both places.

python/amici/ode_export.py
```diff
@@ -1258,14 +1258,16 @@ def _generate_symbol(self, name: str) -> None:
             ])
             return
         elif name == 'dtcldp':
+            # check, whether the CL consists of only one state. Then,
+            # sensitivities drop out, syms can be deleted
+            state_set = {state.get_id() for state in self._states}
```
FFroehlich, line 1263: shouldn't this be the same as `set(self.syms('x'))`?

python/amici/ode_export.py
```diff
@@ -1258,14 +1258,16 @@ def _generate_symbol(self, name: str) -> None:
...
-                    sp.Symbol(f's{strip_pysb(tcl.get_id())}__'
-                              f'{strip_pysb(par.get_id())}',
-                              real=True)
+                [sp.Symbol(f's{strip_pysb(tcl.get_id())}__'
+                           f'{strip_pysb(par.get_id())}', real=True)
                     for par in self._parameters
-                ]
-                for tcl in self._conservationlaws
+                 ]
+                for tcl in self._conservationlaws if len(
```
FFroehlich, line 1269: implement check as function, e.g. `conservation_has_multispecies`? Inherently adds documentation through the function name and simplifies maintenance by reducing duplication.

paulstapor:
Okay, I think I know now, how to solve this. However, there are two solutions. None of those is perfect. I would like to propose them and discuss, if either one of those should be chosen or whether we should go for something completely different.

The problem (coarse grained):
In the routine `_total_derivative`, the product `dwdtcl * dtcldp` is computed. For large-scale models like in CanPathPro, this takes many hours. Important to note: `dtcldp` is zero, so this could be circumvented.

The problem (fine grained):
The routine `_total_derivative` checks, whether one of the components is zero. However, for `dtcldp`, the matrix to multiply will always be the matrix with the symbols, not with the equations. The symbols however are never zero even if the equations are, since the symbols identifiers are identifiers, which are later written in the header file of `dtcldp.h`.

What doesn't solve the problem:
1. Writing symbolically `0` in the symbol matrices fails, as otherwise, the ode_exporter creates header files including stuff like `#define Matrix[[0]] dtcldp[0]`.
No need to tell, that the compiler will crash then.
2. Deleting the symbols from the matrix, for which the equations are zero doesn't work either: If some real conservation laws are left, the matrix `_syms['dtcldp']` won't be empty, the check for it being zero will (correctly) fail. However, now `_syms['dtcldp']` has the wrong dimension, if some 1-species-CLs were in the model. Again, this leads to crashing.

Solutions I can think of:
1. In `_total_derivative`, we may want to check not only for the symbols being 0, but maybe also for the equations here:
https://github.com/ICB-DCM/AMICI/blob/1269134a62ea2b215ceb15fe2fd7d9b3de601c19/python/amici/ode_export.py#L1661
This is not ideal for several reasons, but definitely resolves the problem.

2. We might want to write symbolic 0's in the symbols of `dtcldp`. However, we then need to check whether we would redefine `0` with a compiler command when writing header files and skip this line to be written here:
https://github.com/ICB-DCM/AMICI/blob/1269134a62ea2b215ceb15fe2fd7d9b3de601c19/python/amici/ode_export.py#L2166

As said, both involves somewhat hand-made fixes in routines, which usually shouldn't be touched. I don't really see any other good solution, but I'm happy for any suggestion.
My favorite would be solution 2, as preprocessing text when writing header files seems still somewhat natural to me, rather than fiddling around in derivative routines.
But I'm open to suggestions, of course...

FFroehlich:
> Okay, I think I know now, how to solve this. However, there are two solutions. None of those is perfect. I would like to propose them and discuss, if either one of those should be chosen or whether we should go for something completely different.
> 
> The problem (coarse grained):
> In the routine `_total_derivative`, the product `dwdtcl * dtcldp` is computed. For large-scale models like in CanPathPro, this takes many hours. Important to note: `dtcldp` is zero, so this could be circumvented.
> 
> The problem (fine grained):
> The routine `_total_derivative` checks, whether one of the components is zero. However, for `dtcldp`, the matrix to multiply will always be the matrix with the symbols, not with the equations. The symbols however are never zero even if the equations are, since the symbols are identifiers, which are later written in the header file of `dtcldp.h`.
> 
> What doesn't solve the problem:
> 
> 1. Writing symbolically `0` in the symbol matrices fails, as otherwise, the ode_exporter creates header files including stuff like `#define Matrix[[0]] dtcldp[0]`.
>    No need to tell, that the compiler will crash then.
> 2. Deleting the symbols from the matrix, for which the equations are zero doesn't work either: If some real conservation laws are left, the matrix `_syms['dtcldp']` won't be empty, the check for it being zero will (correctly) fail. However, now `_syms['dtcldp']` has the wrong dimension, if some 1-species-CLs were in the model. Again, this leads to crashing.
> 
> Solutions I can think of:
> 
> 1. In `_total_derivative`, we may want to check not only for the symbols being 0, but maybe also for the equations here:
>    https://github.com/ICB-DCM/AMICI/blob/1269134a62ea2b215ceb15fe2fd7d9b3de601c19/python/amici/ode_export.py#L1661
>    
>    This is not ideal for several reasons, but definitely resolves the problem.
> 2. We might want to write symbolic 0's in the symbols of `dtcldp`. However, we then need to check whether we would redefine `0` with a compiler command when writing header files and skip this line to be written here:
>    https://github.com/ICB-DCM/AMICI/blob/1269134a62ea2b215ceb15fe2fd7d9b3de601c19/python/amici/ode_export.py#L2166
> 
> As said, both involves somewhat hand-made fixes in routines, which usually shouldn't be touched. I don't really see any other good solution, but I'm happy for any suggestion.
> My favorite would be solution 2, as preprocessing text when writing header files seems still somewhat natural to me, rather than fiddling around in derivative routines.
> But I'm open to suggestions, of course...

Ah sorry, didn't think of that. How about we just modify the syms to be zero where appropriate, keep the `eq` assignment to syms as before and change `_write_index_files` to check whether the written symbol is zero (or a zero matrix) (to avoid issues with the gerenated header files).

paulstapor:
> Ah sorry, didn't think of that. How about we just modify the syms to be zero where appropriate, keep the `eq` assignment to syms as before and change `_write_index_files` to check whether the written symbol is zero (or a zero matrix) (to avoid issues with the gerenated header files).

this is pretty exactly my second suggestion, which I was also in favor of. Good to know, that it's not only me thinking that. ;)

FFroehlich:
> > Ah sorry, didn't think of that. How about we just modify the syms to be zero where appropriate, keep the `eq` assignment to syms as before and change `_write_index_files` to check whether the written symbol is zero (or a zero matrix) (to avoid issues with the gerenated header files).
> 
> this is pretty exactly my second suggestion, which I was also in favor of. Good to know, that it's not only me thinking that. ;)

Yeah, also just realized that 👍

python/amici/ode_export.py
paulstapor: Maybe... will check. If so, your suggestions doesn't access private members directly and is hence nicer!

python/amici/ode_export.py
paulstapor: As discussed below, will do.

python/amici/ode_export.py
paulstapor: Maybe easier to read than a 6-line list comprehension indeed...

paulstapor:
All done.
I run the long taking tests locally as well, but now (finally), things should be fine...

All these PRs and the time spent on this CL part in SBML (more than one week) tell me, that CLs are something, which one really, really does *not* want to touch...

FFroehlich:
> All done.
> I run the long taking tests locally as well, but now (finally), things should be fine...
> 
> All these PRs and the time spent on this CL part in SBML (more than one week) tell me, that CLs are something, which one really, really does _not_ want to touch...

Just feel free to push to the PR. Generally speeds up things as CI runs for a long time and we can already do reviews.

paulstapor:
Indeed, I forgot to push... 🤦‍♂️ 
However, the human cancer model can now be wrapped in < 1h again. We will have to adapt the import time for the performance tests slightly, however, as yet some additional computations are involved now...

python/amici/ode_export.py
```diff
@@ -2168,6 +2166,8 @@ def _write_index_files(self, name: str) -> None:
 
         for index, symbol in enumerate(symbols):
             symbol_name = strip_pysb(symbol)
+            if str(symbol) == '0':
```
FFroehlich, line 2170: Why the string based comparison here? How sure are you this is always going to be 0 and non 0.0. Whats wrong with `symbol.is_zero`?

paulstapor: Had the symbolic comparison first.
Turns out: Not all qunatities actually pass symbols, some (e.g., dwdx) pass strings! Then, symbolic comparison fails... Therefore the string.
Pretty sure, that `0` will do, as in all cases, which I might think of, the symbolic zeros are initialized by sympy.zeros(). This really creates a '0', whereas '0.0' may only happen if floats are involved. Anyhow, the only routines needing this currently is dtcldp.

However, I fully agree that these should *always* be symbols here. I'm happy to clean this up in the next PR. For the moment, these improvments really have to go into develop now...

FFroehlich: Yeah thats fine, `syms` should always yield symbols. fine to not address now, but please open an issue.

paulstapor: Done, see #1061 .

paulstapor:
Free to merge?

FFroehlich:
> Free to merge?

Approved and all review comments addressed – yes!

Issue #1067, opened 2020-04-29
simply returns invalid piecewise function


jvanhoefer:
https://github.com/ICB-DCM/AMICI/blob/7d632b52975b9ef5af2b389d9eaeb38fbd40f1cb/python/amici/sbml_import.py#L1028

For the input `piecewise(k_light, time < 60, k_nolight)` the call of `sp.sympify` returns `Piecewise((k_light, time < 60), (k_nolight, True))`, which is no longer in the format of a piecewise function [expected by libsbm](http://sbml.org/Special/Software/libSBML/docs/formatted/python-api/libsbml-math.html)l and hence can not be parsed in the following.

FFroehlich:
should be fixable by using `sympy.printing.mathml` instead of str and then piping through `libsbml.writeMathMLToString(new_mathml)` (or `libsbml.readMathMLFromString` and `libsbml.formulaToL3String` if necessary). Should also make replacing "**" by "^" obsolete.

PR #1069, opened 2020-04-30
fix sympy rules, fixes issue #1067

jvanhoefer:
as suggested, goes directly via MathML


python/amici/sbml_import.py
```diff
@@ -1025,14 +1025,13 @@ def _process_rules(self) -> None:
...
                 for nested_rule in rules:
-                    nested_formula = sp.sympify(
-                        sbml.formulaToL3String(nested_rule.getMath()),
-                        locals=self.local_symbols)
-                    nested_formula = \
-                        str(nested_formula.subs(variable, formula)).replace('**', '^')
-                    if nested_rule.setFormula(nested_formula) != sbml.LIBSBML_OPERATION_SUCCESS:
-                        raise SBMLException(f'Formula {nested_formula} cannot be parsed by libSBML!')
+                    nested_rule_math_ml = sp.printing.print_mathml(
+                        sbml.formulaToL3String(nested_rule.getMath()))
```
paulstapor, line 1029: Sorry, I'm not sure what `sp.printing.print_mathml` *exactly* does. However, I think the following things are to be considered. If those are fulfilled, things should be fine:
 * replacement of `**` with `^`
 * accounting for local parameters
And I would definitely check the SBML test suite :)
Additionally, if AMICI now can parse additional SBML models, I would account for that in the test-suite, such that we won't support for those any more. :)

FFroehlich: `print_mathml` generates a mathml string that can be correctly parsed by libsbml, s we don't have to perform any acrobatics here to make libsbml understand what we are outputting here.

python/amici/sbml_import.py
```diff
@@ -1025,14 +1025,13 @@ def _process_rules(self) -> None:
...
             if variable in rulevars:
                 for nested_rule in rules:
-                    nested_formula = sp.sympify(
-                        sbml.formulaToL3String(nested_rule.getMath()),
-                        locals=self.local_symbols)
-                    nested_formula = \
-                        str(nested_formula.subs(variable, formula)).replace('**', '^')
-                    if nested_rule.setFormula(nested_formula) != sbml.LIBSBML_OPERATION_SUCCESS:
-                        raise SBMLException(f'Formula {nested_formula} cannot be parsed by libSBML!')
+                    nested_rule_math_ml = sp.printing.print_mathml(
```
FFroehlich, line 1028: we lost the `nested_formula.subs(variable, formula)` here, which is the key thing this part of the code is supposed to do (flatten out nested rules).

FFroehlich:
There are 3 sbml tests currently failing, please make sure they pass before merging!

jvanhoefer:
I added a test, if `sp.sympify(...).subs(variable, formula)` returns a `sp.Function` object, since this can not be parsed to MathML. This seems to resolve the three SBML tests. @FFroehlich are you fine with this?

FFroehlich:
> I added a test, if `sp.sympify(...).subs(variable, formula)` returns a `sp.Function` object, since this can not be parsed to MathML. This seems to resolve the three SBML tests. @FFroehlich are you fine with this?

Generally okay, as I don't think we want to support every last bit of sbml functionality. However:

- I am not quite sure what this means in terms of #1067. I would assume that this makes it impossible to have piecewise functions in nested rules?
- I think that it's equally problematic if the `sp.Function` is somewhere contained as operator in a `sp.Basic` variable which is currently not covered by the exception. Thus exception handling should probably be implemented more generic.

jvanhoefer:
So, now we substitute stuff with sympy again... Sometimes `sbml.readMathMLFromString(nested_rule_math_ml)` does produce `None` as result, since sympy can not handle the formula of the rule. In this case I throw an `SBMLException`...

If you are fine with this solution now (+SBMLtests pass), I would merge :)

FFroehlich:
Looks good, fine to merge!

jvanhoefer:
We (meaning @LarsFroehling and me ) were actually still looking into this, since it did not work properly on Lars example. It seems, like Lars found the reason for this, but wanted to push the adaptions today/tomorrow

jvanhoefer:
sorry for not making this clear here :/

LarsFroehling:
as @jvanhoefer said I am a bit late unfortunately. 
I think the MathML string needs to be wrapped with a header before it can be read by `readMathMLFromString`

Where would be the best place to commit this?

FFroehlich:
Sorry about that, just open a new PR for this branch.

PR #1070, opened 2020-05-02
Cleanup Sonarcloud issues

FFroehlich:


FFroehlich:
Should fix most of the bugs/issues that sonarcloud marks as "blocker", bit disappointed that sonarcloud doesn't track this in any meaningfull way.

dweindl:
> bit disappointed that sonarcloud doesn't track this in any meaningfull way.

Same here. Should show stats for removed code smells. Feature request? :)

include/amici/defines.h
```diff
@@ -18,37 +18,37 @@ constexpr double pi = 3.14159265358979323846;
...
+
+constexpr int AMICI_PREEQUILIBRATE=           -1;
 
 #ifndef booleantype
 #define booleantype int
 #endif
 
 #ifndef FALSE
-#define FALSE 0
+constexpr int FALSE=0;
```
dweindl, line 47: With the replacements below, is this still being used anywhere?

FFroehlich: unfortunately yes, the solver has to use it in a couple of places.

FFroehlich: done, we are gonna wait for CI for a while anyways :D

PR #917, opened 2020-01-21
Adjoint sensitivities, fix #876

paszkow:
Computing adjoint sensitivities by solving a linear system for the steady-state solution of the ode problem.

PR #1075, opened 2020-05-06
Enable Newton solver with simulation

paulstapor:
This is a suggestion how to fix an existing problem in the steadystate solver (happy about alternative suggestions how to handle it):
Currently, if a model has the timepoint `inf` specified and forward sensis and Newton's method fails due to a poor initial guess, the steadystate solver will integrate a long time with full forward sensis.
However, this is not ideal (or even completely prohibitive) some applications:
If the Jacobian is invertible, it would be fine to simulate without sensis and then just solve the linear system of equations. This can be way faster...

One point I'm not happy about in this solution: I had to implement a function in the solver code to switch off forward sensi computation. As this one is to be used in steadystateproblem, it must be part of the public API of solver. However, this function touches stuff in the solver object, that should imho *not* be touchable by a public API...
Hence, I would be happy about suggestions how to solve this better...


paulstapor:
Okay, it is 100% reasonable, that the failing test does indeed fail. I'm just confused about the fact that it worked beforehand... 🤔 😕 
Using the steady-state solver with adjoints outside preequilibration is not supposed to have a well-defined behavior (yet, will change with #917) and should trigger some raise, actually.
It's interesting that we have a test on this. It is substantially more stunning, that the test passed beforehand... :joy:

I will add a jupyter notebook as documentation of the steadystate solver within this PR, which explains what the steadystate solver *should* do... (in my opinion). Happy to excahnge opinions on that...

FFroehlich:
> This is a suggestion how to fix an existing problem in the steadystate solver (happy about alternative suggestions how to handle it):
> Currently, if a model has the timepoint `inf` specified and forward sensis and Newton's method fails due to a poor initial guess, the steadystate solver will integrate a long time with full forward sensis.
> However, this is not ideal (or even completely prohibitive) some applications:
> If the Jacobian is invertible, it would be fine to simulate without sensis and then just solve the linear system of equations. This can be way faster...
> 
> One point I'm not happy about in this solution: I had to implement a function in the solver code to switch off forward sensi computation. As this one is to be used in steadystateproblem, it must be part of the public API of solver. However, this function touches stuff in the solver object, that should imho _not_ be touchable by a public API...
> Hence, I would be happy about suggestions how to solve this better...

Thats part of the public API of CVODES/IDAS, so I don't see any reason why we shouldn't expose this functionality in AMICI. The implementation absolutely fine from my point of view.

FFroehlich:
> Okay, it is 100% reasonable, that the failing test does indeed fail. I'm just confused about the fact that it worked beforehand... 🤔 😕
> Using the steady-state solver with adjoints outside preequilibration is not supposed to have a well-defined behavior (yet, will change with #917) and should trigger some raise, actually.
> It's interesting that we have a test on this. It is substantially more stunning, that the test passed beforehand... 😂
> 
> I will add a jupyter notebook as documentation of the steadystate solver within this PR, which explains what the steadystate solver _should_ do... (in my opinion). Happy to excahnge opinions on that...

True, we never actually check whether `sllh` has any meaningful content for that test, so this might have passed just by chance. Fine with me to temporarily disable the `amici.SensitivityMethod_adjoint` case and readd in #917.

FFroehlich:
Please have a look at my changes and if tests pass and everything looks fine, feel free to merge!

python/tests/test_preequilibration.py
```diff
@@ -243,26 +243,36 @@ def test_raise_postequilibration_with_adjoints(preeq_fixture):
...
         solver.setSensitivityMethod(sensi_meth)
         solver.setNewtonMaxSteps(0)
         solver.SteadyStateSensitivityMethod = \
-            amici.SteadyStateSensitivityMode_simulationFSA
+            amici.SteadyStateSensitivityMode.simulationFSA
         # add rdatas
         rdatas[sensi_meth] = amici.runAmiciSimulation(model, solver, edata)
+        assert rdatas[sensi_meth]['status'] == amici.AMICI_SUCCESS
 
     for variable in ['llh', 'sllh']:
```
paulstapor, line 267: I didn't add it in my commits, but checked it rather by hand:
In an ideal world, also sx0 and sx_ss are checked for here...

paulstapor, line 267:
```suggestion
    for variable in ['llh', 'sllh', 'sx_ss', 'sx0']:
```

python/tests/test_preequilibration.py
```diff
@@ -234,3 +235,50 @@ def test_parameter_in_expdata(preeq_fixture):
...
+
+    rdata = amici.runAmiciSimulation(model, solver, edata)
+    assert rdata['status'] == amici.AMICI_ERROR
+
+    edata.t_presim = 0.0
+    edata.fixedParametersPresimulation = ()
+
+    rdatas = {}
+    for sensi_meth in [amici.SensitivityMethod.forward,
+                       amici.SensitivityMethod.adjoint]:
```
paulstapor, line 257: That's just a question, not necessary, but:
Do we also want to not only check FSA and ASA, but also FD?

python/tests/test_preequilibration.py
```diff
@@ -234,3 +235,50 @@ def test_parameter_in_expdata(preeq_fixture):
...
+        # set sensi method
+        solver.setSensitivityMethod(sensi_meth)
+        solver.setNewtonMaxSteps(0)
+        solver.SteadyStateSensitivityMethod = \
+            amici.SteadyStateSensitivityMode.simulationFSA
```
paulstapor, line 262: Add another loop for Newton solver as well?
Or are we `overtesting` then?

PR #1081, opened 2020-05-07
sbml import, fix conversion between formats

LarsFroehling:
Wrap the MathML string with a header so it can be properly read by `readMathMLFromString` which is expecting an XML string

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -1032,7 +1032,12 @@ class SbmlImporter:
 
                     nested_rule_math_ml = sp.printing.mathml(nested_formula)
 
-                    nested_rule_math_ml_ast_node = sbml.readMathMLFromString(nested_rule_math_ml)
+                    header = "<?xml version='1.0' encoding='UTF-8'?>\n " \
+                             "<math xmlns='http://www.w3.org/1998/Math/MathML'>\n"
+                    footer = "</math>"
+                    nested_rule_math_ml_wrapped = f'{header}{nested_rule_math_ml}{footer}'
+
+                    nested_rule_math_ml_ast_node = sbml.readMathMLFromString(nested_rule_math_ml_wrapped)
 
                     if nested_rule_math_ml_ast_node is None:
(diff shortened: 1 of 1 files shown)


LarsFroehling:
I am not entirely sure but I think there are still issues with the `nested_rule_math_ml = sp.printing.mathml(nested_formula)` part for certain parameter names. 
E.g. for me `k_1` would become `<ci><mml:msub><mml:mi>k</mml:mi><mml:mi>1</mml:mi></mml:msub></ci>` which can't be parsed later and would result in an SBMLException. I am not sure how to solve this though.

LarsFroehling:
The built on windows seems to fail since it can't find `cblas.h` https://travis-ci.com/github/ICB-DCM/AMICI/jobs/329543992#L4815

I am not sure how this is connected to the changes above.

FFroehlich:
> The built on windows seems to fail since it can't find `cblas.h` https://travis-ci.com/github/ICB-DCM/AMICI/jobs/329543992#L4815
> 
> I am not sure how this is connected to the changes above.

usually a hiccup in the CI, should go away after restarting, don't worry about it.

FFroehlich:
> I am not entirely sure but I think there are still issues with the `nested_rule_math_ml = sp.printing.mathml(nested_formula)` part for certain parameter names.
> E.g. for me `k1` would become `<ci><mml:msub><mml:mi>k</mml:mi><mml:mi>1</mml:mi></mml:msub></ci>` and `k_p` would become `<ci><mml:msub><mml:mi>k</mml:mi><mml:mi>p</mml:mi></mml:msub></ci>` which can't be parsed later and would result in an SBMLException. I am not sure how to solve this though.

We don't have to ensure every last bit of compatibility with SBML. As long as the importer throws an appropriate exception, we are generally fine. If you have identified a specific issue and can throw a more informative error message on how to resolve it (remove the `_` in the parameter name) thats even better.

FFroehlich:
gonna merge now for the 0.11.0 release, if you find a better solution feel free to open another PR

PR #1086, opened 2020-05-11
Fix MathML conversion

lcontento:
SymPy's default MathML printer replaces underscores with subscripts, resulting in a MathML formula unsupported by libSBML.
For example,
```
import sympy
x = sympy.symbols('x_i')
sympy.mathml(x)
```
results in
```
<ci><mml:msub><mml:mi>x</mml:mi><mml:mi>i</mml:mi></mml:msub></ci>
```

Since I recycled the MathML writer code from my own project, it also includes the correct SBML way of saving the time variable. Since the result of MathML printing in AMICI will never actually be used to produce a SBML file, omitting such step would be possible in this case, but I left it in for completeness' sake.


FFroehlich:
Great solution 👍

PR #1094, opened 2020-05-25
fix, power function call in unscale_parameter

plakrisenko:


dweindl:
> why did no linting tool find this???

because we ignore those due to lots of false positives

Issue #1095, opened 2020-05-25
Newest sympy version breaks PEtab test suite


elbaraim:
The sympy package got updated to version 1.6, leading to failure of the tests for the PEtab suite: see trace in https://travis-ci.com/github/ICB-DCM/pyPESTO/builds/168030743

This is repeated for some of the cases in the PEtab suite:
```
ERROR:test.test_petab_suite:Cannot set values of <class 'sympy.matrices.immutable.ImmutableDenseMatrix'>
```

yannikschaelte:
I can confirm that it fails for sympy 1.6, works for 1.5.1. actually, quite a bit changed in the release https://github.com/sympy/sympy/wiki/release-notes-for-1.6

FFroehlich:
Will try to have a look tomorrow.

PR #1096, opened 2020-05-25
Workaround for #1095 - pin sympy to < 1.6 until symbolic code is updated

dweindl:
#1095


dweindl:
Maybe easier to just fix it...

PR #1098, opened 2020-05-26
Fixes sympy, fixes #1095

### LINKED ISSUE #1095: Newest sympy version breaks PEtab test suite

FFroehlich:
waiting for https://github.com/pysb/pysb/pull/504

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -1488,6 +1488,8 @@ class ODEModel:
             )
 
             if dx0_fixed_parametersdx.is_zero is not True:
+                if isinstance(self._eqs[name], ImmutableDenseMatrix):
+                    self._eqs[name] = MutableDenseMatrix(self._eqs[name])
                 for ip in range(self._eqs[name].shape[1]):
                     self._eqs[name][:, ip] += smart_multiply(
                         dx0_fixed_parametersdx, self.sym('sx0')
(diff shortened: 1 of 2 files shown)


python/amici/ode_export.py
```diff
@@ -1294,6 +1295,13 @@ def _generate_symbol(self, name: str) -> None:
                 if state.conservation_law is None
             ])
             return
+        elif name == 'sx0':
+            self._syms[name] = sp.Matrix([
```
dweindl, line 1299: Is that used somewhere?

paulstapor: I think its just for having more indicative names for the symbols than the default ones, which are currently generated...

FFroehlich: No it actually had the wrong dimensions. This is theoretically used for `sx0_fixedParameters`, but currently practically not used since the matrix is multiplied with is always zero (which is good since sx0 is missing from the function signature and would thus lead to failures, so this fixes some of the issues but not all).

Issue #1073, opened 2020-05-06
Clean-up steady-state solver code


paulstapor:
Currently, the workSteadyStateProblem routine isn't exactly the nicest part of Amici...
Things to be changed here in some close future:
 - [x] Remove the triple try-catch nesting, when determining whether or not to use simulation 
 - [x] Allow steady-state sensi computation via solving the linear system, if steady-state was computed via simulation: Currently, that's not possible and full FSA computation is carried out, which is fatal for large models.
 - [x] Add flag "simulationOnly" to steadyStateSensitivityMethod, which does not even try to launch the Newton solver.
 - [x] ~#1675~
 - [x] Merge #917 
 - [x] ~Replace custom implementation of iterative linear solver in Newton Solver by~ Interface GMRES, BICGSTAB, and TFQMR (do it analogously to solver.cpp) (#1164 )
 - [ ] correct the (partly) erroneous or misleading output into rdata of the steady state solver
 - [ ] create figure for documenting the steady state solver code via sphinx, rather than having something prerendered
 - [x] #1737

dweindl:
>     * [ ]  Remove the triple try-catch nesting, when determining whether or not to use simulation

Very much in favor. This is a huge pain to debug.

FFroehlich:
> * [ ]  correct the (partly) erroneous or misleading output into rdata of the steady state solver

I have no idea what this refers to, anyone?

dweindl:
Closing this.

Addressed, except for:

>     * [ ]  correct the (partly) erroneous or misleading output into rdata of the steady state solver

Nobody seems to know what that was about. Unlikely to change.

>     * [ ]  create figure for documenting the steady state solver code via sphinx, rather than having something prerendered

Part of https://github.com/AMICI-dev/AMICI/issues/2582

PR #1099, opened 2020-05-27
adjoints + steadystate

paulstapor:
Just want to see whether checks pass.
PR for this one will be done by @paszkow from his fork, after confirming his scripts still run and results are correct.


paulstapor:
Okay, quite some sonarcloud issues to address. Really happy we have that in place! :)
Will do soonish.

paulstapor:
So, from my side, this should be mergable now.
Doing some last checks on this with @paszkow , but beyond that, the code should be fine...

include/amici/rdata.h
```diff
@@ -225,24 +225,27 @@ class ReturnData {
     /** employed order forward problem (dimension: nt) */
     std::vector<int> order;
 
-    /** computation time of forward solve [s] */
+    /** computation time of forward solve [ms] */
```
paulstapor, line 228: We used to have time reporting not uniformly. At least for one of them, the unit was incorrect. Now, everything should actually be in milliseconds.
No preference, whether seconds or milliseconds are better. We just need to have it the same way for all cpu times...

dweindl: preferring seconds. but either is fine, if coherently used. good to include it already in the variable/function name.

paulstapor: @paszkow proposed to use the chorno library (https://en.cppreference.com/w/cpp/chrono), which looks nicer and may be more precise than what we currently have... 
Maybe something for the future?

src/backwardproblem.cpp
```diff
@@ -59,44 +65,55 @@ void BackwardProblem::workBackwardProblem() {
...
 
     int it = model->nt() - 1;
-    model->initializeB(xB, dxB, xQB);
-    handleDataPointB(it);
-    solver->setupB(&which, model->getTimepoint(it), model, xB, dxB, xQB);
+    /* If we have posteq, infty timepoints were already treated */
+    for (int jt = model->nt() - 1; jt >= 0; jt--)
+        if (std::isinf(model->getTimepoint(it)))
+            --it;
+    if (it < model->nt() - 1){
```
FFroehlich, line 72: cant we just initialize `xQB` to zeros for `it < model->nt() - 1` and pass it to `initializeB`? I think that would be better readable. Also shouldn't if/else cases be switched here?

src/backwardproblem.cpp
```diff
@@ -59,44 +65,55 @@ void BackwardProblem::workBackwardProblem() {
...
-            it--;
+            /* reinit states */
+            solver->reInitB(which, t, xB, dxB);
+            solver->quadReInitB(which, xQB);
         }
 
-        /* reinit states */
-        solver->reInitB(which, t, xB, dxB);
-        solver->quadReInitB(which, xQB);
+        /* we still need to integrate from first datapoint to tstart */
```
FFroehlich, line 109: I believe this code should be outside the loop, we only want to run this once.

src/newton_solver.cpp
```diff
@@ -236,6 +262,43 @@ void NewtonSolverIterative::prepareLinearSystem(int ntry, int nnewt) {
...
+                           "computation for steady state problems.");
+    }
+
+    // Get the Jacobian and its diagonal for preconditioning
+    model->fJB(*t, 0.0, *x, dx, xB, dxB, xdot, ns_J.get());
+
+    // Get the diagonal and ensure negativity of entries is ns_J. Note that diag(JB) = -diag(J).
+    model->fJDiag(*t, ns_Jdiag, 0.0, *x, dx);
+
+    ns_p.set(1.0);
```
FFroehlich, line 294: duplicate code, move to seperate function?

src/newton_solver.cpp
```diff
@@ -271,7 +325,7 @@ void NewtonSolverIterative::linsolveSPBCG(int ntry, int nnewt,
     double alpha = 1.0;
 
     // can be set to 0 at the moment
```
FFroehlich, line 327: I believe this comment is deprecated

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 #
 # Build AMICI library
 #
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.12)
```
dweindl, line 4: What was that for if there is no other change in CMake files?

src/newton_solver.cpp
paulstapor: yikes. This code will be removed soon anyway, current issue, see #1073.
This is really an old handcrafted reimplementation of BICGSTAB. Prefer not to do changes now, but to fully remove and correctly reimplement the remaining linear solvers (i.e., all except DENSE and KLU).

src/backwardproblem.cpp
```diff
@@ -59,44 +65,55 @@ void BackwardProblem::workBackwardProblem() {
     }
 
     int it = model->nt() - 1;
-    model->initializeB(xB, dxB, xQB);
-    handleDataPointB(it);
-    solver->setupB(&which, model->getTimepoint(it), model, xB, dxB, xQB);
+    /* If we have posteq, infty timepoints were already treated */
+    for (int jt = model->nt() - 1; jt >= 0; jt--)
+        if (std::isinf(model->getTimepoint(it)))
```
dweindl, line 70:
```suggestion
        if (std::isinf(model->getTimepoint(jt)))
```

?

Otherwise this loop does not make much sense, does it?

src/backwardproblem.cpp
```diff
@@ -59,44 +65,55 @@ void BackwardProblem::workBackwardProblem() {
...
+            --it;
+    if (it < model->nt() - 1){
+        model->initializeB(xB, dxB, dxB);
+    } else {
+        model->initializeB(xB, dxB, xQB);
+    }
 
-    --it;
+    if ((it >= 0 || discs.size() > 0) && model->getTimepoint(it) > model->t0())
```
dweindl, line 78:
```suggestion
    if ((it >= 0 || !discs.empty()) && model->getTimepoint(it) > model->t0())
```

CMakeLists.txt
paulstapor: @paszkow pointed out that using the option `--parallel` requires at least 3.12., this is why I just put it...

dweindl: Hmm... that would have to be checked in the build script. Here it is too late.

FFroehlich:
are JB and JSparseB generated in both MATLAB and Python interfaces?

src/newton_solver.cpp
```diff
@@ -12,11 +12,13 @@
 #include <cstring>
 #include <ctime>
 #include <cmath>
+#include <iostream>
```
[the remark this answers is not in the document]
paulstapor: True, aye... Looks old...

CMakeLists.txt
paulstapor: okay, I see. ... yes, okay. I'll remove for the moment. Let's do that later and not mix up too many things.

paulstapor:
> are JB and JSparseB generated in both MATLAB and Python interfaces?

Yes, also Matlab generates JB and JSparseB

src/backwardproblem.cpp
paulstapor: The problem ist that, if we have postequilibration (which is the case for `it < model->nt() - 1`), xQB already contains the quadratures from backward postequilibration. Hence, in that case, we don't want to overwrite them, but still want to reinitialize xB and dxB.
I have another idea, which should simplify this. Will commit.

src/backwardproblem.cpp
paulstapor: True. Done.

paulstapor:
If tests pass now, would be strongly in favor of merging now... 
If anyone present, for any reason knows why these changes should not be merged, let him speak now or forever hold his peace.

PR #1106, opened 2020-05-29
cache edatas for simulate_petab

yannikschaelte:
Allow to pass edatas to the `simulate_petab` function to improve performance.


python/amici/petab_objective.py
```diff
@@ -25,6 +25,7 @@
 logger = get_logger(__name__)
 
 AmiciModel = Union[amici.Model, amici.ModelPtr]
+ExpData = Union[amici.ExpData, amici.ExpDataPtr]
```
dweindl, line 28: Maybe we want to define them in `__init__.py`?

python/amici/petab_objective.py
```diff
@@ -63,6 +65,8 @@ def simulate_petab(
     :param simulation_conditions:
```
dweindl, line 65: add: meaningless if edatas and parameter_mapping is passed

python/amici/petab_objective.py
yannikschaelte: > Title misleading, as it doesn't cache, but only allows for caching them outside. Otherwise good.

True

> Maybe we want to define them in __init__.py?

I can do that. Only these two or is there more where x and xPtr should be used synonymously?

yannikschaelte: fun fact: it is defined in `__init__.py` already :rofl:

dweindl: Why not just use those then... :D

dweindl: > I can do that. Only these two or is there more where x and xPtr should be used synonymously?

Not aware of any, but possible

yannikschaelte: > Why not just use those then... :D

Done ;)

python/amici/petab_objective.py
yannikschaelte: added comment, and generate below only if required.

PR #1109, opened 2020-05-29
Speed-up computation of sx0, reduce file size

### LINKED ISSUE #1107: Find out why model import got slower with recent sympy
Partially recovered with #1104, but still not back at previous speed...

dweindl:
Resolves #1107

* Keep index list of subset of states for which x0_fixedParameters was computed
* Only write non-zero values to sx0 function file


python/amici/ode_export.py
```diff
@@ -2372,7 +2372,8 @@ def _get_function_body(self,
                 for index, formula in enumerate(
                         self.model.eq('x0_fixedParameters')
```
FFroehlich, line 2373:
```suggestion
                        self.model.eq('x0')
```

python/amici/ode_export.py
```diff
@@ -2372,7 +2372,8 @@ def _get_function_body(self,
...
                 ):
-                    if formula != 0 and formula != 0.0:
+                    if formula != 0 and formula != 0.0 \
```
FFroehlich, line 2375:
```suggestion
                    if any([sym in formula.free_symbols for sym in k]):
```

python/amici/ode_export.py
```diff
@@ -2372,7 +2372,8 @@ def _get_function_body(self,
...
+                            and symbol[index, ipar] != 0:
```
FFroehlich, line 2376:
```suggestion
```

python/amici/ode_export.py
FFroehlich: alternatively the respective index set can be stored when computing `fx0_fixedParameters`

FFroehlich:
`fsx0_fixedParameters`, as with `fx0_fixedParameters`, is a special case where one actually needs to overwrite with (possibly zero) values whenever `eq` contains `k`

FFroehlich:
> `fsx0_fixedParameters`, as with `fx0_fixedParameters`, is a special case where one actually needs to overwrite with (possibly zero) values whenever `eq` contains `k`

this may actually be exploited further since `fsx0_fixedParameters` will only be required for that index set. So if that index-set is stored, computation of `fx0_fixedParameters` and `fsx0_fixedParameters` can be limited to that index set and expanded to full dimensionality when writing the respective cpp files.

dweindl:
Thanks for input. Sounds like something you could do 50x faster than me :grimacing:

FFroehlich:
should be waaaay faster now :)

dweindl:
> should be waaaay faster now :)

:crossed_fingers:

FFroehlich:
> > FAILED tests/petab_test_suite/test_petab_suite.py::test_case[0010] - assert F...
> 
> > FD check  failed for llh at index ip=2 for 1 indices:
> > max(adev): 0.2742372404596072, max(rdev): 0.9996354851429122
> 
> Something doesn't fit

should be fixed now

dweindl:
> should be waaaay faster now :)

It is. Better than earlier performance. Thanks!

> OKAY: 2018.81 < 2500

FFroehlich:
> OKAY: 1861.80 < 2500
💪

Issue #1128, opened 2020-06-10
Best way to implement fully adjoint handling of preequilibration


paulstapor:
Hi,

Before I start fiddling around in parts of the C++ functionality I'm not yet familiar with, I wanted to ask for opinions:

My overall aim is to implement a fully adjoint handling of the post- and pre-equilibration logic, as this will be extremely helpful for working with some large-scale models. Here's how I imagine it (in rough terms):

In amici.cpp, we currently have the following logic
```
1. work preequilibration
2. work forward problem
3. work postequilibration
4. work backward problem
Last. processReturnData
```
#1099 adds work postequilibration backward to it. My aim (and the reason, why I'm asking this question) would be adding: `5. work preequilibration backward`. 

Technically, this could already be implemented after merging #1099 (or #917). However, it would still still restrict the fully adjoint usage to systems with well-conditioned Jacobian. Hence, I already started working on a simulation based version of #1099, which works also for singular Jacobians.

Here's what I would need to make things happen:
1. In the beginning of `workFowardProblem`, the function `updateAndReinitStatesAndSensitivities` is called. A corresponding counterpart would be needed in the end of `workBackwardProblem`. I've been writing down the math for it. As long as (A) we only have an update on the constant parameters (or fixed states) and not in the solver states (`x_solver`), this amounts to a heaviside function at a fixed timepoint. This does not influence the adjoint state, luckily. Hence, implementation is pretty doable.
(B) If we have other types of updates (really updating solver states, e.g., due to conserved quantities), this will be more complicated and not work right away.
2. To discern cases (A) and (B), it would be good to add a member to every model (during python generation), which indicates whether we have the easy case (A), or whether the model has complicated conservation laws (causing case (B)).

What do you think about it?

FFroehlich:
Agree, sounds like a reasonable setup. Also agree that such an indicator would make sense.

paulstapor:
fixed on develop now...

PR #1129, opened 2020-06-12
Refactor SteadystateProblem triple nested try catch clause

paulstapor:
This should at least resolve the nested try catch and reduce the overall complexity of the steady state solver.


src/steadystateproblem.cpp
```diff
@@ -148,6 +83,89 @@ void SteadystateProblem::workSteadyStateProblem(Solver *solver, Model *model,
                                     SensitivityOrder::first);
 }
 
+NewtonStatus SteadystateProblem::findSteadyState(Solver *solver,
+                                                 NewtonSolver *newtonSolver,
+                                                 Model *model, int it) {
+    /* First, try to run the Newton solver */
```
FFroehlich, line 89: could we flatten/deduplicate this further by introducing `findSteadyStateNewton` and `findSteadyStateSimulation`  where exception behaviour is modulated by a flag which is only passed on the second call to `findSteadyStateNewton`?

Would also be nice if we wouldn't have to have duplicate code to copy of the number of steps. Could be implemented by a context manager which is either explicitly in the exception or goes out of context when the status is returned, see https://github.com/ICB-DCM/AMICI/blob/ffb5f33001047556a718c504ad17bc81ed3e206d/include/amici/forwardproblem.h as example

dweindl: > context manager

Wouldn't use too many of those. In python they are fine, but in C++ usually we don't want to have side effects in destructors. Especially destructors should never `throw`.

src/steadystateproblem.cpp
```diff
@@ -83,30 +84,67 @@ void SteadystateProblem::workSteadyStateProblem(Solver *solver, Model *model,
...
+                steady_state_status[0] = SteadyStateStatus::failed_damping;
+                break;
+            default:
+                steady_state_status[0] = SteadyStateStatus::failed;
+        }
+    }
+
+    /* copy number of linear steps used */
+    if (maxSteps > 0)
+        std::copy_n(newtonSolver->getNumLinSteps().begin(),
```
FFroehlich, line 134: I believe this needs a check for `newton_retry` to switch between `numlinsteps.begin()` and `&numlinsteps.at(maxSteps)`

src/rdata.cpp
```diff
@@ -188,11 +190,11 @@ void ReturnData::processPreEquilibration(SteadystateProblem const &preeq,
         for (int ip = 0; ip < nplist; ip++)
             writeSlice(sx_rdata[ip], slice(sx_ss, ip, nx));
     }
-    /* Get cpu time for Newton solve in seconds */
-    preeq_cpu_time = preeq.getCPUTime() / 1000;
-    preeq_status = static_cast<int>(preeq.getNewtonStatus());
+    /* Get cpu time for Newton solve in milliseconds */
```
FFroehlich, line 193: cpu time for simulation is reported in ms though, right?

paulstapor: I wanted to have things in the same unit... Now, everything should be in milliseconds, hopefully, but will recheck

src/steadystateproblem.cpp
paulstapor: I really appreciate the idea of further flattening out the logic. I think that makes it much more transparent.

src/steadystateproblem.cpp
paulstapor: True! Thanks!

src/steadystateproblem.cpp
```diff
@@ -191,9 +262,20 @@ bool SteadystateProblem::checkConvergence(const Solver *solver, Model *model) {
     return converged;
 }
 
+bool SteadystateProblem::checkSteadyStateSuccess() {
+    /* Did one of the attempts yield s steady state? */
+    if (steady_state_status[0] == SteadyStateStatus::success ||
```
FFroehlich, line 267: I think using `std::any_of` would be more idiomatic. Should be something like `std::any_of(steady_state_status.begin(), steady_state_status.end(), [](SteadyStateStatus status){return status == SteadyStateStatus::success;})`

May even be shorter but definitely better readability and easier to mantain if we decide to extend `steady_state_status`.

src/steadystateproblem.cpp
FFroehlich: Agreed, much cleaner. Pretty happy with the result, thanks for putting in the effort!

paulstapor: Welcome and thanks! :)
As however, we need some extended functionality in this thing (the fully adjoint preequilibration) our large scale project, leaving the code in the 2017-version was not really an option any more... ;)
There will be two more PRs, unfortunately, until then:
1. The fully adjoint preequilibration, first only using the algebraic solution (i.e., solving the linear system of equations), see #1128 .
2. Adding an option to workSteadyStateBackward (then automatically preeq and posteq) to use simulation, including interfacing the cvodes api for the forward simulation quadratures...

However, then preequilibration should have become solid, also for really large scale models with whatever conservation laws...

paulstapor:
Would merge then, if it's fine.

PR #1131, opened 2020-06-15
Various cleanup

dweindl:
Remove a number of code smells

### DIFF
--- a/src/backwardproblem.cpp
+++ b/src/backwardproblem.cpp
@@ -45,7 +45,7 @@ BackwardProblem::BackwardProblem(const ForwardProblem &fwd,
                                  fwd.model->nx_solver * fwd.model->nJ));
             }
         }
-        
+
     }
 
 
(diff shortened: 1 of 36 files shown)


src/backwardproblem.cpp
```diff
@@ -131,12 +131,12 @@ void BackwardProblem::handleEventB() {
                                           xdot_disc,
                                           xdot_old_disc);
 
-        for (int ix = 0; ix < model->nxtrue_solver; ++ix) {
-            for (int iJ = 0; iJ < model->nJ; ++iJ) {
-                if (model->nz > 0) {
+        if (model->nz > 0) {
+            for (int ix = 0; ix < model->nxtrue_solver; ++ix) {
+                for (int iJ = 0; iJ < model->nJ; ++iJ) {
```
FFroehlich, line 136: better vectorization if we swap around loops over ix and jJ?

dweindl: For rhs better as is, for lhs better if swapped. Improvement unclear, I'd say. But anyhow not vectorizable because loop body can throw.

PR #1137, opened 2020-06-18
Speedup modelname.cpp compilation by using char arrays instead of std…

dweindl:
…::string

Follow-up to #1105

tested with icpc (ICC) 19.0.5.281 20190815; -O3:

> Elapsed (wall clock) time (h:mm:ss or m:ss): 0:01.17
> Maximum resident set size (kbytes): 249984

before, terminated at:
> Elapsed (wall clock) time (h:mm:ss or m:ss): 2:19:33
> Maximum resident set size (kbytes): 443383356
> Fatal compilation error: Out of memory asking for 18446744072184442880.

### DIFF
--- a/.github/workflows/test-large-model.yml
+++ b/.github/workflows/test-large-model.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - develop
       - master
-      - feature_fast_index_compilation
+      - feature_speedup_compilation
 
   pull_request:
     branches:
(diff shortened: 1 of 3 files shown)


FFroehlich:
Why though?

dweindl:
> Why though?

Not completely sure. Wasn't able to find the specific optimization causing the problem. Would guess that it's related to somehow trying to inline or rearrange the 30K std::string() calls.

PR #1141, opened 2020-06-22
Fix and cleanup extension import logic

### LINKED ISSUE #1140: Development installation of AMICI fails if incompatible extension exists from previous installation
... due to e.g.:
>     AttributeError: module 'amici._amici' has no attribute 'ReturnDataPtr_posteq_cpu_timeB_get'

Could check the call stack for `setup.py`, and in that case ignore `AttributeErrors` or completely skip import of `_amici` in `__init__.py`. May also simplify the current import logic.

dweindl:
* Don't import AMICI extension when amici is imported from setup.py (Closes #1140)

* Cleanup import logic: Removing unused SWIG-generated files and import based on existance of remaining ones

* Avoid unnecessary imports when called from setup.py

### DIFF
--- a/python/amici/__init__.py
+++ b/python/amici/__init__.py
@@ -81,14 +81,15 @@ def _imported_from_setup() -> bool:
     # in case we are imported from setup.py, this will be the AMICI package
     # root directory (otherwise it is most likely the Python library directory,
     # we are not interested in)
-    package_root = os.path.dirname(os.path.dirname(__file__))
+    package_root = os.path.realpath(os.path.dirname(os.path.dirname(__file__)))
 
     for frame in getouterframes(currentframe()):
         # Need to compare the full path, in case a user tries to import AMICI
         # from a module `*setup.py`. Will still cause trouble if some package
         # requires the AMICI extension during its installation, but seems
         # unlikely...
-        if frame.filename == os.path.join(package_root, 'setup.py'):
+        frame_path = os.path.realpath(os.path.expanduser(frame.filename))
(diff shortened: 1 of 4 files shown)


python/amici/__init__.py
```diff
@@ -84,41 +106,43 @@ def _get_commit_hash():
...
-                # `setup` or after an installation with `--no-clibs`.
+if not _imported_from_setup():
+    try:
+        # Try importing AMICI SWIG-interface with HDF5 enabled
+        from . import amici
+        from .amici import *
+        hdf5_enabled = True
+        has_clibs = True
+    except ModuleNotFoundError:
+        # import from setuptools or installation with `--no-clibs`
```
FFroehlich, line 117: can we also detect this using frame inspection, this would reduce remaining complexity substantially.

dweindl: The first part of the comment is now obsolete (will remove). The second part is maybe misleading, it's not during installation, it's after being installed with `--no-clibs`. We could circumvent that by writing a "configuration" file during installation to include whether we have HDF5 support or have the extension at all. That would clean up that mess quite well :)

FFroehlich: How about `rm amici_without_hdf5.py` if hdf5 was detected during installation and `rm amici.py;mv amici_without_hdf5.py amici.py` if it was not?

dweindl: Sounds easier. Should do the job as well.
That this removes the possibility of rebuilding the extension from the installed package later on is probably not of practical relevance...

PR #1143, opened 2020-06-25
Backward preequlibration

### LINKED ISSUE #1128: Best way to implement fully adjoint handling of preequilibration

paulstapor:
Closes #1128. So far only for the algebraic solution. Simulation-based solution will be another PR.


paulstapor:
What happened to SonarCloud? Or does that only trigger after successful build?

paulstapor:
I could add tests on the Exceptions of the steadystate solver, as those seem to be not covered by tests atm. Any opinion? Or rather go for another PR then?

Sorry this PR involves so many files. Most files have little changes, but some are altered strongly...

include/amici/model.h
```diff
@@ -79,6 +79,8 @@ class Model : public AbstractModel {
      * @param nx_solver number of state variables with conservation laws applied
      * @param nxtrue_solver number of state variables of the non-augmented model
      * with conservation laws applied
+     * @param nx_solver_reinit number of state variables with conservation laws
+     * applied which would be reinitiliazed after preequilibration
```
FFroehlich, line 83:
```suggestion
     * subject to reinitialization
```
(also includes presimulation)

include/amici/model.h
```diff
@@ -258,6 +261,12 @@ class Model : public AbstractModel {
      */
     int ncl() const;
 
+    /**
+     * @brief Number of solver states to be reinitiliazed after preequilibration
```
FFroehlich, line 265:
```suggestion
     * @brief Number of solver states subject to reinitiliazion
```

src/backwardproblem.cpp
```diff
@@ -37,8 +37,8 @@ BackwardProblem::BackwardProblem(const ForwardProblem &fwd,
         for (int it = 0; it < fwd.model->nt(); it++) {
             if (std::isinf(fwd.model->getTimepoint(it))) {
                 if (!posteq)
-                    throw AmiException("Model has non-finite timpoint but"
-                    "postequilibration did not run");
+                    throw AmiException("Model has non-finite timpoint but "
```
FFroehlich, line 40:
```suggestion
                    throw AmiException("Model has non-finite timepoint, but "
```

src/rdata.cpp
```diff
@@ -404,8 +406,36 @@ void ReturnData::processBackwardProblem(ForwardProblem const &fwd,
...
+                sllh.at(ip) -= llhS0[ip] + xQB[ip * model.nJ];
+            } else {
+                s2llh.at(iJ - 1 + ip * (model.nJ - 1)) -=
+                    llhS0[ip + iJ * model.nplist()] + xQB[iJ + ip * model.nJ];
+            }
+        }
+    }
+}
+
+void ReturnData::handleSx0Backward(Model &model,
```
FFroehlich, line 427: lets learn from past mistakes and add some documentation that gives an intuition why what is done here makes sense.

src/steadystateproblem.cpp
```diff
@@ -209,6 +196,69 @@ void SteadystateProblem::findSteadyStateBySimulation(Solver *solver,
...
+        /* If preequilibration but not adjoint mode, there's nothing to do */
+        if (solver->getSensitivityMethodPreequilibration() !=
+            SensitivityMethod::adjoint)
+            return false;
+
+        /* If we need to reinitialize solver states, this won't work yet. */
+        if (model->nx_reinit() > 0)
+            throw NewtonFailure(AMICI_NOT_IMPLEMENTED,
+                "Adjoint preequilibration with reinitialization of "
+                "non-constant states is not xet implemented. Stopping.");
```
FFroehlich, line 212:
```suggestion
                "non-constant states is not yet implemented. Stopping.");
```

paulstapor:
> Looks good, but really hard to judge the math/correctness without additional documentation

Thanks, but fully agree... I'm in the processing of texing these things down...
I could either include them in the notebook on equilibration, in an additional notebook, or start an own texfile on the math, which is somehow included in the Amici documentation. ....
Any preference?

include/amici/model.h
```diff
@@ -1177,6 +1186,9 @@ class Model : public AbstractModel {
      */
     int nxtrue_solver{0};
 
+    /** number of solver states to be reinitilized after preequilibration */
```
dweindl, line 1189:
```suggestion
    /** number of solver states to be reinitialized after preequilibration */
```

include/amici/model_dae.h
```diff
@@ -36,6 +36,8 @@ class Model_DAE : public Model {
      * @param nx_solver number of state variables with conservation laws applied
      * @param nxtrue_solver number of state variables of the non-augmented model
      with conservation laws applied
+     * @param nx_solver_reinit number of state variables with conservation laws
+     * applied which would be reinitiliazed after preequilibration
```
dweindl, line 40:
```suggestion
     * applied which would be reinitialized after preequilibration
```

include/amici/model_ode.h
```diff
@@ -36,6 +36,8 @@ class Model_ODE : public Model {
      * @param nx_solver number of state variables with conservation laws applied
      * @param nxtrue_solver number of state variables of the non-augmented model
      with conservation laws applied
+     * @param nx_solver_reinit number of state variables with conservation laws
+     * applied which would be reinitiliazed after preequilibration
```
dweindl, line 40:
```suggestion
     * applied which would be reinitialized after preequilibration
```

include/amici/rdata.h
```diff
@@ -237,13 +238,18 @@ class ReturnData {
...
+    double preeq_cpu_timeB = 0.0;
+
     /** flags indicating success of steady state solver  (postequilibration) */
     std::vector<SteadyStateStatus> posteq_status;
 
     /** computation time of the steady state solver [ms]  (postequilibration) */
     double posteq_cpu_time = 0.0;
 
-    /** computation time of the Newton solver of the backward problem [ms] */
+    /** computation time of the steady state solverof the backward problem [ms]
```
dweindl, line 251:
```suggestion
    /** computation time of the steady state solver of the backward problem [ms]
```

include/amici/rdata.h
```diff
@@ -633,6 +644,29 @@ class ReturnData {
...
+     * @param preeq Steadystateproblem for preequilibration
+     * @param llhS0 contribution to likelihood for inital state sensitivities
+     * @param xQB vector with quadratures from adjoint computation
+     */
+    void handleSx0Backward(Model &model, SteadystateProblem const &preeq,
+                           std::vector<realtype> &llhS0, AmiVector &xQB);
+
+    /**
+     * @brief Updates contribution to likelihood for inital state sensitivities
+     * (llhS0), if no preequilibration was run or if forward sensis were used
```
dweindl, line 661:
```suggestion
     * (llhS0), if no preequilibration was run or if forward sensitivities were used
```

include/amici/rdata.h
```diff
@@ -633,6 +644,29 @@ class ReturnData {
...
+     * @param model model that was used for forward/backward simulation
+     * @param llhS0 contribution to likelihood for inital state sensitivities
```
dweindl, line 663:
```suggestion
     * @param llhS0 contribution to likelihood for initial state sensitivities
```

include/amici/rdata.h
```diff
@@ -633,6 +644,29 @@ class ReturnData {
      */
     void getEventSensisFSA(int iroot, int ie, realtype t, Model &model,
                            ExpData const *edata);
+
+    /**
+     * @brief Updates contribution to likelihood for inital state sensitivities
+     * (llhS0), if preequilibration was run in adjoint mode
+     * @param model model that was used for forward/backward simulation
+     * @param preeq Steadystateproblem for preequilibration
+     * @param llhS0 contribution to likelihood for inital state sensitivities
```
dweindl, line 653:
```suggestion
     * @param llhS0 contribution to likelihood for initial state sensitivities
```

include/amici/steadystateproblem.h
```diff
@@ -79,6 +92,15 @@ class SteadystateProblem {
                                      Model *model,
                                      int it);
 
+    /**
+     * Compute the quadrature in steady state abckward mode by solvingthe
```
dweindl, line 96:
```suggestion
     * Compute the quadrature in steady state backward mode by solving the
```

include/amici/steadystateproblem.h
```diff
@@ -277,6 +301,18 @@ class SteadystateProblem {
...
+     * @brief Accessor for xQB
+     * @return xQB
+     */
+    AmiVector const& getAdjointQuadrature() const { return xQB; }
+
+    /**
+     * @brief Accessor for hasQuadrature
+     * @return hasQuadrature
+     */
+    const bool getHasQuadrature() const { return hasQuadrature; }
```
dweindl, line 314:
```suggestion
    const bool hasQuadrature() const { return hasQuadrature_; }
```

(and rename ḿember)

python/tests/test_compare_conservation_laws_sbml.py
```diff
@@ -135,5 +161,33 @@ def test_compare_conservation_laws_sbml(edata_fixture):
     )
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore")
-        rdata = get_results(model_without_cl, edata=edata_fixture, sensi_order=1)
+        rdata = get_results(model_without_cl, edata=edata, sensi_order=1)
         assert rdata['status'] == amici.AMICI_ERROR
+
+def test_adjoint_preequilibration(edata_fixture):
+    # get the model
+    sys.path.insert(0, os.path.abspath('model_constant_species_cl'))
```
dweindl, line 169: `amici.import_model_module()`

src/steadystateproblem.cpp
```diff
@@ -209,6 +196,69 @@ void SteadystateProblem::findSteadyStateBySimulation(Solver *solver,
...
+                                                      Model *model) {
+    /* compute the integral over the adjoint state xBintegral */
+    try {
+        newtonSolver->prepareLinearSystemB(0, -1);
+        newtonSolver->solveLinearSystem(xB);
+    } catch (NewtonFailure const &ex) {
+        if (ex.error_code == AMICI_SINGULAR_JACOBIAN)
+            throw NewtonFailure(ex.error_code, "Steady state backward "
+                                "computation failed due to unsuccessful "
+                                "factorization of RHS Jacobian. ");
```
dweindl, line 236:
```suggestion
                                "factorization of RHS Jacobian.");
```

src/steadystateproblem.cpp
```diff
@@ -209,6 +196,69 @@ void SteadystateProblem::findSteadyStateBySimulation(Solver *solver,
...
+        throw NewtonFailure(ex.error_code, "Steady state backward "
+                            "computation failed. ");
```
dweindl, line 238:
```suggestion
                            "computation failed.");
```

src/steadystateproblem.cpp
```diff
@@ -209,6 +196,69 @@ void SteadystateProblem::findSteadyStateBySimulation(Solver *solver,
...
+    }
+
+    /* Will need to write quadratures: resize */
+    xQB.reset();
+    return true;
+}
+
+void SteadystateProblem::computeSteadyStateQuadrature(NewtonSolver *newtonSolver,
+                                                      Model *model) {
+    /* compute the integral over the adjoint state xBintegral */
```
dweindl, line 228: There is no `xBintegral`?!

paulstapor:
I now included documentation on usage an math in the equilibration logic notebook and tried to address all other issues. Let's see what comes from sonarcloud. Beyond possible few code smells, this PR should be finished from my side...

paulstapor:
Currently, I can't reproduce the error form the testsuite locally (tried running matlab tests, involving some changes in the corresponding routines). Didn't find any segfault there...

paulstapor:
Can reproduce the segfault by locally building the tests. However, it doesn't occur when running in lldb debugger, only when called directly without debugger... :sob:

paulstapor:
Finally found something which I think might be causing the segfault (for *whatever* reason)...

python/tests/test_compare_conservation_laws_sbml.py
```diff
@@ -55,22 +68,33 @@ def generate_models():
     # load both models
     sys.path.insert(0, os.path.abspath(model_name))
     sys.path.insert(0, os.path.abspath(model_name_cl))
-    model_without_cl_module = importlib.import_module(model_name)
-    model_with_cl_module = importlib.import_module(model_name_cl)
+    model_without_cl_module = amici.import_model_module(model_name)
```
FFroehlich, line 71: 👍

src/rdata.cpp
```diff
@@ -404,8 +406,46 @@ void ReturnData::processBackwardProblem(ForwardProblem const &fwd,
...
+    }
+}
+
+void ReturnData::handleSx0Backward(const Model &model,
+                                   SteadystateProblem const &preeq,
+                                   AmiVector &xQB) const {
+    /* If preequilibration is run in adjoint mode, the scalar product of sx0
+       with its adjoint counterpart (see handleSx0Forward()) is not necessary:
+       the actual simulation is "extended" by the preequilibration time.
+       At initialization (at t=-inf), the adjoint state is in steady state (= 0)
```
FFroehlich, line 433: this only works if the the jacobian has full rank right? Otherwise the adjoint state could be any vector from the null space?

src/solver.cpp
```diff
@@ -525,6 +528,16 @@ void Solver::setSensitivityMethod(const SensitivityMethod sensi_meth) {
     this->sensi_meth = sensi_meth;
 }
 
+void Solver::setSensitivityMethodPreequilibration(const SensitivityMethod sensi_meth_preeq) {
+    if (rdata_mode == RDataReporting::residuals &&
```
FFroehlich, line 532: Duplicate, please add seperate funtion with `sensi_meth` as argument.

src/solver.cpp
```diff
@@ -525,6 +528,16 @@ void Solver::setSensitivityMethod(const SensitivityMethod sensi_meth) {
...
+        sensi_meth_preeq == SensitivityMethod::adjoint)
+        throw AmiException("Adjoint Sensitivity Analysis is not compatible with"
+                           " only reporting residuals!");
+    if (sensi_meth_preeq != this->sensi_meth_preeq)
```
FFroehlich, line 536: Is this check really necessary? I don't think `sensi_meth_preeq` affects any memory allocation in this `Solver` instance, but only influences what happens when the SteadyStateSolver is generated, right?

src/steadystateproblem.cpp
```diff
@@ -138,27 +129,30 @@ void SteadystateProblem::findSteadyState(Solver *solver,
 void SteadystateProblem::findSteadyStateByNewtonsMethod(NewtonSolver *newtonSolver,
                                                         Model *model,
                                                         bool newton_retry) {
+    int ind = newton_retry ? 2 : 0;
```
FFroehlich, line 132: 👍

src/steadystateproblem.cpp
```diff
@@ -79,40 +87,23 @@ void SteadystateProblem::workSteadyStateProblem(Solver *solver, Model *model,
 }
 
 void SteadystateProblem::workSteadyStateBackwardProblem(Solver *solver,
-                                                        Model *model) {
+                                                        Model *model,
+                                                        BackwardProblem *bwd) {
+    /* initialize and check if there is something to be done */
+    bool computeBackward = initializeBackwardProblem(solver, model, bwd);
```
FFroehlich, line 93:
```suggestion
    if (!initializeBackwardProblem(solver, model, bwd))
```

src/steadystateproblem.cpp
```diff
@@ -79,40 +87,23 @@ void SteadystateProblem::workSteadyStateProblem(Solver *solver, Model *model,
...
+    if (!computeBackward)
```
FFroehlich, line 94:
```suggestion
```

src/steadystateproblem.cpp
```diff
@@ -209,6 +203,69 @@ void SteadystateProblem::findSteadyStateBySimulation(Solver *solver,
...
+
+        /* If we have a backward problem, we're in preequilibration.
+           Hence, quantities like t, x, and xB must be set. */
+        solver->reInit(t, x, x);
+        solver->updateAndReinitStatesAndSensitivities(model);
+        xB.copy(bwd->getAdjointState());
+    }
+
+    /* Will need to write quadratures: resize */
+    xQB.reset();
```
FFroehlich, line 229: this doesn't resize, it just sets all entries to zero.

src/steadystateproblem.cpp
```diff
@@ -209,6 +203,69 @@ void SteadystateProblem::findSteadyStateBySimulation(Solver *solver,
...
+    return true;
+}
+
+void SteadystateProblem::computeSteadyStateQuadrature(NewtonSolver *newtonSolver,
+                                                      Model *model) {
+    /* compute the integral over the adjoint state xB */
```
FFroehlich, line 235: huh, this code looks like it solves a linear system and does not compute an integral?

FFroehlich: I suppose this computes the quadratures but does not require integration in the steadystate case?

src/steadystateproblem.cpp
```diff
@@ -231,6 +288,10 @@ void SteadystateProblem::writeErrorString(std::string *errorString,
                                           SteadyStateStatus status) const {
     /* write error message according to steady state status */
     switch (status) {
+        case SteadyStateStatus::failed_too_long_simulation:
+            (*errorString).append(": Simulation beyond t=1e100 without finding "
```
FFroehlich, line 292: don't explicitely mention time here or extract value from variable, high chance of going stale.

src/steadystateproblem.cpp
paulstapor: This is what makes it nice:
In steady state, this integral has an analytical solution, which is given by the solution of this linear system, which is substantially faster than carrying out integration. However, if the Jacobian is singular, we have to carry out integration -> This is exactly what's planned for my next pr.

FFroehlich: Thats what I assumed, would be great if you could give a brief documentation that it can be and why it can be computed analytically ;)

src/rdata.cpp
paulstapor: In principle, this will work always, not depending on the Jacobian.
What depends however on the Jacobian having full rank is the computation of `xQBpreeq`. The analytical solution is indeed *only* available for full-rank systems. But a solution based on numerical integration will always work (as long as passing xB from BackwardProblem to SteadyStateProblem is possible, which is currently not the case if solver states must be reinitialized).

matlab/tests/testModels.m
```diff
@@ -8,14 +8,18 @@ function testModels()
     warning('off','AMICI:mex:CVODES:CVode:TOO_MUCH_WORK')
 
     ignoredTests = {'/model_jakstat_adjoint/sensiadjointemptysensind', ...
-                    '/model_jakstat_adjoint/sensiforwardemptysensind'};
+                    '/model_jakstat_adjoint/sensiforwardemptysensind', ...
```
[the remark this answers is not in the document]
paulstapor: Would have to retry. This script is anyhow somewhat outdated...
When cleaning up these things and adapting the matlab to the python interface, I can see, whether they are still necessary and why. I can remove them for the moment, don't know, what was going wrong there precisely by heart...

src/steadystateproblem.cpp
paulstapor: Was an resize originally, just didn't update the comment. Thanks! ;)

paulstapor:
Really not getting how sonarcloud is computing the code coverage. Apart from exceptions, basically all of the new code is covered `python/tests/test_compare_conservation_laws_sbml.py::test_adjoint_pre_and_post_equilibration`...

paulstapor:
Okay, would be happy to merge this today. Next (basically last) PR is already being prepared...

src/solver.cpp
```diff
@@ -515,14 +516,25 @@ void Solver::applySensitivityTolerances() const {
...
+    if (sensi_meth != this->sensi_meth)
+        resetMutableMemory(nx(), nplist(), nquad());
+    sensitivityMethod(&this->sensi_meth, sensi_meth);
+}
+
+void Solver::setSensitivityMethodPreequilibration(const SensitivityMethod sensi_meth_preeq) {
+    sensitivityMethod(&this->sensi_meth_preeq, sensi_meth_preeq);
+}
+
+void Solver::sensitivityMethod(SensitivityMethod *member,
```
FFroehlich, line 531: For future reference: This implementation probably overstretches the goals of code reuse and can easily be a bit confusing. Generally, never pass the reference of a class member as argument to a function. Best solution here would be to check the value in this method and set the value outside of this function

paulstapor: Well, you told me to write something reusable for both sensitivityMethod setters... ;)
This was my honest best guess... I can rewrite that in the upcoming PR, if it's helpful. No problem...

FFroehlich: The primary purpose of avoiding duplication is maintainability and not so much about enacting the maximum amount of abstraction possible (that usually leads to complex unreadable code). De-duplicating the value checks make sense since those checks may change and you want to have them in a central place and not change them in multiple locations. De-duplicating variable assignments doesn't really improve maintainability, as you can't programmatically access class members in C++, and passing pointers to class members to class methods is confusing as it obfuscates the fact that those pointers point to class members.

PR #1145, opened 2020-07-02
Correctly parse piecewise functions in rules

lcontento:
If a piecewise function is present inside a SBML assignment rule, sometimes (probably depending on rule ordering) it may not be correctly parsed. Use of `_parse_special_functions` after all instances of reading rule formulas solves this problem.

As a note, in my project (which makes heavy use of nested rules) most of the time inside `_process_rules` is spent sympifying formulas or converting them to MathML in order to be written back into the SBML document. A possible idea to speed up things would be to convert all rules to SymPy formulas at the beginning and then operate only on the SymPy objects without touching the SBML anymore.

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -1030,7 +1030,7 @@ class SbmlImporter:
                 for nested_rule in rules:
 
                     nested_formula = sp.sympify(
-                        sbml.formulaToL3String(nested_rule.getMath()),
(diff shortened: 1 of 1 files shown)


dweindl:
> A possible idea to speed up things would be to convert all rules to SymPy formulas at the beginning and then operate only on the SymPy objects without touching the SBML anymore.

Sounds good to me. I would actually like to have that for all relevant SBML entities, not just rules. Quite some time is spent iterating those things in large models. This would then also be the place to collect all local symbols.

PR #1146, opened 2020-07-06
Adjoint equilibration by simulation

paulstapor:
If we have a singular Jacobian, we can't use the adjoint + steadystate funcitonality yet.
This PR would change this, by trying to generate the least possible computational overhead.


include/amici/abstract_model.h
```diff
@@ -68,6 +68,39 @@ class AbstractModel {
                         const AmiVector &dx, int ip, const AmiVector &sx,
                         const AmiVector &sdx, AmiVector &sxdot) = 0;
 
+    /**
```
paulstapor, line 71: As shortly mentioned: I'm not sure, whether implementing these additional functions into abstract_model is really what we want. I needed this, because backward steadystate integration operates on members of model (or rather: uses them, not really changes them). An alternative would be modifying the solver::setup(), in order to allow custom RHS and Jacobian functions to be passed to it.

include/amici/rdata.h
```diff
@@ -265,6 +265,12 @@ class ReturnData {
      */
     std::vector<int> preeq_numlinsteps;
 
+    /**
+     * number of simulation steps for adjoint steady state problem
+     * (preequilibration) [== 0 if analytical solution worked, > 0 otherwise]
+     */
+    int preeq_numstepsB = 0;
```
paulstapor, line 272: We get an additional readout here, actually maybe even two: The time point, when steadystate was reached in backward integration could also be outputted, or even the backward steadystate (however: no clue what this would be good for atm): For models with singular Jacobian, the steadystate of xB does not need to be 0, but can be just some vector in the left-kernel of J.

paulstapor: We could also think of doing something like a `backward_status`, since forward, we have a 3-vector for this...

include/amici/solver.h
```diff
@@ -100,11 +100,14 @@ class Solver {
...
      * @param sdx0 initial derivative state sensitivities
+     * @param steadystate (optional) flag indicating steadystate case
+     * @param xQ0 initial quadrature vector
      */
 
     void setup(realtype t0, Model *model, const AmiVector &x0,
                const AmiVector &dx0, const AmiVectorArray &sx0,
-               const AmiVectorArray &sdx0) const;
+               const AmiVectorArray &sdx0, bool steadystate = false,
+               const AmiVector &xQ0 = AmiVector(0)) const;
```
paulstapor, line 110: Somehow not happy to modify the setup routine here... But need it to inform the solver, that we want to do forward quadratures and use another RHS function... Any ideas on that? Could this be done better?

include/amici/solver.h
```diff
@@ -636,6 +640,13 @@ class Solver {
      */
     const AmiVector &getAdjointQuadrature(int which, realtype t) const;
 
+    /**
```
paulstapor, line 643: We get a whole lot of new functions, as now, CVode-Forward-Quadrature routines are interfaced. 
I wanted to do forward quadratures, since this stores no forward trajectory in contrast to the standard adjoint approach. In Steadystate, trajectories are constant, so we don't need to store them.
Anyone feeling uncomfy with that solution? Tried to go for the most computational efficient thing here...

include/amici/steadystateproblem.h
```diff
@@ -93,13 +93,32 @@ class SteadystateProblem {
...
 
     /**
-     * Compute the quadrature in steady state backward mode by solving the
+     * Handles the computation of the quadrature in steady state backward mode
+     *
+     * @param newtonSolver pointer to the newtonSolver solver object
+     * @param solver pointer to the solver object
+     * @param model pointer to the model object
+     */
+    void computeSteadyStateQuadrature(NewtonSolver *newtonSolver,
```
paulstapor, line 102: I split the old backward routine up to flatten out logic, similar to what was done in the forward part.

include/amici/steadystateproblem.h
```diff
@@ -143,22 +162,25 @@ class SteadystateProblem {
...
      * @param rtol relative tolerance
+     * @param ewt error weight vector
      * @return root-mean-square norm
      */
     realtype getWrmsNorm(AmiVector const &x,
                          AmiVector const &xdot,
                          realtype atol,
-                         realtype rtol);
+                         realtype rtol,
+                         AmiVector &ewt);
```
paulstapor, line 172: If we have quadratures backward, the ewt-vector will have another size. this is why we now need to pass it to getWrmsNorm

include/amici/steadystateproblem.h
```diff
@@ -204,6 +229,16 @@ class SteadystateProblem {
...
 
+    /**
+     * initialize backward computation by setting state, time, adjoint
+     * state and checking for preequilibration mode
+     *
+     * @param model pointer to the model object
+     * @param yQ vector to be multiplied with dxdotdp
+     * @param yQB resulting vector after multiplication
+     */
+    void getQBfromQ(Model *model, const AmiVector &yQ, AmiVector &yQB);
```
paulstapor, line 240: Is there some better name for this? Personally don't like it...

python/tests/test_compare_conservation_laws_sbml.py
```diff
@@ -88,6 +88,7 @@ def get_results(model, edata=None, sensi_order=0,
     # get the solver, set the properties
     solver = model.getSolver()
     solver.setNewtonMaxSteps(20)
+    solver.setMaxSteps(15000) # the model takes a while to equilibrate...
```
paulstapor, line 91: We don't need this anymore, have to take it out...

src/model_ode.cpp
```diff
@@ -394,6 +394,42 @@ void Model_ODE::fqBdot(realtype t, N_Vector x, N_Vector xB, N_Vector qBdot) {
     }
 }
 
+void Model_ODE::fxBdot_ss(const realtype t, const AmiVector &xB,
```
paulstapor, line 397: If anyone has a better idea for the function name than just appending `_ss`, feel free to suggest... ;)

src/solver.cpp
```diff
@@ -116,8 +117,15 @@ void Solver::setup(const realtype t0, Model *model, const AmiVector &x0,
 
             applyTolerancesFSA();
         } else if (sensi_meth == SensitivityMethod::adjoint) {
-            /* Allocate space for the adjoint computation */
-            adjInit();
+            if (steadystate) {
+                /* Allocate space for forward quadratures */
+                quadInit(xQ0);
+                applyQuadTolerances();
+                setSparseJacFn_ss();
```
paulstapor, line 124: Found it easier to overwrite the JacFcn, rather than also changing the intialization function of the linear solver. In the long run, this line should be replaced with a function, with passes the correct Jacobian function depending on the linear solver. So currently, this logic will **only** work with KLU, I'm afraid.

src/solver.cpp
```diff
@@ -521,20 +546,24 @@ SensitivityMethod Solver::getSensitivityMethodPreequilibration() const { return
...
-    sensitivityMethod(&this->sensi_meth, sensi_meth);
+    sensitivityMethod(sensi_meth, false);
 }
 
 void Solver::setSensitivityMethodPreequilibration(const SensitivityMethod sensi_meth_preeq) {
-    sensitivityMethod(&this->sensi_meth_preeq, sensi_meth_preeq);
+    sensitivityMethod(sensi_meth_preeq, true);
 }
 
-void Solver::sensitivityMethod(SensitivityMethod *member,
```
paulstapor, line 531: Hope this recoding is better now than the fully abstract version I implemented before.

src/solver_cvodes.cpp
```diff
@@ -976,6 +1035,60 @@ static int fqBdot(realtype t, N_Vector x, N_Vector xB, N_Vector qBdot,
...
+ * @param t timepoint
+ * @param xB Vector with the adjoint states
+ * @param xBdot Vector with the adjoint right hand side
+ * @param user_data object with user input @type Model_ODE
+ * @return status flag indicating successful execution
+ */
+static int fxBdot_ss(realtype t, N_Vector xB, N_Vector xBdot,
+                     void *user_data) {
+    auto model = static_cast<Model_ODE *>(user_data);
+    model->fxBdot_ss(t, xB, xBdot);
```
paulstapor, line 1050: Not sure how, but maybe there's another way to implement this, which does not operate on the model Jacobian... Opinions?

src/steadystateproblem.cpp
```diff
@@ -19,13 +19,14 @@
 namespace amici {
 
 SteadystateProblem::SteadystateProblem(const Solver &solver, const Model &model)
-    : delta(model.nx_solver), ewt(model.nx_solver),
+    : delta(model.nx_solver), ewt(model.nx_solver), ewtQB(model.nplist()),
```
paulstapor, line 22: As said: Need that to control errors in quadratures...

src/steadystateproblem.cpp
```diff
@@ -225,11 +226,15 @@ bool SteadystateProblem::initializeBackwardProblem(Solver *solver,
     }
 
     /* Will need to write quadratures: set to 0 */
+    xQ.reset();
     xQB.reset();
+    xQBdot.reset();
```
paulstapor, line 231: I originally wanted to avoid introducing this function... Turned out, we need it...

src/steadystateproblem.cpp
```diff
@@ -238,45 +243,71 @@ void SteadystateProblem::computeSteadyStateQuadrature(NewtonSolver *newtonSolver
...
-    } else {
-        for (int ip=0; ip<model->nplist(); ++ip)
-            xQB[ip] = N_VDotProd(xB.getNVector(),
-                                 model->dxdotdp.getNVector(ip));
+    /* set starting timepoint for the simulation solver */
+    t = 0;
+    /* xQ was written in getQuadratureByLinSolve() -> reset */
+    xQ.reset();
+    /* initialize the Jacobian */
+    model->writeSteadystateJB(t, 0, x, x, xB, xB, xB);
```
paulstapor, line 299: Here's where we reset the Model Jacobian... As said, not sure whether we maybe want to avoid it and go another way.

src/steadystateproblem.cpp
```diff
@@ -363,37 +411,40 @@ bool SteadystateProblem::getSensitivityFlag(const Model *model,
...
+bool SteadystateProblem::checkConvergence(const Solver *solver,
+                                          Model *model,
+                                          SensitivityMethod checkSensitivities) {
     /* get RHS and compute weighted error norm */
-    model->fxdot(t, x, dx, xdot);
-    wrms = getWrmsNorm(x, xdot, solver->getAbsoluteToleranceSteadyState(),
-                       solver->getRelativeToleranceSteadyState());
+    if (checkSensitivities == SensitivityMethod::adjoint) {
+        getQBfromQ(model, xQ, xQB);
+        getQBfromQ(model, xB, xQBdot);
```
paulstapor, line 429: Originally, I wanted to avoid computing these two thing in every step. I thought integrating xB to xQ and then multiplying with dxdotdp to get xQB is enough, because I hoped that xQBdot and xQdot (=xB) always become small together. However, xB can remain strictly positive, while xQBdot goes to zero. So we need to really check the whole xQBdot thing here...

src/steadystateproblem.cpp
```diff
@@ -527,10 +586,14 @@ void SteadystateProblem::getSteadystateSimulation(Solver *solver,
...
-        solver->writeSolution(&t, x, dx, sx);
+        if (backward) {
+            solver->writeSolution(&t, xB, dx, sx, xQ);
+        } else {
+            solver->writeSolution(&t, x, dx, sx, xQ);
+        }
 
         /* Check for convergence */
-        converged = checkConvergence(solver, model);
+        converged = checkConvergence(solver, model, sensitivityFlag);
```
paulstapor, line 596: I was thinking whether it might make sense to do this check for convergence only once in 10 steps, because I have the feeling that it may be time-consuming if done always...

paulstapor:
Travis is failing due to some error with homebrew. Beyond this (and codacy code smells), this code is functional, tested, and does what we want it to do, as far as I can tell. Would be happy about comments on it! :)

include/amici/steadystateproblem.h
FFroehlich: I think its fine as-is. maybe replace `get` with `compute` since its technically not a getter method.

src/steadystateproblem.cpp
```diff
@@ -363,37 +411,40 @@ bool SteadystateProblem::getSensitivityFlag(const Model *model,
...
-    /* If we also integrate forward sensis, we need to also check those:
-       Check if: sensis enabled && steadyStateSensiMode == simulation */
-    bool checkForwardSensis =
-        solver->getSensitivityOrder() > SensitivityOrder::none &&
-        model->getSteadyStateSensitivityMode() ==
-            SteadyStateSensitivityMode::simulationFSA;
-
     /* get RHS of sensitivities and compute weighted error norm */
-    if (checkForwardSensis) {
+    if (checkSensitivities == SensitivityMethod::forward) {
```
FFroehlich, line 440: Add a comment that `xdot` convergence needs to be checked for both `checkSensitivities == SensitivityMethod::forward` and `SensitivityOrder::none` and why it doesn't need to be checked for `checkSensitivities == SensitivityMethod::adjoint`.

Might be cleaner to have this in a switch statement, but no strong opinion.

include/amici/steadystateproblem.h
```diff
@@ -326,6 +367,8 @@ class SteadystateProblem {
     AmiVector delta;
     /** error weights */
```
FFroehlich, line 368: state this has dimension of states?

include/amici/steadystateproblem.h
```diff
@@ -326,6 +367,8 @@ class SteadystateProblem {
...
     AmiVector ewt;
+    /** error weights for backward quadratures */
```
FFroehlich, line 370: state this has dimensions of parameters/quadratures, and that both of them are only serve as temporary storage during wrms computation?

include/amici/steadystateproblem.h
FFroehlich: If you feel like its worthwhile mentioning this in the PR, it probably makes sense to have this as comment in the code. I added some suggestions where I feel like this should be documented.

include/amici/solver.h
FFroehlich: Looks fine to me.

src/forwardproblem.cpp
```diff
@@ -106,7 +106,7 @@ void ForwardProblem::workForwardProblem() {
             // Solve for nextTimepoint
             while (t < nextTimepoint) {
                 int status = solver->run(nextTimepoint);
-                solver->writeSolution(&t, x, dx, sx);
+                solver->writeSolution(&t, x, dx, sx, dx);
```
FFroehlich, line 109: passing dx as dummy not be the best idea since `xQ` gets assigned after `dx`

src/solver.cpp
```diff
@@ -68,7 +68,8 @@ void Solver::runB(const realtype tout) const {
 
 void Solver::setup(const realtype t0, Model *model, const AmiVector &x0,
                    const AmiVector &dx0, const AmiVectorArray &sx0,
-                   const AmiVectorArray &sdx0) const {
+                   const AmiVectorArray &sdx0, bool steadystate,
```
FFroehlich, line 71: would prefer to have steadystate as last argument in the signature (to have consistent types grouped).

src/solver.cpp
FFroehlich: if it only works with KLU, please throw an error if a different linear solver is set.

src/solver.cpp
FFroehlich: Definitely better, would have kept the assignment out of the function though and called the function `checkSensitivityMethod`. Should be overall less code, right?

include/amici/solver.h
FFroehlich: I see your point. Is is possible to write a function that creates a solver instance using the regular setup routine and then just attaches different functions where appropriate by calling all the init functions again?

src/steadystateproblem.cpp
FFroehlich: not sure what you mean by this comment.

include/amici/solver.h
FFroehlich: This would have the added benefit that the logic why these things are set the way they are can be documented in a central location.

FFroehlich:
looks like my comments were somehow split across and are now partly duplicated, sorry about that. Apparently doesn't like if the the review is done in multiple sessions.

include/amici/abstract_model.h
paulstapor: Good to know indeed. :+1:

src/solver.cpp
paulstapor: If the re-calling of init functions works, we won't need the change in signature any more. If this approach doesn't work, I will come back to this.

src/steadystateproblem.cpp
```diff
@@ -238,45 +243,71 @@ void SteadystateProblem::computeSteadyStateQuadrature(NewtonSolver *newtonSolver
...
-           xB(t) = exp( t * JB^T(x_ss, p) ) * xB(0)
-       This integral xBI is given as the solution of
-           JB^T(x_ss, p) * xBI = xB(0)
-       So we first try to solve the linear system, if possible. */
+    /* Try to compute the analytical solution for quadrature algebraically */
+    getQuadratureByLinSolve(newtonSolver, model);
+
+    /* Analytical solution didn't work, perform simulation instead */
+    if (!hasQuadrature())
+        getQuadratureBySimulation(solver, model);
```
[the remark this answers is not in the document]
paulstapor: No, we just call that once:
`workSteadyStateBackwardProblem()` calls `computeSteadyStateQuadrature()` calls `getQuadratureBySimulation()` calls `getSteadystateSimulation()`, and only the last one actually carries out integration. So there's only one.
Admittedly, all these functions are rather short and one could avoid maybe one level of nesting, but I preferred having it this way...

src/steadystateproblem.cpp
```diff
@@ -238,45 +243,71 @@ void SteadystateProblem::computeSteadyStateQuadrature(NewtonSolver *newtonSolver
...
-                                             xB.getNVector(), plist, true);
-        if (model->ndxdotdp_implicit > 0)
-            model->dxdotdp_implicit.multiply(xQB.getNVector(),
-                                             xB.getNVector(), plist, true);
-    } else {
-        for (int ip=0; ip<model->nplist(); ++ip)
-            xQB[ip] = N_VDotProd(xB.getNVector(),
-                                 model->dxdotdp.getNVector(ip));
+    /* set starting timepoint for the simulation solver */
+    t = 0;
```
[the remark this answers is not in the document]
paulstapor: Would be fine as well. I actually think that the precise time set has no influence on the result, so I just set it to 0, but maybe model->t0() is more consistent, so let's go with that.

src/steadystateproblem.cpp
paulstapor: Hm... May reduce the overall dependencies a bit, yes. However, atm this only works with KLU. In the near future, `writeSteadystateJB()` might include a switch depending on the linear solver... Not 100% sure yet, what's the best solution...
For the moment however, we can go with your solution, as it makes things look a little cleaner... and avoids exposing this function via abstract_model to the outside world, which is a good thing.

src/steadystateproblem.cpp
paulstapor: okay, true. This could be restructured...

src/steadystateproblem.cpp
paulstapor: Okay. I can at some point check this for the human benchmark model. Not a big deal to change 3 lines, recompile and monitor cpu time

include/amici/defines.h
```diff
@@ -152,7 +152,7 @@ enum class SteadyStateStatus {
 enum class SteadyStateContext {
     newton = 0,
     storage = 1,
-    integration = 2
+    creation = 2
```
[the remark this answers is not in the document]
paulstapor: Just to keep track for myself, what I've already addressed:
Done ✅

include/amici/solver.h
paulstapor: Seems not to be possible naively, as re-calling CVode/IDAInit leads to unwanted behavior (it unsets the linear solver module, for example).
Trying to set it in the solver memory directly now.

src/steadystateproblem.cpp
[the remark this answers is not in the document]
FFroehlich: You are right, the forward part is only called during `workSteadyStateBackwardProblem` and the backward part is `workSteadyStateProblem`, so thats fine. 

I think the nesting is absolutely fine, what I am more concerned about is the dual use of `getSteadystateSimulation`. Depending on whether `backward` is false or not, those routines do completely different things, even though they are using the same code. I think it would be cleaner to split the functionality in three functions: two functions that implement everything that is common specific to `backward=true` and `backward=false` and one function that handles the commonality between the two, i.e. the simulation + convergence checks.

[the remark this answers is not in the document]
FFroehlich: Ignore the previous comment, I think it doesnt actually fix much as there is no good way to work around the `solver->writeSolution` calls. I think the best thing to do here is add a comment in the code that there are two ways to end up in this function and that they serve two different purposes.

python/tests/test_compare_conservation_laws_sbml.py
```diff
@@ -200,12 +201,11 @@ def test_adjoint_pre_and_post_equilibration(edata_fixture):
...
 
-            # run simulations with different ways of preequilibration
-            # using a model with singular Jacobian
+    # --- compare fully adjoint approach to simulation with singular Jacobian ----
             raa = get_results(model, edata=edata, sensi_order=1,
                               sensi_meth=amici.SensitivityMethod.adjoint,
                               sensi_meth_preeq=amici.SensitivityMethod.adjoint,
                               reinitialize_states=reinit)
-                              
+
```
FFroehlich, line 209: what about combinations of pre and postequilibration?

src/steadystateproblem.cpp
```diff
@@ -425,28 +432,35 @@ bool SteadystateProblem::checkConvergence(const Solver *solver,
                                           SensitivityMethod checkSensitivities) {
     /* get RHS and compute weighted error norm */
     if (checkSensitivities == SensitivityMethod::adjoint) {
-        getQBfromQ(model, xQ, xQB);
-        getQBfromQ(model, xB, xQBdot);
+        /* In the adjoitncase, only xQB contributes to the gradient,
```
FFroehlich, line 435:
```suggestion
        /* In the adjoint case, only xQB contributes to the gradient,
```

src/steadystateproblem.cpp
```diff
@@ -425,28 +432,35 @@ bool SteadystateProblem::checkConvergence(const Solver *solver,
...
-                    solver->getRelativeToleranceSteadyStateSensi(), ewt);
-                converged = wrms < RCONST(1.0);
-            }
+    /* We're not in adjoint mode. Check convergence of xdot first */
+    model->fxdot(t, x, dx, xdot);
+    wrms = getWrmsNorm(x, xdot, solver->getAbsoluteToleranceSteadyState(),
+                       solver->getRelativeToleranceSteadyState(), ewt);
+
+    /* Do we need sensitivities? If not, we're done */
+    if (checkSensitivities == SensitivityMethod::none)
```
FFroehlich, line 451:
```suggestion
    if (checkSensitivities != SensitivityMethod::forward)
```
We don't need convergence of `sx` for adjoint sensis, right?

src/sundials_matrix_wrapper.cpp
```diff
@@ -174,6 +174,12 @@ void SUNMatrixWrapper::reset() {
         SUNMatZero(matrix);
 }
 
+void SUNMatrixWrapper::scale(realtype a) {
```
FFroehlich, line 177: 👍

include/amici/model_dae.h
```diff
@@ -232,6 +232,47 @@ class Model_DAE : public Model {
     void fqBdot(realtype t, N_Vector x, N_Vector dx, N_Vector xB, N_Vector dxB,
                 N_Vector qBdot);
 
+    void fxBdot_ss(const realtype t, const AmiVector &xB,
+                   const AmiVector &dxB, AmiVector &xBdot) override;
+
+    /** implementation of fxBdot for steady state case at the N_Vector level
```
dweindl, line 238:
```suggestion
    /**
      * @brief Implementation of fxBdot for steady state case at the N_Vector level
```

include/amici/model_dae.h
```diff
@@ -232,6 +232,47 @@ class Model_DAE : public Model {
...
+     * @param t timepoint
+     * @param xB Vector with the adjoint state
+     * @param dxB Vector with the adjoint derivative states
+     * @param xBdot Vector with the adjoint right hand side
+     */
+    void fxBdot_ss(realtype t, N_Vector xB, N_Vector dxB, N_Vector xBdot);
+
+    /** implementation of fqBdot for steady state case at the N_Vector level
```
dweindl, line 246:
```suggestion
    /** 
      * @brief Implementation of fqBdot for steady state case at the N_Vector level
```

include/amici/model_ode.h
```diff
@@ -130,7 +130,7 @@ class Model_ODE : public Model {
                    const AmiVector &dxB, const AmiVector &xBdot,
                    SUNMatrix JB) override;
 
-    /** implementation of fJSparseB at the N_Vector level, this function
+   /** implementation of fJSparseB at the N_Vector level, this function
```
dweindl, line 133:
```suggestion
   /** 
      * @brief Implementation of fJSparseB at the N_Vector level, this function
```

include/amici/model_ode.h
```diff
@@ -226,6 +226,45 @@ class Model_ODE : public Model {
      */
     void fqBdot(realtype t, N_Vector x, N_Vector xB, N_Vector qBdot);
 
+    void fxBdot_ss(const realtype t, const AmiVector &xB,
+                   const AmiVector & /*dxB*/, AmiVector &xBdot) override;
+
+    /** implementation of fxBdot for steady state case at the N_Vector level
```
dweindl, line 232:
```suggestion
    /**
       * @brief Implementation of fxBdot for steady state case at the N_Vector level
```

include/amici/model_ode.h
```diff
@@ -226,6 +226,45 @@ class Model_ODE : public Model {
...
+     * @param t timepoint
+     * @param xB Vector with the states
+     * @param xBdot Vector with the adjoint right hand side
+     */
+    void fxBdot_ss(realtype t, N_Vector xB, N_Vector xBdot);
+
+    /** implementation of fqBdot for steady state case at the N_Vector level
```
dweindl, line 239:
```suggestion
    /**
      * @brief Implementation of fqBdot for steady state case at the N_Vector level
```

include/amici/steadystateproblem.h
```diff
@@ -93,13 +93,32 @@ class SteadystateProblem {
                                      int it);
 
     /**
-     * Compute the quadrature in steady state backward mode by solving the
+     * Handles the computation of the quadrature in steady state backward mode
```
dweindl, line 96:
```suggestion
     * @brief Handles the computation of the quadrature in steady state backward mode
```

include/amici/steadystateproblem.h
```diff
@@ -93,13 +93,32 @@ class SteadystateProblem {
...
+     *
+     * @param newtonSolver pointer to the newtonSolver solver object
+     * @param solver pointer to the solver object
+     * @param model pointer to the model object
+     */
+    void computeSteadyStateQuadrature(NewtonSolver *newtonSolver,
+                                      const Solver *solver, Model *model);
+
+    /**
+     * Computes the quadrature in steady state backward mode by solving the
```
dweindl, line 106:
```suggestion
     * @brief Computes the quadrature in steady state backward mode by solving the
```

include/amici/steadystateproblem.h
```diff
@@ -93,13 +93,32 @@ class SteadystateProblem {
...
      * linear system defined by the backward Jacobian
      *
      * @param newtonSolver pointer to the newtonSolver solver object
      * @param model pointer to the model object
      */
-    void computeSteadyStateQuadrature(NewtonSolver *newtonSolver, Model *model);
+    void getQuadratureByLinSolve(NewtonSolver *newtonSolver, Model *model);
+
+    /**
+     * Computes the quadrature in steady state backward mode by numerical
```
dweindl, line 115:
```suggestion
     * @brief Computes the quadrature in steady state backward mode by numerical
```

include/amici/steadystateproblem.h
```diff
@@ -143,22 +162,25 @@ class SteadystateProblem {
...
      */
     realtype getWrmsNorm(AmiVector const &x,
                          AmiVector const &xdot,
                          realtype atol,
-                         realtype rtol);
+                         realtype rtol,
+                         AmiVector &ewt);
 
     /**
      * Checks convergence for state and respective sensitivities
```
dweindl, line 175:
```suggestion
     * @brief Checks convergence for state and respective sensitivities
```

include/amici/steadystateproblem.h
```diff
@@ -143,22 +162,25 @@ class SteadystateProblem {
...
+     * @param checkSensitivities flag whether sensitivities should be checked
      * @return boolean indicating convergence
      */
-    bool checkConvergence(const Solver *solver,
-                          Model *model);
+    bool checkConvergence(const Solver *solver, Model *model,
+                          SensitivityMethod checkSensitivities);
 
     /**
      * Runs the Newton solver iterations and checks for convergence to steady
```
dweindl, line 186:
```suggestion
     * @brief Runs the Newton solver iterations and checks for convergence to steady
```

include/amici/steadystateproblem.h
```diff
@@ -177,20 +199,23 @@ class SteadystateProblem {
      *
      * @param solver pointer to the solver object
      * @param model pointer to the model object
+     * @param backward flag indicating adjoint mode (including quadrature)
      */
-    void getSteadystateSimulation(Solver *solver, Model *model);
+    void runSteadystateSimulation(Solver *solver, Model *model, bool backward);
 
     /**
      * initialize CVodeSolver instance for preequilibration simulation
```
dweindl, line 207:
```suggestion
     * @brief Initialize CVodeSolver instance for preequilibration simulation
```

include/amici/steadystateproblem.h
```diff
@@ -177,20 +199,23 @@ class SteadystateProblem {
...
      * @return solver instance
      */
     std::unique_ptr<Solver> createSteadystateSimSolver(const Solver *solver,
                                                        Model *model,
-                                                       bool integrateForwardSensis) const;
+                                                       bool forwardSensis,
+                                                       bool backward) const;
 
     /**
      * initialize backward computation by setting state, time, adjoint
```
dweindl, line 221:
```suggestion
     * @brief Initialize backward computation by setting state, time, adjoint
```

include/amici/steadystateproblem.h
```diff
@@ -204,6 +229,16 @@ class SteadystateProblem {
     bool initializeBackwardProblem(Solver *solver, Model *model,
                                    const BackwardProblem *bwd);
 
+    /**
+     * compute the backward quadratures, which contribute to the gradient (xQB)
```
dweindl, line 233:
```suggestion
     * @brief Compute the backward quadratures, which contribute to the gradient (xQB)
```

include/amici/sundials_matrix_wrapper.h
```diff
@@ -145,6 +145,12 @@ class SUNMatrixWrapper {
      */
     void reset();
 
+    /**
+     * @brief multiply with a scalar
```
dweindl, line 149:
```suggestion
     * @brief multiply with a scalar (in-place)
```

src/sundials_matrix_wrapper.cpp
```diff
@@ -174,6 +174,12 @@ void SUNMatrixWrapper::reset() {
...
+    if (matrix)
+        for (int i = 0; i < static_cast<int>(nonzeros()); ++i)
```
dweindl, line 179: Move `static_cast<int>(nonzeros())` above loop to call only once and not for each element.

src/solver.cpp
paulstapor: This one is fixed now. Snudials doesn't offer a public API to only set the RHS method, but this can still be changed (and is changed now) in the Sundials Memory.
Usually, I'd be uncomfortable in changing stuff which is not publicly accessible. In this case however, the init function which sets the RHS function really does nothing with it but checking for NULL and setting, and beyond that the RHS is never touched again. So just changing it in the memory is safe enough, I think...

python/tests/test_compare_conservation_laws_sbml.py
paulstapor: This is covered by this test case when running with edata_full from the fixture.

paulstapor: If we're doing adjoint, we don't reach that point anyway but have returned before.

src/steadystateproblem.cpp
FFroehlich: > If we're doing adjoint, we don't reach that point anyway but have returned before.

I assume this belongs here. Right its technically correct, but the code is not very explicit about the intentions here. This is also illustrated by the fact that the current code will break if adjoint is used. I would propose the following implementation:

```
if (checkSensitivities == SensitivityMethod::adjoint) {
        computeQBfromQ(model, xQ, xQB);
        computeQBfromQ(model, xB, xQBdot);
        wrms = getWrmsNorm(xQB, xQBdot, solver->getAbsoluteToleranceQuadratures(),
                           solver->getRelativeToleranceQuadratures(), ewtQB);
} else {
        model->fxdot(t, x, dx, xdot);
        wrms = getWrmsNorm(x, xdot, solver->getAbsoluteToleranceSteadyState(),
                           solver->getRelativeToleranceSteadyState(), ewt);
}
bool converged = wrms < RCONST(1.0);

if (checkSensitivities != SensitivityMethod::forward)
        return converged;
```

Does that make sense?

paulstapor: It's logically equivalent, so totally fine to do that... Will implement (addressing the code review atm anyway)

FFroehlich: http://nemo.oudeis.fyi/posts/make-it-explicit/

To clarify why I make a fuss about this: Try writing down sentences for every line that says exactly what the code is supposed to be doing including all conditionals (if this becomes overly lengthy, its a good indication that the code is too convoluted). If your comment says

> /* Do we need sensitivities? If not, we're done */

but the code actually implements

> if (checkSensitivities == SensitivityMethod::none)

that raises a red flag with me.

paulstapor: > http://nemo.oudeis.fyi/posts/make-it-explicit/
> 
> To clarify why I make a fuss about this: Try writing down sentences for every line that says exactly what the code is supposed to be doing including all conditionals (if this becomes overly lengthy, its a good indication that the code is too convoluted). If your comment says
> 
> > /* Do we need sensitivities? If not, we're done */
> 
> but the code actually implements
> 
> > if (checkSensitivities == SensitivityMethod::none)
> 
> that raises a red flag with me.

No, that's totally okay to make a fuzz about it. I now start to actually see your point and not only your comment.
Maybe on a more general note: Are there tutorials about software development you can recommend?A substantial amount of our working time goes into this, but I have never attended a lecture or a course about this, it's all learning by doing. If there's an easy way to improve the situation (good free online course//tutorial) that's highly welcome...

FFroehlich: > > http://nemo.oudeis.fyi/posts/make-it-explicit/
> > To clarify why I make a fuss about this: Try writing down sentences for every line that says exactly what the code is supposed to be doing including all conditionals (if this becomes overly lengthy, its a good indication that the code is too convoluted). If your comment says
> > > /* Do we need sensitivities? If not, we're done */
> > 
> > 
> > but the code actually implements
> > > if (checkSensitivities == SensitivityMethod::none)
> > 
> > 
> > that raises a red flag with me.
> 
> No, that's totally okay to make a fuzz about it. I now start to actually see your point and not only your comment.
> Maybe on a more general note: Are there tutorials about software development you can recommend?A substantial amount of our working time goes into this, but I have never attended a lecture or a course about this, it's all learning by doing. If there's an easy way to improve the situation (good free online course//tutorial) that's highly welcome...

Unfortunately no. Usually everybody has their own opinion and it's often language specific. @dweindl might have some pointers.

FFroehlich: As you may see from the post, giving good advice is also not easy ;)

dweindl: software development is a pretty wide field. depends on what you are looking for specifically. maybe we discuss that elsewhere.

paulstapor:
Seems like tests are running again.
So, I would be happy to merge this thing as soon as possible now, it's been open for a while...
I could however 
1. still address some of the code smells 
2. include retrieving also the backward steadystate timepoint, if found by simulation.
Any opinions on this?

FFroehlich:
> Seems like tests are running again.
> So, I would be happy to merge this thing as soon as possible now, it's been open for a while...
> I could however
> 
> 1. still address some of the code smells
> 2. include retrieving also the backward steadystate timepoint, if found by simulation.

Wouldn't worry about 2) that right now, if somebody needs it, it can easily be added.

> 
> On behalf of 1.:
> How do we want to handle cvodes_solver and idas_solver directly operating on members of solver? Sonarcloud doesn't like this (which I can understand to some degree), but we're doing this at quite some spots in the code... I feel like introducing a protected API for cvodes_solver and idas_solver to operate on members of solver is beyond this PR...
> Any opinions on this?

I don't think this should be addressed here. I don't see a big benefit of refactoring `Solver`, so I am not super inclined to resolve most those code smells (at least those that complain about non-const protected Solver attributes).

All the shadowing complaints are something we generally have in AMICI, don't address this here.

The `void *` related issues, we can't do much since thats what Sundials needs. Similarly, I don't think we can mark them as `const` as Sundials will complain, but that's something that might actually work.

FFroehlich:
> Seems like tests are running again.

Doxygen still complains about a few things though ;)

dweindl:
>     1. still address some of the code smells

There are some valid remarks that should be addressed, in particular:

* > This function should be declared const.
* > unused parameter
* > Value stored to ... is never read

Some others are not addressable without major refactoring that is definitely out of scope here, and again others are not impossible to address without changing sundials as pointed out by Fabi. So don't get lost there...

paulstapor:
Done, except for the "value is never read".
This value, which is presumably never read, is in fact heavily used (and even written to) by sundials, so in the ThirdParty code...

dweindl:
> Done, except for the "value is never read".
> This value, which is presumably never read, is in fact heavily used (and even written to) by sundials, so in the ThirdParty code...

It is not. It is only locally overwritten. Sundials will not get anything from calling `fJSparseB_ss`. I guess you will notice a large step number there...

dweindl:
PS:

Note the difference between:
ODE:
```
void Model_ODE::fJSparseB_ss(SUNMatrix JB) {
    /* Just copy the model Jacobian */
    SUNMatCopy(J.get(), JB);
}
```

DAE:
```
void Model_DAE::fJSparseB_ss(SUNMatrix JB) {
    /* Just pass the model Jacobian on to JB */
    JB = J.get();
}
```

paulstapor:
> PS:
> 
> Note the difference between:
> ODE:
> 
> ```
> void Model_ODE::fJSparseB_ss(SUNMatrix JB) {
>     /* Just copy the model Jacobian */
>     SUNMatCopy(J.get(), JB);
> }
> ```
> 
> DAE:
> 
> ```
> void Model_DAE::fJSparseB_ss(SUNMatrix JB) {
>     /* Just pass the model Jacobian on to JB */
>     JB = J.get();
> }
> ```

Ah, true, yes, I confused this one with the ODE code! Thanks for pointing that out! 
I corrected it in the ODE code, but not in the DAE code!
Yes, fixed now, thanks a lot!

dweindl:
Just thinking: Although the number of integration steps varies from machine to machine, we may want to include it in our tests with maybe a 10% tolerance... Would have been helpful in a number of cases that were quite debugging-intensive.

paulstapor:
Very good point. This also holds true for the Newton Solver, where also a similar issue has come up at some point, when a minus sign had been switched...
I opened an issue for this (#1155).

paulstapor:
Just addressed the last code smell which there was (at least the last one which was realistically addressable).
Fine to merge?

And if so: New patch release soon?

dweindl:
> And if so: New patch release soon?

Definitely

PR #1152, opened 2020-07-14
MSVC compiler support and proper CI tests on Windows

thomassligon:
First time compiling openBLAS with MSVC on Travis.


scripts/installOpenBLAS.ps1
```diff
@@ -1,17 +1,31 @@
 Write-Host 'script installOpenBLAS.ps1 started'
-$VerbosePreference = "Continue" # display verbose messages
+# $VerbosePreference = "Continue" # display verbose messages
```
FFroehlich, line 2:
```suggestion
```

scripts/installOpenBLAS.ps1
```diff
@@ -1,17 +1,31 @@
...
 New-Item -Path 'C:\BLAS' -ItemType Directory -Force # create directory
 # Enforce stronger cryptography
 [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12
-$uri = 'https://sourceforge.net/projects/openblas/files/v0.3.6/OpenBLAS-0.3.6-x64.zip/download'
-$output = 'C:\BLAS\OpenBLAS-0.3.6-x64.zip'
+# $uri = 'https://sourceforge.net/projects/openblas/files/v0.3.6/OpenBLAS-0.3.6-x64.zip/download'
```
FFroehlich, line 6:
```suggestion

```

scripts/installOpenBLAS.ps1
```diff
@@ -1,17 +1,31 @@
...
+$output = 'C:\BLAS\v0.3.10.zip'
 # Invoke-WebRequest $uri -OutFile $output
 $webclient = New-Object System.Net.WebClient
 $webclient.DownloadFile($uri,"$output")
-Expand-Archive -Path 'C:\BLAS\OpenBLAS-0.3.6-x64.zip' -DestinationPath 'C:\BLAS\OpenBLAS-0.3.6-x64' -Force # expand zip file
-New-Item -Path Env:BLAS_LIBS -Value "/LIBPATH:C:\BLAS\OpenBLAS-0.3.6-x64\lib libopenblas.lib" -Force # create environment variable
-New-Item -Path Env:BLAS_CFLAGS -Value "/IC:\BLAS\OpenBLAS-0.3.6-x64\include" -Force # create environment variable
-Get-ChildItem 'C:\BLAS\OpenBLAS-0.3.6-x64' -Recurse # check for files
+Expand-Archive -Path 'C:\BLAS\v0.3.10.zip' -DestinationPath 'C:\BLAS\OpenBLAS-v0.3.10' -Force # expand zip file
+#Set-Location "C:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10" # change directory
```
FFroehlich, line 13:
```suggestion
```

scripts/installOpenBLAS.ps1
```diff
@@ -1,17 +1,31 @@
...
+C:\Users\travis\build\AMICI\scripts\compileBLAS.cmd
+New-Item -Path 'C:\BLAS\lib' -ItemType Directory -Force # create directory
+Copy-Item "C:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10\lib\Release\openblas.lib" -Destination "C:\BLAS\lib" -Recurse
+New-Item -Path 'C:\BLAS\bin' -ItemType Directory -Force # create directory
+Copy-Item "C:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10\lib\openblas.dll" -Destination "C:\BLAS\bin" -Recurse
+Get-ChildItem -Path "C:\BLAS" -Include "openblas.lib" -Recurse # check for file
+Get-ChildItem -Path "C:\BLAS" -Include "openblas.dll" -Recurse # check for file
+# New-Item -Path Env:BLAS_LIBS -Value "/LIBPATH:C:\BLAS\\lib libopenblas.lib" -Force # create environment variable
```
FFroehlich, line 21:
```suggestion
```

dweindl:
Regarding CI failure: 

http://www.swig.org/Release/README:
```
Troubleshooting
===============
In order to operate correctly, SWIG relies upon a set of library
files.  If after building SWIG, you get error messages like this,

    $ swig foo.i
    :1. Unable to find 'swig.swg'
    :3. Unable to find 'tcl8.swg'

it means that SWIG has either been incorrectly configured or
installed.  To fix this:

    1.  Make sure you remembered to do a 'make install' and that
        the installation actually worked.  Make sure you have
        write permission on the install directory.

    2.  If that doesn't work, type 'swig -swiglib' to find out
        where SWIG thinks its library is located.

    3.  If the location is not where you expect, perhaps
        you supplied a bad option to configure.  Use
        ./configure --prefix=pathname to set the SWIG install
        location.   Also, make sure you don't include a shell
        escape character such as ~ when you specify the path.

    4.  The SWIG library can be changed by setting the SWIG_LIB
        environment variable.  However, you really shouldn't
        have to do this.

If you are having other troubles, you might look at the SWIG Wiki at
https://github.com/swig/swig/wiki.
```

thomassligon:
My conclusion after this discussion was that I must have added the 2 test lines in the wrong place, and that provoked the swig error, so I just commented out the 2 lines. Now, I am still getting a swig error, and it seems to have something to do with doxygen. There may also be some overlap with #1169, so I checked to see where I have a copy of amici.py and the only one I found was in site-packages/amici, and it is from the last installation (pip install amici), so I could remove it by uninstalling amici, but that shouldn't have any effect on the Travis installation, since it is a completely different machine.
```
...
HDF5 library NOT found. Building AMICI WITHOUT HDF5 support.
running sdist
Testing SWIG executable swig4.0... FAILED.
Testing SWIG executable swig3.0... FAILED.
Testing SWIG executable swig... SUCCEEDED.
Found SWIG version (4, 0, 2)
Running SWIG: swig -doxygen -c++ -python -py3 -threads -Iamici/swig -Iamici/include -DAMICI_SWIG_WITHOUT_HDF5 -outdir C:\Users\travis\build\AMICI\python\sdist/amici -o amici/amici_wrap_without_hdf5.cxx amici/swig/amici.i
(1) : Error: Unable to find 'swig.swg'
(3) : Error: Unable to find 'python.swg'
amici\swig\amici.i(3) : Error: Unable to find 'exception.i'
amici\swig\amici.i(28) : Error: Unable to find 'stl.i'
The command "cd python/sdist && python setup.py sdist" failed and exited with 1 during .
Your build has been stopped.
Traceback (most recent call last):
  File "setup.py", line 251, in <module>
    main()
  File "setup.py", line 245, in main
    'Topic :: Scientific/Engineering :: Bio-Informatics',
  File "C:\Python37\lib\site-packages\setuptools\__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "C:\Python37\lib\distutils\core.py", line 148, in setup
    dist.run_commands()
  File "C:\Python37\lib\distutils\dist.py", line 966, in run_commands
    self.run_command(cmd)
  File "C:\Python37\lib\distutils\dist.py", line 985, in run_command
    cmd_obj.run()
  File "C:\Users\travis\build\AMICI\python\sdist\amici\custom_commands.py", line 264, in run
    self.run_swig()
  File "C:\Users\travis\build\AMICI\python\sdist\amici\custom_commands.py", line 277, in run_swig
    generate_swig_interface_files()
  File "C:\Users\travis\build\AMICI\python\sdist\amici\setuptools.py", line 257, in generate_swig_interface_files
    + sp.stdout.decode('utf-8'))
AssertionError: Swigging AMICI failed:
```

FFroehlich:
> My conclusion after this discussion was that I must have added the 2 test lines in the wrong place, and that provoked the swig error, so I just commented out the 2 lines. Now, I am still getting a swig error, and it seems to have something to do with doxygen. There may also be some overlap with #1169, so I checked to see where I have a copy of amici.py and the only one I found was in site-packages/amici, and it is from the last installation (pip install amici), so I could remove it by uninstalling amici, but that shouldn't have any effect on the Travis installation, since it is a completely different machine.
> 
> ```
> ...
> HDF5 library NOT found. Building AMICI WITHOUT HDF5 support.
> running sdist
> Testing SWIG executable swig4.0... FAILED.
> Testing SWIG executable swig3.0... FAILED.
> Testing SWIG executable swig... SUCCEEDED.
> Found SWIG version (4, 0, 2)
> Running SWIG: swig -doxygen -c++ -python -py3 -threads -Iamici/swig -Iamici/include -DAMICI_SWIG_WITHOUT_HDF5 -outdir C:\Users\travis\build\AMICI\python\sdist/amici -o amici/amici_wrap_without_hdf5.cxx amici/swig/amici.i
> (1) : Error: Unable to find 'swig.swg'
> (3) : Error: Unable to find 'python.swg'
> amici\swig\amici.i(3) : Error: Unable to find 'exception.i'
> amici\swig\amici.i(28) : Error: Unable to find 'stl.i'
> The command "cd python/sdist && python setup.py sdist" failed and exited with 1 during .
> Your build has been stopped.
> Traceback (most recent call last):
>   File "setup.py", line 251, in <module>
>     main()
>   File "setup.py", line 245, in main
>     'Topic :: Scientific/Engineering :: Bio-Informatics',
>   File "C:\Python37\lib\site-packages\setuptools\__init__.py", line 145, in setup
>     return distutils.core.setup(**attrs)
>   File "C:\Python37\lib\distutils\core.py", line 148, in setup
>     dist.run_commands()
>   File "C:\Python37\lib\distutils\dist.py", line 966, in run_commands
>     self.run_command(cmd)
>   File "C:\Python37\lib\distutils\dist.py", line 985, in run_command
>     cmd_obj.run()
>   File "C:\Users\travis\build\AMICI\python\sdist\amici\custom_commands.py", line 264, in run
>     self.run_swig()
>   File "C:\Users\travis\build\AMICI\python\sdist\amici\custom_commands.py", line 277, in run_swig
>     generate_swig_interface_files()
>   File "C:\Users\travis\build\AMICI\python\sdist\amici\setuptools.py", line 257, in generate_swig_interface_files
>     + sp.stdout.decode('utf-8'))
> AssertionError: Swigging AMICI failed:
> ```

No the swig errors are unrelated to the lines you added. The choco package was update to `swig 4.0.2` which leads to these errors. For guidance on how to resolve them, please see @dweindl message above.

FFroehlich:
I fixed the swig errors and reverted some of your changes that seem to have resulted from a bad merge. Not sure what happened there, but please try to address merge conflicts in a different way in the future.

From what I can tell, the travis run is currently still failing due to a missing DLL, not sure whats up with that.

```
ImportError while loading conftest 'C:\Users\travis\build\AMICI\python\tests\conftest.py'.
python\tests\conftest.py:8: in <module>
    import amici
C:\Python37\lib\site-packages\amici\__init__.py:135: in <module>
    from . import amici_without_hdf5 as amici
C:\Python37\lib\site-packages\amici\amici_without_hdf5.py:30: in <module>
    from . import _amici
ImportError: DLL load failed: The specified module could not be found.
```

thomassligon:
Hi Fabian, Daniel,
I'm sorry for the problems with merging. Of course I knew that first pull then push was the way to do it, but that still leads to a merge. I would really need to save the edits, the do pull, then edit, then push. When I get this merge scenario I get a merged file where I don't understand what things like >>>>HEAD etc. mean. I'll try to learn that before I allow another merge. Today, I just deleted my clone of AMICI and started fresh.

The problem with the missing DLL might just be a problem with the PATH, so I added the following statement:
`        - export PATH=${PATH}:C:\\BLAS\\bin`

If that doesn't fix it, I'll need to search through all dependencies and see what is missing.

Rant 1: On my first IT job, in 1978, I got a "file not found" message. Why can't the system just say which file was not found? Why does python import just say 
`ImportError: DLL load failed: The specified module could not be found.`
and refuse to tell me which module was not found? Why to I have to search through all dependencies until I find it? It's been 44 years and I still hate this message.

Rant 2: Unix/Linux has only one copy of each environment variable, but Windows has 3: Just changing an environment variable changes it in the current process and it is gone in the next process. Changing it in the machine or user does not take effect until the next logon. This is probably why we needed to set BLAS_LIBS and BLAS_CFLAGS in Git BASH instead of Powershell. But these are Windows environment variables, so it would be good to know how this works. Where are the effects of `Git BASH export` stored, and how do they get into Windows? Are the existing Windows environment variables overwritten or appended?

thomassligon:
Hello Fabian,
where do you see this Import Error? The only error I see is
`The command "python -m pytest --ignore-glob=*petab* ./python/tests" exited with 4.`
but I don't see why it is failing.

FFroehlich:
> Hello Fabian,
> where do you see this Import Error? The only error I see is
> `The command "python -m pytest --ignore-glob=*petab* ./python/tests" exited with 4.`
> but I don't see why it is failing.

Travis doesn't seem to show the error where they actually appear. It looks like the import errors are gone though, and currently the runs are failing as pysb is missing. Thats something I can fix.

thomassligon:
Here is a little remark on a warning you are getting. It is:
```
  C:\Python37\lib\site-packages\sympy\__init__.py:676: SymPyDeprecationWarning: 
  
  importing sympy.core.function with 'from sympy import *' has been
  deprecated since SymPy 1.6. Use import sympy.core.function instead.
  See https://github.com/sympy/sympy/issues/18245 for more info.
  
    deprecated_since_version="1.6").warn()
```
Here, `import *` is causing a problem because it imports something you are not using and causes the deprecation warning. You can avoid that by importing explicitly what is needed and nothing else, which is also faster.

dweindl:
That's afaik a pysb issue, don't worry about that for this pull request.

FFroehlich:
From what I can tell the errors that are currently occuring are due to AMICI not being able to compile the models it generated. I have temporarily set the model compilation to be always verbose, hopefully this will reveal what the underlying issue is.

Moreover, `test_cmake_compilation` is being run, which suggests that the environment variable `AMICI_SKIP_CMAKE_TESTS` is not set in a way that it is accesible from python in `os.environ`.

FFroehlich:
> That's afaik a pysb issue, don't worry about that for this pull request.

Correct, pysb tests will have to be disabled anyways since it would require installation of BNG, which is probably not worthwhile dvelving into. The warning could be fixed by installing from git@master though.

thomassligon:
I tried adding export to this statement
`export AMICI_SKIP_CMAKE_TESTS=TRUE`
In trying to understand BASH, I read
[https://stackoverflow.com/questions/234742/setting-environment-variables-in-linux-using-bash](https://stackoverflow.com/questions/234742/setting-environment-variables-in-linux-using-bash)
and it looks to me like you can write
`export AMICI_SKIP_CMAKE_TESTS=TRUE`
or
```
AMICI_SKIP_CMAKE_TESTS=TRUE
export AMICI_SKIP_CMAKE_TESTS
```

thomassligon:
It looks like the `export` helped.
`python/tests/test_misc.py::test_cmake_compilation SKIPPED                [  7%]`

The link command for `_test.cp37-win_amd64.pyd` cannot find `'build\temp.win-amd64-3.7\Release\swig\test_wrap.obj'`
```
3.7\Release\wrapfunctions.obj /OUT:C:\Users\travis\AppData\Local\Temp\tmpqsp40uon\test\_test.cp37-win_amd64.pyd /IMPLIB:build\temp.win-amd64-3.7\Release\swig\_test.cp37-win_amd64.lib /LIBPATH:C:\BLAS\lib openblas.lib
LINK : fatal error LNK1181: cannot open input file 'build\temp.win-amd64-3.7\Release\swig\test_wrap.obj'
error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.16.27023\\bin\\HostX86\\x64\\link.exe' failed with exit status 1181
```

I don't see where `test_wrap.obj` is created. There is
```
swigging swig/test.i to swig/test_wrap.cpp
C:\ProgramData\chocolatey\bin\swig.exe -python -c++ -modern -outdir test -IC:\Python37\lib\site-packages\amici\swig -IC:\Python37\lib\site-packages\amici\include -o swig/test_wrap.cpp swig/test.i
```
but I don't see any compilation from `test_wrap.cpp` to `test_wrap.obj`.

dweindl:
I'll never understand why setuptools/distutils shows only linker invocation but not compiler invocation.

`test_wrap.cpp` should be automatically added to the extension sources and compiled by setuptools. Not sure where this is going wrong. Either it is not compiled at all, or it ends up in a different directory...

thomassligon:
Here is a compiler invocation on Windows:

```
 C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\bin\HostX86\x64\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -Iamici/include -Iamici/ThirdParty/SuiteSparse/KLU/Include/ -Iamici/ThirdParty/SuiteSparse/AMD/Include/ -Iamici/ThirdParty/SuiteSparse/COLAMD/Include/ -Iamici/ThirdParty/SuiteSparse/BTF/Include/ -Iamici/ThirdParty/SuiteSparse/SuiteSparse_config -Iamici/ThirdParty/SuiteSparse/include -Iamici/ThirdParty/sundials/include -Iamici/ThirdParty/sundials/src -Iamici/ThirdParty/gsl/ "-IC:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\ATLMFC\include" "-IC:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\include" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\winrt" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.18362.0\cppwinrt" /EHsc /Tpamici\src\abstract_model.cpp /Fobuild\temp.win-amd64-3.7\amici\src\abstract_model.obj /IC:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10 /std:c++14
    abstract_model.cpp
```

This is probably not be relevant, but I noticed:

```
  Found link https://files.pythonhosted.org/packages/16/cc/9c323a1901727ce1ab606d0355baa21e737f4f6a70ad581d4f04512610ed/sympy-0.7.0.tar.gz#sha256=d204407cfdbeeed476243625d836fe751bf134ca97820b70cd9972ad7589a63e (from https://pypi.org/simple/sympy/), version: 0.7.0
  Skipping link: unsupported archive format: .exe: 
```
and the same for `.msi` and `.egg`

dweindl:
> Here is a compiler invocation on Windows:

Is there any `cl.exe` line containing `test_wrap.cpp`?

thomassligon:
No. Just to be sure, I also searched in the raw log. I found 10 occurrences of test_wrap, but nothing involving a compilation.

thomassligon:
For some reason, none of the test C modules are being compiled. That's why there are no calls to cl.exe in this part of the test. Now I am trying to debug this, and I can reproduce the Travis logic by doing some imports and calling
`python -m pytest --ignore-glob=*petab* --ignore-glob=*pysb* ./python/tests`
but I haven't succeeded in getting it to run in the debugger. For that; I would probably need to call a python script directly.

thomassligon:
Is there anything I can do to help?
Here is what I can see now:
There are 6 calls to swig for 5 `*_wrap` modules:
```
swigging swig/test_model_presimulation.i to swig/test_model_presimulation_wrap.cpp
swigging swig/test_model_steadystate_scaled.i to swig/test_model_steadystate_scaled_wrap.cpp
swigging swig/test_likelihoods.i to swig/test_likelihoods_wrap.cpp
swigging swig/model_constant_species_cl.i to swig/model_constant_species_cl_wrap.cpp
swigging swig/test.i to swig/test_wrap.cpp
swigging swig/test.i to swig/test_wrap.cpp
```
However, there are no calls to `cl.exe` for these modules. Then when the `*_wrap.obj` is used in a call to `link.exe`, it fails with 
`LNK1181: cannot open input file`
So the problem is: Why isn't the compiler called?
Also: Does the redundant swig for `test_wrap` cause a problem?

thomassligon:
Question: Which module(s) should compile `test_wrap.cpp` and the other test C files?
In order to compile Sundials and AMICI, the following modules are called:
```
AMICI/python/sdist/setup.py
AMICI/python/sdist/setup_clibs.py
```
There must be something analogous for the test C programs.

dweindl:
Building OpenBLAS can be optimized I guess. Currently compiling 2002 files. 

```
-- Building Single Precision

-- Building Double Precision

-- Building Complex Precision
```
We'll only use double. The multi-threaded version we can also skip.

.travis.yml
```diff
@@ -27,32 +27,25 @@ matrix:
...
         # as installation from sdist fails because of reasons...
         - choco install python --version 3.7.5
-        - choco install -y swig --version 4.0.1
+        - choco install -y swig --version=4.0.1
         - python -m pip install --upgrade pip
         - git clone -c core.symlinks=true https://github.com/ICB-DCM/AMICI.git && cd AMICI
         - if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then git checkout -qf $TRAVIS_COMMIT; elif [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then git fetch --update-head-ok origin pull/$TRAVIS_PULL_REQUEST/head:$TRAVIS_BRANCH && git checkout $TRAVIS_BRANCH; fi
         # run BLAS installation script
         - powershell -File 'C:\Users\travis\build\AMICI\scripts\installOpenBLAS.ps1';export BLAS_LIBS BLAS_CFLAGS
         # define Windows environment variables in BLAS because PowerShell definition didn't do the trick
```
dweindl, line 36: that means the following lines can be removed?

.travis.yml
```diff
@@ -27,32 +27,25 @@ matrix:
         # stick to python 3.7 until there is a 3.8 wheel for windows
         # as installation from sdist fails because of reasons...
         - choco install python --version 3.7.5
-        - choco install -y swig --version 4.0.1
+        - choco install -y swig --version=4.0.1
         - python -m pip install --upgrade pip
         - git clone -c core.symlinks=true https://github.com/ICB-DCM/AMICI.git && cd AMICI
         - if [[ "$TRAVIS_PULL_REQUEST" == "false" ]]; then git checkout -qf $TRAVIS_COMMIT; elif [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then git fetch --update-head-ok origin pull/$TRAVIS_PULL_REQUEST/head:$TRAVIS_BRANCH && git checkout $TRAVIS_BRANCH; fi
         # run BLAS installation script
         - powershell -File 'C:\Users\travis\build\AMICI\scripts\installOpenBLAS.ps1';export BLAS_LIBS BLAS_CFLAGS
```
dweindl, line 35: the `export` isn't doing anything, is it?

dweindl: in case it is actually doing something, please use `&&` instead of `;`

thomassligon:
Is/was the export doing anything? It shouldn't be necessary...testing.

dweindl:
> python/tests/test_compare_conservation_laws_sbml.py::test_compare_conservation_laws_sbml PASSED [  1%]

There we go...

dweindl:
> ========== 41 passed, 44 skipped, 187 warnings in 238.88s (0:03:58) ===========

:muscle: :muscle:  :muscle: 

This will still need some cleanup though.

@FFroehlich Do we need pysb tests here? For me it would be fine to skip. Or address separately. If it works on Linux/OSX and sbml import works on windows, then any issues are rather likely to lie outside of amici...

scripts/installOpenBLAS.ps1
```diff
@@ -1,17 +1,25 @@
...
+Copy-Item "C:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10\lib\openblas.dll" -Destination "C:\BLAS\bin" -Recurse
+Get-ChildItem -Path "C:\BLAS" -Include "openblas.lib" -Recurse # check for file
+Get-ChildItem -Path "C:\BLAS" -Include "openblas.dll" -Recurse # check for file
+[System.Environment]::SetEnvironmentVariable("BLAS_LIBS", "/LIBPATH:C:\BLAS\lib openblas.lib", [System.EnvironmentVariableTarget]::Machine)
+[System.Environment]::SetEnvironmentVariable("BLAS_LIBS", "/LIBPATH:C:\BLAS\lib openblas.lib", [System.EnvironmentVariableTarget]::User)
+[System.Environment]::SetEnvironmentVariable("BLAS_LIBS", "/LIBPATH:C:\BLAS\lib openblas.lib", [System.EnvironmentVariableTarget]::Process)
+[System.Environment]::SetEnvironmentVariable("BLAS_CFLAGS", "-IC:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10", [System.EnvironmentVariableTarget]::Machine)
+[System.Environment]::SetEnvironmentVariable("BLAS_CFLAGS", "-IC:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10", [System.EnvironmentVariableTarget]::User)
+[System.Environment]::SetEnvironmentVariable("BLAS_CFLAGS", "-IC:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10", [System.EnvironmentVariableTarget]::Process)
```
dweindl, line 22: These 6 lines are not required for travis. @thomassligon : will that store those variables permanently? I guess the the ::Machine lines work only as admin user?!

dweindl: If they have any value for a regular user, we can leave them in. Otherwise they should be removed.

dweindl:
Summary:

* Build OpenBLAS from source
* Cleanup old scripts
* Update docs (MSVC)
* C++: Define math constants as fallback
* Set correct C++ standard for model compilation on Windows
* No monkey patching of build_ext in the presence of swig files on Windows

thomassligon:
@dweindl I started to test my most recent understanding, but it conflicted with your work, so I will finish this a bit later.
A few months ago, we saw that the environment definition in the Powershell script was not working when compilation was done, but doing the definition in BASH, outside of Powershell, worked. Then, recently, I learned how to define the environment variables **persistently**. `Process` should work immediately, but disappear when the process is terminated. `User` should take effect when the user (the same user) logs on again, and `Machine` should take effect when any user logs on.
Does `Machine` require admin permissions? Good question, but I haven't seen an error message.
With that, only the `User` line should be necessary.
My plan is to comment out the BASH line and, if that works, comment out Process and Machine, then clean up.

dweindl:
> My plan is to comment out the BASH line and, if that works, comment out Process and Machine, then clean up.

As I wrote, on travis this is not required. No further testing required here. If you convince me that it is of any benefit for regular user installations, it can stay in, otherwise I'll remove it. This is best tested on your personal computer, not here.

> Does `Machine` require admin permissions? Good question, but I haven't seen an error message.

Note that here we do have admin privileges. Requires (local) testing as a non-admin user.

python/amici/custom_commands.py
```diff
@@ -358,17 +358,29 @@ def remove_swig_wrappers(extensions: 'setuptools.Extension',
     """Remove swig wrapper files not needed by the built extensions"""
 
     # remove swig-python-wrapper files
-    unused_swig_wrappers = {'amici/amici_wrap.cxx',
-                            'amici/amici_wrap_without_hdf5.cxx'}
+    unused_swig_wrappers = {
+        # they will also have '/' on windows!
```
FFroehlich, line 362: 😮

include/amici/defines.h
```diff
@@ -1,20 +1,61 @@
...
-#include <cmath>
+#if !defined(_USE_MATH_DEFINES)
+#define _USE_MATH_DEFINES
+#endif
+
 #include <functional>
 #include <string>
+#include <cmath>
+
+/* Math constants in case _USE_MATH_DEFINES is not supported */
```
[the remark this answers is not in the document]
dweindl: According to M$VC documentation, it should be supported, but here it didn't seem to work. Maybe because the first inclusion didn't have it defined. At least we have some fallback now...

python/amici/custom_commands.py
dweindl: Yeah, no clue..

scripts/installOpenBLAS.ps1
dweindl: Removing now. To be added later if relevant. Time to merge this monster.

FFroehlich:
> > ========== 41 passed, 44 skipped, 187 warnings in 238.88s (0:03:58) ===========
> 
> 💪 💪 💪
> 
> This will still need some cleanup though.
> 
> @FFroehlich Do we need pysb tests here? For me it would be fine to skip. Or address separately. If it works on Linux/OSX and sbml import works on windows, then any issues are rather likely to lie outside of amici...

sorry, skipping is fine

thomassligon:
@dweindl, @FFroehlich
Congratulations on getting all of this working! It wasn't easy.
I'll try to answer some questions on the BLAS environment variables (BLAS_CFLAGS and BLAS_LIBS) and point to one thing that might need to be updated.
Originally, I created these environment variables in the Powershell script, because it seemed more natural to do the Windows configuration as part of the Windows script. In addition to that, I have never seen any description of how Travis' `Git BASH` coordinates the results of `export`, running in `Git BASH`, with the environment variables on the Windows virtual machine. Are they immediately updated, or overwritten, and when? How does this compare with configuring Linux? And, when are new Windows processes started?
Now, it is clear that we don't need both, creating the variables inside the PS script and outside of it in BASH/Yaml. Cleanup should include removing everything not needed or not used, so I certainly agree with that.
One thing that may be left hanging: Installing openBLAS locally. This is necessary in order for `pip install amici` to work on Windows with MSVC, and it needs to include defining the environment variables. For that reason, I think it is easier to keep them in the PS script and document running the script locally (as I did in my first draft of an installation procedure, which I attached to some part of this discussion recently, but I don't remember exactly where, so here it is again).
[AMICI-Installation.txt](https://github.com/ICB-DCM/AMICI/files/5037607/AMICI-Installation.txt)

PR #1153, opened 2020-07-14
Reduce memory consumption for storing simulation states

### LINKED ISSUE #1150: Improper handling of replicate measurements in ForwardProblem::workForwardProblem / high memory usage
`ForwardProblem::workForwardProblem calls `handleDataPoint` for each time point, which saves the full `SimulationState` on each invocation. For replicate measurements this creates potentially enormous overhead.

paulstapor:
Closes #1150 
The simulation states are no longer stored as a vector but as a map, which avoids storing of replicates and reduces memory consumption, if many timepoint replicates are used and sx is large.
Moreover, the initial state is stored only once, if model->t0() is also a simulation time point.

### DIFF
--- a/include/amici/forwardproblem.h
+++ b/include/amici/forwardproblem.h
@@ -239,7 +239,11 @@ class ForwardProblem {
      * @return state
      */
     const SimulationState &getSimulationStateTimepoint(int it) const {
-        return timepoint_states.find(timepoints.at(it))->second;
+        if (timepoints.at(it) == initial_state.t) {
+            return getInitialSimulationState;
+        } else {
+            return timepoint_states.find(timepoints.at(it))->second;
+        }
     };
 
     /**
--- a/include/amici/forwardproblem.h
+++ b/include/amici/forwardproblem.h
@@ -395,9 +395,6 @@ class ForwardProblem {
     /** current time */
     realtype t;
 
-    /** number of timepoints */
-    std::vector<realtype> timepoints;
-
     /**
      * @brief Array of flags indicating which root has beend found.
      *
--- a/include/amici/forwardproblem.h
+++ b/include/amici/forwardproblem.h
@@ -239,7 +239,7 @@ class ForwardProblem {
      * @return state
      */
     const SimulationState &getSimulationStateTimepoint(int it) const {
-        return timepoint_states[timepoints.at(it)];
+        return timepoint_states.find(timepoints.at(it))->second;
     };
 
     /**
@@ -406,7 +406,7 @@ class ForwardProblem {
     std::map<realtype, SimulationState> timepoint_states;
 
     /** simulation state history at events*/
-    std::map<SimulationState> event_states;
+    std::vector<SimulationState> event_states;
 
     /** simulation state after initialization*/
     SimulationState initial_state;
--- a/include/amici/forwardproblem.h
+++ b/include/amici/forwardproblem.h
@@ -212,8 +212,8 @@ class ForwardProblem {
      * @brief Returns maximal time point index for which simulations are available
      * @return index
      */
-    int getTimepointCounter() const {
-        return static_cast<int>(timepoints.size() - 1);
+    realtype getFinalTime() const {
+        return final_state.t;
     }
 
     /**
@@ -239,10 +239,10 @@ class ForwardProblem {
      * @return state
      */
     const SimulationState &getSimulationStateTimepoint(int it) const {
-        if (timepoints.at(it) == initial_state.t) {
+        if (model->getTimepoint(it) == initial_state.t) {
             return getInitialSimulationState();
         } else {
-            return timepoint_states.find(timepoints.at(it))->second;
+            return timepoint_states.find(model->getTimepoint(it))->second;
         }
     };
 
--- a/include/amici/forwardproblem.h
+++ b/include/amici/forwardproblem.h
@@ -209,8 +209,8 @@ class ForwardProblem {
     }
 
     /**
-     * @brief Returns maximal time point index for which simulations are available
-     * @return index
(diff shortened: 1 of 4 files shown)


paulstapor:
Just to understand whether this directly opens a similar issue for events:
We also have a `std::vector<SimulationState> event_states`, but my gut  feeling tells me that a similar transformation as for `timepoint_states` will not be applicable, due to the fact that event tend to alter the simulation state. So if multiple events occur at the same timepoint, we really want to store the different simulation states, right?

include/amici/forwardproblem.h
```diff
@@ -391,6 +395,9 @@ class ForwardProblem {
     /** current time */
     realtype t;
 
+    /** number of timepoints */
+    std::vector<realtype> timepoints;
```
dweindl, line 399: Why duplicate here and not just use `model->getTimepoint(it)`?

src/forwardproblem.cpp
```diff
@@ -288,8 +288,12 @@ void ForwardProblem::storeEvent() {
 }
 
 void ForwardProblem::handleDataPoint(int it) {
-    timepoint_states.push_back(getSimulationState());
+    timepoints.push_back(getTime());
     solver->storeDiagnosis();
+    /* We only store the simulation state if it's not the initial state, as the
+       initial state is stored anyway and we want to avoid storing it twice */
+    if (t != model->t0())
+        timepoint_states.insert(std::make_pair(t, getSimulationState()));
```
dweindl, line 296:
```suggestion
        timepoint_states[t] = getSimulationState();
```
simpler? :)

paulstapor: Simpler, but `insert` checks for existence of that key, whereas `[t]` overwrites the entry if existing. SO I would prefer to go with `insert`, as this avoids writing the same content multiple times, if fine. :)

include/amici/forwardproblem.h
paulstapor: Was also my first idea, but timepoints used in ForwardProblem are not necessarily the same as in `model`, as ForwardProblem does not contain those with `t = inf`. As I was not 100% sure of the behavior with respect to the code part in
https://github.com/ICB-DCM/AMICI/blob/b10a8a3514eb1ff3ed727b9c2831e45a310986f1/src/rdata.cpp#L259
I decided to go for something which is logically fully equivalent. But if there's another way to handle the lines in rdata, I'm fine to go with `model->getTimepoints`... :)

src/forwardproblem.cpp
dweindl: Good point. But then we would want to skip creation of a new SimulationState instance (i.e. not calling getSimulationState()) which will still happen. Overriding it in the map is way cheaper than that...

paulstapor:
I don't find the "rerun"-button for the errored Travis-CI-Check...

dweindl:
> I don't find the "rerun"-button for the errored Travis-CI-Check...

Did you log in?

paulstapor:
> > I don't find the "rerun"-button for the errored Travis-CI-Check...
> 
> Did you log in?

Ah, dammit. Logged in into github, but not Travis. Rerunning seems to work via github in case of failure. In case of error, you need to log in into Travis as well. ;)

FFroehlich:
> Just to understand whether this directly opens a similar issue for events:
> We also have a `std::vector<SimulationState> event_states`, but my gut feeling tells me that a similar transformation as for `timepoint_states` will not be applicable, due to the fact that event tend to alter the simulation state. So if multiple events occur at the same timepoint, we really want to store the different simulation states, right?



> Just to understand whether this directly opens a similar issue for events:
> We also have a `std::vector<SimulationState> event_states`, but my gut feeling tells me that a similar transformation as for `timepoint_states` will not be applicable, due to the fact that event tend to alter the simulation state. So if multiple events occur at the same timepoint, we really want to store the different simulation states, right?

Not entirely sure. I think for nested events, only the sensitivity for the last timepoint is relevant, but we might be reporting event outputs based on previous states (current implementation might also be unintentional in this regard). Would keep as-is for now.

include/amici/forwardproblem.h
FFroehlich: That implementation wasn't very neat to begin. Probably makes more sense to extract maximum simulation time from forward problem and invalidate for all `model->getTimepoints` after that.

paulstapor: Okay, I can change that accordingly. :+1:

include/amici/forwardproblem.h
```diff
@@ -212,8 +212,8 @@ class ForwardProblem {
      * @brief Returns maximal time point index for which simulations are available
```
dweindl, line 212: Needs updating

include/amici/forwardproblem.h
```diff
@@ -239,7 +239,11 @@ class ForwardProblem {
      * @return state
      */
     const SimulationState &getSimulationStateTimepoint(int it) const {
-        return timepoint_states.at(it);
+        if (model->getTimepoint(it) == initial_state.t) {
+            return getInitialSimulationState();
+        } else {
```
dweindl, line 244: Remove `else` after `return`

src/forwardproblem.cpp
```diff
@@ -288,7 +288,11 @@ void ForwardProblem::storeEvent() {
 }
 
 void ForwardProblem::handleDataPoint(int it) {
-    timepoint_states.push_back(getSimulationState());
+    /* We only store the simulation state if it's not the initial state, as the
+       initial state is stored anyway and we want to avoid storing it twice */
+    if (t != model->t0() && timepoint_states.count(t) == 0)
```
dweindl, line 293: :+1:

PR #1160, opened 2020-07-17
Cleanup CI

FFroehlich:
Move more functionality to GitHub Actions, fix coverage report


.github/workflows/deploy_gh_pages.yml
```diff
@@ -0,0 +1,41 @@
+name: Doxygen
```
dweindl, line 1:
```suggestion
name: Doxygen / gh-pages
```

.github/workflows/test-petab-test-suite.yml
```diff
@@ -28,35 +28,33 @@ jobs:
...
-    - name: Create AMICI sdist
+    - name: Install python package
       run: |
-        cd python/sdist \
-          && check_time.sh create_sdist /usr/bin/python3 setup.py sdist
-    - name: Install AMICI sdist
-      run: |
-        AMICI_PARALLEL_COMPILE=2 check_time.sh \
-            install_sdist pip3 install -v --user \
-            $(ls -t python/sdist/dist/amici-*.tar.gz | head -1)[petab,pysb]
```
dweindl, line 50: Do we still have the installation from sdist somewhere? Should be tested somewhere.

.github/workflows/test_python_ver_matrix.yml
```diff
@@ -0,0 +1,65 @@
...
+          && pip3 install --upgrade --user setuptools
+          
+
+    # install AMICI
+    - name: Build BNGL
+      run: |
+        scripts/buildBNGL.sh
+    - name: Install python package
+      run: |
+        scripts/installAmiciSource.sh
```
dweindl, line 54: Maybe use sdist here, as that is the way how most users will use it?

include/amici/vector.h
```diff
@@ -61,6 +61,10 @@ class AmiVector {
             N_VMake_Serial(static_cast<long int>(vold.vec_.size()), vec_.data());
     }
 
+    /**
+     * @brief move constructor
+     * @param other vector from which the data will be moved
```
dweindl, line 66: Does doxygen complain about those? It doesn't really add value to document the obvious...

.github/workflows/test_python_ver_matrix.yml
FFroehlich: Had that initially, but having notebooks run outside a virtualenv was a major pain

include/amici/vector.h
FFroehlich: yep

.github/workflows/test-petab-test-suite.yml
FFroehlich: still in use in test-large-model.yaml

PR #1175, opened 2020-07-31
 PySB-based PEtab problems

dweindl:
PEtab so far supports only SBML models. As PySB model import is significantly faster, it is convenient to use those models directly without exporting to SBML first, but still being able to use all PEtab features. Therefore, here comes support for PySB-based PEtab problems in AMICI.

Restrictions:
* only Gaussian noise supported so far (#1176)
* no species or compartments allowed in the condition table (but can be handled via model parameters)

Also fixes:
* `amici.import_model_module` import of different models with the same name
* custom sigma expressions in `amici.pysb_import`

### DIFF
--- a/python/amici/petab_import_pysb.py
+++ b/python/amici/petab_import_pysb.py
@@ -1,0 +1,309 @@
+"""
+PySB-PEtab Import
+------------
+Import a model in the PySB-adapted :mod:`petab`
+(https://github.com/PEtab-dev/PEtab) format into AMICI.
+"""
+
+import logging
+import os
+from typing import List, Dict, Union, Optional, Tuple, Iterable
+
+import libsbml
+import petab
+import sympy as sp
+from petab.C import *
+
+from .logging import get_logger, log_execution_time, set_log_level
+from . import petab_import
+
+logger = get_logger(__name__, logging.WARNING)
+
+
+class PysbPetabProblem(petab.Problem):
+    """Representation of a PySB-model-based PEtab problem
+
+    This class extends :class:`petab.Problem` with a PySB model.
+    The model is augmented with the observation model based on the PEtab
+    observable table.
+    For now, a dummy SBML model is created which allows used the existing
+    SBML-PEtab API.
+
+    :ivar pysb_model:
+        PySB model instance from of this PEtab problem.
+
+    """
+
+    def __init__(self, pysb_model: 'pysb.Model' = None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.pysb_model: 'pysb.Model' = pysb_model
+        import pysb
+
+        # add any required output parameters
+        # sp.Symbol.__str__(comp)
+        locals = {sp.Symbol.__str__(comp): comp for comp in
+                  pysb_model.components if
+                  isinstance(comp, sp.Symbol)}
+        for formula in [*self.observable_df[petab.OBSERVABLE_FORMULA],
+                       *self.observable_df[petab.NOISE_FORMULA]]:
+            sym = sp.sympify(formula, locals=locals)
+            for s in sym.free_symbols:
+                if not isinstance(s, pysb.Component):
+                    p = pysb.Parameter(str(s), 1.0)
+                    locals[sp.Symbol.__str__(p)] = p
+
+
+        # add observables and sigmas to pysb model
+        for (observable_id, observable_formula, noise_formula) \
+                in zip(self.observable_df.index,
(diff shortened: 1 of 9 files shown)


python/amici/pysb_import.py
```diff
@@ -338,9 +339,9 @@ def _get_sigma_name_and_value(
     if obs_name in sigmas:
         if sigmas[obs_name] not in pysb_model.expressions:
             raise ValueError(f'value of sigma {obs_name} is not a '
-                            f'valid expression.')
-        sigma_name = pysb_model.expressions[sigmas[obs_name]].name
-        sigma_value = pysb_model.expressions[sigmas[obs_name]].expand_expr()
+                             f'valid expression.')
+        sigma_name = pysb_model.expressions[sigmas[obs_name].name].name
```
dweindl, line 343: Cleanup after clarifying intended usage

python/amici/petab_import_pysb.py
```diff
@@ -0,0 +1,333 @@
...
+
+        if self.pysb_model is not None:
+            self.sbml_document, self.sbml_model = \
+                create_dummy_sbml(
+                    self.pysb_model,
+                    observable_ids=self.observable_df.index.values
+                    if self.observable_df is not None else None
+                )
+
+    def _add_observation_model(self):
```
FFroehlich, line 59: Wouldn't this functionality ideally be provided by the PEtab library such that it could be reused by other tools?

python/amici/petab_import_pysb.py
```diff
@@ -0,0 +1,333 @@
...
+            visualization_files: PEtab visualization tables
+            observable_files: PEtab observables tables
+            pysb_model_file: PySB model file
+        """
+
+        sbml_model = sbml_document = sbml_reader = None
+        condition_df = measurement_df = parameter_df = visualization_df = None
+        observable_df = None
+
+        if condition_file:
```
FFroehlich, line 128: Shouldn't these file imports be redundant between sbml and pysb import?

python/amici/petab_import_pysb.py
```diff
@@ -0,0 +1,333 @@
...
+
+        if measurement_file:
+            # If there are multiple tables, we will merge them
+            measurement_df = petab.core.concat_tables(
+                measurement_file, petab.measurements.get_measurement_df)
+
+        if parameter_file:
+            parameter_df = petab.parameters.get_parameter_df(parameter_file)
+
+        if sbml_file:
```
FFroehlich, line 139: I thought the SBML dummy is created in `PysbPetabProblem`?

python/amici/petab_import_pysb.py
dweindl: I find it hard to judge how reusable that is for other tools. In general, it could be added there, yes. Would maybe keep it here for now until this has been tested a bit more.

python/amici/petab_import_pysb.py
dweindl: There is some redundancy yes. Would need some refactoring in PEtab (and a new release) first. Therefore, rather later...

python/amici/petab_import_pysb.py
```diff
@@ -0,0 +1,333 @@
...
+                )
+
+    def _add_observation_model(self):
+        """Extend PySB model by observation model as defined in the PEtab
+        observables table"""
+
+        import pysb
+
+        # add any required output parameters
+        locals = {sp.Symbol.__str__(comp): comp for comp in
```
FFroehlich, line 66: `locals` shadows builtin

python/amici/petab_import_pysb.py
dweindl: Right. Cleanup up a bit.

python/amici/petab_import_pysb.py
dweindl: Right, thx.

python/amici/pysb_import.py
dweindl: @FFroehlich : Here things didn't work as documented. Some lines assumed sigmas were provided as `pysb.Expression`, others as `pysb.Expression.name` (`str`). Documentation claims the latter. Seems it isn't tested anywhere. I will then adapt to using `str`?!

FFroehlich: Yes, I think its more user-friendly to require names (`str`) here.

dweindl: Agreed. Done.

PR #1188, opened 2020-08-06
fix ci yamls

FFroehlich:


.github/workflows/test_valgrind.yml
```diff
@@ -4,6 +4,10 @@ on:
     branches:
       - develop
       - master
+  pull_request:
+    branches:
+      - develop
```
dweindl, line 9: I think we removed it before for merges to develop because it was taking rather long. But also fine to keep.

PR #1189, opened 2020-08-06
refactor CI, fix pypi deploy

FFroehlich:


.github/workflows/deploy_release.yml
```diff
@@ -53,24 +53,16 @@ jobs:
     - name: apt
       run: |
         sudo apt-get update \
-          && sudo apt-get install -y swig3.0
-    - name: pip
+          && sudo apt-get install -y swig
```
dweindl, line 56: Do we have Ubuntu20 meanwhile? On the old one, "swig" package was version 1 or 2, I think

.github/workflows/test_python_cplusplus.yml
```diff
@@ -149,25 +137,9 @@ jobs:
...
-        scripts/buildCpputest.sh
-
-    - name: Build BNG
-      run: |
-        scripts/buildBNGL.sh
-
     - name: Build AMICI
       run: |
-        scripts/buildAmici.sh
+        scripts/buildAll.sh
```
dweindl, line 142: I actually prefer the separate steps. Comes with some redundancies, but at least it's super easy to see which steps succeeded without scrolling through walls of text :grimacing:

dweindl: Can we somehow induce text folding from within a script?

dweindl: https://github.community/t/has-github-action-somthing-like-travis-fold/16841 apparently

FFroehlich: Okay split `buildAmici` and `buildDependencies`, ThirdParty stuff should be pretty stable by now, right?

dweindl: Sounds good.

.github/workflows/deploy_release.yml
FFroehlich: Oh didn't realise that, non of the other scripts install `swig3.0`. 

> swig is already the newest version (3.0.12-1).

seems fine, no/

.github/workflows/test_python_cplusplus.yml
FFroehlich: Actually, in case of building problems, the test_install is probably the first one to fail, so lets just split up that one and keep the compact version elsewhere.

.github/workflows/deploy_release.yml
dweindl: Oh, seems that got fixed... smart move... didn't check in quite a while...

.github/workflows/test_python_cplusplus.yml
dweindl: ok

PR #1193, opened 2020-08-07
Update URLs after move to new github organization AMICI-dev

dweindl:


dweindl:
CI links still to do. Not sure which ones will change exactly...

PR #1199, opened 2020-08-12
Bugfix for piecewise functions

lcontento:
* If a piecewise function has conditions using logical operators and is found inside a SBML assignment rule, then logical operators must be parsed before each sympification.
* Makes MathML printing of `sympy.Float` round-trippable (can be removed if https://github.com/sympy/sympy/pull/19958 is accepted). However, libSBML will truncate floats anyway, so the Sympy -> SBML -> Sympy round-trip cannot be achieved...

### DIFF
--- a/.github/workflows/deploy_branch.yml
+++ b/.github/workflows/deploy_branch.yml
@@ -7,25 +7,6 @@ on:
         - develop
         
 jobs:
-  dockerhub:
-    # https://github.com/marketplace/actions/publish-docker
-    name: Deploy Dockerhub
-
-    runs-on: ubuntu-latest
-
-    steps:
-    - uses: actions/checkout@master
-    - run: git archive -o docker/amici.tar.gz --format=tar.gz HEAD
-    - name: Publish to Registry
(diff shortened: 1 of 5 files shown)


FFroehlich:
Not sure whats wrong here, need to investigate why tests aren't starting as expected.

FFroehlich:
Okay looks like dockerhub deployment doesn't work from forks as secrets are not available. Kinda makes sense. @dweindl How do you feel about only pushing to dockerhub on release?

FFroehlich:
> Okay looks like dockerhub deployment doesn't work from forks as secrets are not available. Kinda makes sense. @dweindl How do you feel about only pushing to dockerhub on master/develop?

dweindl:
> Okay looks like dockerhub deployment doesn't work from forks as secrets are not available. Kinda makes sense. @dweindl How do you feel about only pushing to dockerhub on master/develop?

Works for me. Not sure if even used by anybody...
If needed on other branches in the future, this can probably be set up to ignore forks.

PR #1208, opened 2020-08-19
Update documentation for Python interface

### LINKED ISSUE #654: Directly input ODE systems
The idea is to replace reaction networks by ODE systems so that one can easily run parameter estimation on ODE systems that are not motivated by reaction networks. I want to run parameter estimation on ODE systems with pyPESTO so this would be the core extension necessary for that, I think the scenario appears often. I talked to @paulstapor about this and he suggested this rough guideline of what would have to be done:

This files takes SBML-files as input https://github.com/ICB-DCM/AMICI/blob/master/python/amici/sbml_import.py, there would need to be an an alternative standardised format that directly corresponds to ODE systems which would be accepted in a similar file. The output format (refer to the class ODEModel) has to be consistent with the above mentionend file and is documented here https://github.com/ICB-DCM/AMICI/blob/master/python/amici/ode_export.py.

dweindl:
* Add PySB import, PEtab import, plain ODEs (Closes #654)
* Cleanup
* Update

### DIFF
--- a/documentation/PYTHON.md
+++ b/documentation/PYTHON.md
@@ -1,55 +1,145 @@
 # Python Interface {#python_interface}
 
-In the following we will give a detailed overview how to specify models in  Python and how to call the generated simulation files.
+In the following we will give a detailed overview how to specify models in
+Python and how to call the generated simulation files.
 
 ## Model Definition
 
-This guide will guide the user on how to specify models in Python using SBML. For example implementations see the examples in the python/examples directory.
+This guide will guide the user on how to specify models to import and simulate
+them using the Python interface.
+For example implementations see the examples in the `python/examples`
+directory.
 
-### SBML input
+### SBML import
 
-First, import an sbml file using the `amici.sbml_import.SbmlImporter` class:
+[SBML](http://sbml.org/) is a commonly used format for specifying systems
+biology models. To import an SBML model into AMICI, first, load an SBML file
+using the `amici.sbml_import.SbmlImporter` class:
 
     import amici
-    sbmlImporter = amici.SbmlImporter('model_steadystate_scaled.sbml')
-    
-the sbml document as imported by [libSBML](http://sbml.org/Software/libSBML) is available as 
+    sbml_importer = amici.SbmlImporter('model_steadystate_scaled.sbml')
 
-    sbml = sbmlImporter.sbml
+the SBML model as imported by [libSBML](http://sbml.org/Software/libSBML)
+is available as
 
-### Constants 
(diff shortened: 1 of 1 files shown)


documentation/PYTHON.md
```diff
@@ -1,87 +1,150 @@
...
+is available as
 
-### Constants 
+    sbml_model = sbml_importer.sbml
 
-parameters that should be considered constants can be specified in a list of strings specifying the respective SbmlId of a parameter.
+#### Constants
 
-    constantParameters=['k4']
+Model parameters that should be considered constants can be specified in a list
```
FFroehlich, line 29: We should ideally say/repeat (or link to a definition) what constants are, i.e. no sensitivity computation w.r.t. those values, they can change for preequi/presim/sim and usually correspond to experimental conditions.

documentation/PYTHON.md
```diff
@@ -1,87 +1,150 @@
...
+`amici.sbml_import.SbmlImporter.sbml2amici`, passing all the previously defined
+model specifications:
 
-    sbmlImporter.sbml2amici('test', 'test', 
-                            observables=observables,
-                            constantParameters=constantParameters,
-                            sigmas=sigmas)
+    sbml_importer.sbml2amici('test', 'test',
+                             observables=observables,
+                             constant_parameters=constant_parameters,
```
FFroehlich, line 62: 👍

documentation/PYTHON.md
```diff
@@ -1,87 +1,150 @@
...
+This guide will guide the user on how to specify models to import and simulate
+them using the Python interface.
+For example implementations see the examples in the `python/examples`
+directory.
 
-### SBML input
+### SBML import
 
-First, import an sbml file using the `amici.sbml_import.SbmlImporter` class:
+[SBML](http://sbml.org/) is a commonly used format for specifying systems
```
FFroehlich, line 15: As this section involves code, doesn't it make more sense to also move this to an example notebook?

FFroehlich: same applies to later parts of this documentation.

documentation/PYTHON.md
dweindl: Yes. I'd like to include some kind of glossary to not repeat that everywhere.

documentation/PYTHON.md
dweindl: Had the same feeling. But didn't want to just remove it. I think we want to have the notebooks rendered into the sphinx documentation in the future. Once they are in the proper places, I'd drop these code examples. Would leave it until then...

PR #1216, opened 2020-08-20
Run swig only for buildext, not for sdist

dweindl:


FFroehlich:
> Don't really get what is happening here, but as long tests pass I'm happy.

Primarily having build-specific swig-outdir and fixing some swig stuff?

dweindl:
> > Don't really get what is happening here, but as long tests pass I'm happy.
> 
> Primarily having build-specific swig-outdir and fixing some swig stuff?

:D

Swig interface was packed into sdist, which should allow installing amici even if no swig is available. This was 1) not compatible with switching on/off hdf5 support as tried in #1213 2) was redundant with running swig during package installation. Now swig is only run during package installation. That means it is required for installation from pypi. But anyways, installing amici with C++ extension, but without swig available for model wrapping is not too helpful.

PR #1222, opened 2020-08-21
"Discrete" noise models

yannikschaelte:
For some projects (in particular when dealing with population dynamics SIR models), binomial and negative binomial noise would be of interest. Thus, it may be good to inherently support this in AMICI. In the provided implementation, we do not enforce discreteness on the simulations (this can be manually specified by the user by using round functions in the model), because this would destroy gradient-based optimization. Via gamma interpolation, the noise model should be meaningful/valid nonetheless.

* For some special functions, we have to add a dict translating to e.g. amici functions to cxxcode.
* polygamma is only available in the header-only boost library, which is automatically included if needed (needs to be available on the system, otherwise the user will notice I guess)
* Boost import works on Ubuntu and OSX. The test on Windows is currently disabled because reasons.

Further:
* Also check gradients for all noise models with adjoints.


python/amici/ode_export.py
```diff
@@ -2325,6 +2339,11 @@ def _write_function_file(self, function: str) -> None:
             f'}} // namespace model_{self.model_name}',
         ])
 
+        # check custom functions
+        for fun in CUSTOM_FUNCTIONS:
+            if 'include' in fun and any(fun['c++'] in line for line in lines):
```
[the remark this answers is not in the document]
dweindl: Right.

Regarding checking for boost, I think this might be somewhat of a pain. For me okay to leave that part, until this really causes problems... A general warning, that model building trouble maybe due to the additional requirement will do the job for me...

dweindl:
>     * Thus `n` should come from the experiment (this would be the `sigma` in the AMICI formulation, which probably sooner or later should be generalized).

Makes sense... maybe to discuss for v1.0...

.github/workflows/test_python_cplusplus.yml
```diff
@@ -49,7 +49,8 @@ jobs:
             libhdf5-serial-dev \
             python3-venv \
             swig \
-            lcov
+            lcov \
+            libboost-dev
```
dweindl, line 53:
```suggestion
            libboost-math-dev
```

python/amici/ode_export.py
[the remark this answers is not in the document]
yannikschaelte: If that is fine, I would leave out the checking for boost then for the moment. Where should one add the warning that optional dependencies may be the problem (with link to the INSTALL.md)?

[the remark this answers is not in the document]
dweindl: Something like this? https://github.com/AMICI-dev/AMICI/pull/1222/commits/6ea1364886a868f966fd09eef5d5ada0d83a76fd
Feel free to revert...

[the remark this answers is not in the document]
dweindl: PS: ideally something like that would be part of the generated model code, so that it is also displayed when building the extension independently of the import, but I guess that would quite some distutils pain again and is probably not used by too many people anyways.

python/tests/test_sbml_import_special_functions.py
```diff
@@ -0,0 +1,140 @@
...
+    for sensi_method in [amici.SensitivityMethod.forward,
+                         amici.SensitivityMethod.adjoint]:
+        solver = model.getSolver()
+        solver.setSensitivityMethod(sensi_method)
+        solver.setSensitivityOrder(amici.SensitivityOrder.first)
+        check_derivatives(
+            model, solver, edata, assert_fun, atol=1e-1, rtol=1e-1,
+            check_least_squares=False)
+
+    """Test for m > y, i.e. in region with 0 density"""
```
dweindl, line 110: Although I see the visual advantage of using `""""` strings instead of comments here, I'd avoid that, as static code analysis tools will complain about "statement with no effect".

yannikschaelte:
@lcontento I would for the moment leave the current parameterization. This way, with the optional boost checks it will be possible to pass alternative parameterizations via the Callable model definition in sbml2amici. When you need to use this then for real, I suggest finalizing the parameterization you need in a hard-coded manner and assigning a name to it (then it will also be easier to use from PEtab). I think eventually we may want to have alternative e.g. neg-bino parameterizations hard-coded.

INSTALL.md
```diff
@@ -491,10 +487,28 @@ They are automatically installed when installing the python package.
 The C++ interface requires `cmake` and a cblas-compatible BLAS to be
 installed.
 
-
 ### Optional
 
-__SuperLU_MT__, "a general purpose library for the direct solution of large,
+#### SuperLU_MT
+
+A "a general purpose library for the direct solution of large,
```
dweindl, line 494: "A a"?

python/amici/sbml_import.py
```diff
@@ -1891,6 +1891,21 @@ def nllh_y_string(str_symbol):
             y, m, sigma = _get_str_symbol_identifiers(str_symbol)
             return f'log(2*{sigma}*{m}*log(10)) ' \
                 f'+ Abs(log({y}, 10) - log({m}, 10)) / {sigma}'
+    elif noise_distribution in ['binomial', 'lin-binomial']:
```
dweindl, line 1894: Shouldn't they be mentioned in some docstring? same for next ones.

dweindl (approved):

INSTALL.md
yannikschaelte: I English much good.

PR #1224, opened 2020-08-25
Hierarchical Derivatives for Expressions

FFroehlich:
Avoids flattening out expressions, which can be a bad idea as they get more complicated and lengthy. Also enables to directly have pysb observables as amici observables, as there isn't really any reason not to.


dweindl (approved):
:+1: 

Good to avoid adding redundant expressions for observables.

Getting rid of the flattening may also address some of my issues with bloated model code...

FFroehlich:
Adds two things:

1. Allows Expressions to depend on each other as long as they are sorted according to their dependency (which is always true for pysb)
2. Makes generated c++ code a bit more human readable. This was necessary to check derivatives

I am still not 100% happy with the performance here since dwdp and dwdx can still get quite big and takes long to compute. Adding `dwdw` to the model may be necessary, which would allow the online computation of the respective expressions. Would still require computation of the sparsity patterns, which could be done a bit more cheaply (and doesn't require repeated multiplications, but only partial derivatives). Will address that in a seperate PR, but feedback is welcome.

python/tests/test_ode_export.py
```diff
@@ -8,20 +8,21 @@ def test_csc_matrix():
...
-        = amici.ode_export.csc_matrix(matrix, 'a')
+        = amici.ode_export.csc_matrix(matrix, rownames=['a1', 'a2'],
+                                      colnames=['b1', 'b2'])
 
     assert symbol_col_ptrs == [0, 2, 3]
     assert symbol_row_vals == [0, 1, 1]
     assert sparse_list == sp.Matrix([[1], [2], [3]])
-    assert symbol_list == ['a0', 'a1', 'a2']
-    assert str(sparse_matrix) == 'Matrix([[a0, 0], [a1, a2]])'
+    assert symbol_list == ['da1db1', 'da2db1', 'da2db2']
```
dweindl, line 17: :+1:

dweindl (approved):
Looks good from what I can tell.

PR #1225, opened 2020-08-25
Fix doxygen comments

dweindl:


dweindl:
> How did these issues evade tests?

I had the same question... 
The number of blanks after `WARN_LOGFILE` in the Doxyfile template changed :laughing: 
Fixed in https://github.com/AMICI-dev/AMICI/pull/1226

PR #1227, opened 2020-08-25
Update Doxyfile to recent doxygen

dweindl:
Drop obsolete settings, update config doc


dweindl:
Fixes doxygen warnings about obsolete tags.

PR #1229, opened 2020-08-26
C++ API to readthedocs

dweindl:
Set up breathe and exhale to render doxygen C++ API doc for inclusion by sphinx

Still some trouble with sphinx/breathe/exhale and function overloads (https://github.com/svenevs/exhale/pull/91, https://github.com/svenevs/exhale/issues/90)

Once https://github.com/svenevs/exhale/issues/27 will be addressed, this should be easy to include the matlab interface as well.


FFroehlich:
> Once [svenevs/exhale#27](https://github.com/svenevs/exhale/issues/27) will be addressed, this should be easy to include the matlab interface as well.

Wonder when that will happen 🙄

PR #1233, opened 2020-08-26
Documentation: use sphinx-hoverxref for hover tooltips in documentation

dweindl:
(only working on RTD, not locally)


dweindl:
That works beautifully

image

image

PR #1234, opened 2020-08-26
Update doc

dweindl:


documentation/PYTHON.rst
```diff
@@ -1,11 +1,11 @@
 ****************
-Python Interface
+Python interface
```
FFroehlich, line 2: Why remove capitalization here (and below)?

dweindl: Because it was inconsistently used. (And I prefer only initial uppercase ;))

Can life with either, if consistently.

FFroehlich: okay then lets keep it like that for now.

PR #1235, opened 2020-08-27
Never use stripped symbols for SBML models (Closes #1230)

### LINKED ISSUE #1230: Simple RateRule model and sensitivities
A simple SBML model with one species `z`, one parameter `alpha` and `zdot = alpha` set by rate rule crashes AMICI when I try to compute state sensitivities (if sensitivity is not requested it works). MWE is
```python
with open('test.xml', 'w') as f:
    f.write("""\
<sbml xmlns="http://www.sbml.org/sbml/level2/version5" level="2" version="5">
    <model id="spline_test">
        <listOfCompartments>
            <compartment id="compartment" size="1"/>
        </listOfCompartments>
        <listOfSpecies>
            <species id="z" compartment="compartment" initialAmount="0"/>
        </listOfSpecies>
        <listOfParameters>
            <parameter id="alpha" value="1.0"/>
        </listOfParameters>
        <listOfRules>
            <rateRule variable="z">
                    <ci> alpha </ci>
                </math>
            </rateRule>
        </listOfRules>
    </model>
</sbml>""")

from amici.sbml_import import SbmlImporter
imp = SbmlImporter("test.xml")
imp.sbml2amici("test", "amici_models", compile=True)

import sys
sys.path.insert(0, "amici_models")
import test
model = test.getModel()

import amici
solver = model.getSolver()
solver.setRelativeTolerance(1e-6)
solver.setAbsoluteTolerance(1e-9)
solver.setSensitivityOrder(amici.SensitivityOrder_first)
solver.setSensitivityMethod(amici.SensitivityMethod_forward)

import numpy as np
model.setTimepoints(np.linspace(0, 1, 10))

amici.runAmiciSimulation(model, solver)
```
Is it a bug in AMICI? Or am I doing something wrong?

lcontento:
A possible solution to close https://github.com/AMICI-dev/AMICI/issues/1230.
* It seems to me that stripped symbols should only be used with PySB models, but they are also used in some cases with SBML models with rate rules. This leads to `dxdotdp` not being computed correctly due to mismatches in the SymPy symbol assumptions. This PR makes the stripped symbols exactly the same as the non-stripped ones when the `ODEModel` is created from a SBML file.
* SBML spec does not require the "name" attribute to be set (this lead to stripped symbols being `None`, but the above fix removes this possibility). I am not sure if we want the "name" attribute to be always set in AMICI (in which case the SBML "id" attribute may be a reasonable default, as in this PR) or we should allow for the possibility of it being an empty string.


python/amici/sbml_import.py
```diff
@@ -484,7 +484,10 @@ def _process_species(self) -> None:
         self.symbols['species']['identifier'] = sp.Matrix(
             [sp.Symbol(spec.getId(), real=True) for spec in species]
         )
-        self.symbols['species']['name'] = [spec.getName() for spec in species]
+        self.symbols['species']['name'] = [
+            spec.getName() if len(spec.getName()) != 0 else spec.getId()
```
dweindl, line 488:
```suggestion
            spec.getName() if spec.isSetName() else spec.getId()
```

python/amici/sbml_import.py
```diff
@@ -825,7 +828,8 @@ def _process_parameters(self,
                 [sp.Symbol(par.getId(), real=True) for par in settings['var']]
             )
             self.symbols[partype]['name'] = [
-                par.getName() for par in settings['var']
+                par.getName() if len(par.getName()) != 0 else par.getId()
```
dweindl, line 831:
```suggestion
                par.getName() if par.isSetName() else par.getId()
```

python/amici/sbml_import.py
```diff
@@ -1831,7 +1835,7 @@ def assignmentRules2observables(sbml_model: sbml.Model,
         parameter_id = p.getId()
         if filter_function(p):
             observables[parameter_id] = {
-                'name': p.getName(),
+                'name': p.getName() if len(p.getName()) != 0 else p.getId(),
```
dweindl, line 1838:
```suggestion
                'name': p.getName() if p.isSetName() else p.getId(),
```

dweindl (approved):
Works for me as fix for #1230. Beyond that, I think we should work on getting the SBML/pysb dichotomy out of ODEExporter, as this somehow defeats the purpose of having separate model importers but a joint ODE exporter for the different formats. 

Please wait for approval by @FFroehlich.

FFroehlich:
> Works for me as fix for #1230. Beyond that, I think we should work on getting the SBML/pysb dichotomy out of ODEExporter, as this somehow defeats the purpose of having separate model importers but a joint ODE exporter for the different formats.
> 
> Please wait for approval by @FFroehlich.

Agree with Daniel here. Would having the strip_pysb in there at the moment is primarily for performance reasons (as far as I remember). I am still hoping that we will at some point no longer need any strip_pysb ... . 

Could we fix this issue by defaulting `comp.getName()` to `comp.getId()` if name is not set?

lcontento:
> Could we fix this issue by defaulting `comp.getName()` to `comp.getId()` if name is not set?

Unfortunately this is not sufficient, because the stripped symbols are lacking the `real=True` assumption (I understand that this is necessary for PySB) which the symbols coming from the SBML importer instead have.

FFroehlich:
> > Could we fix this issue by defaulting `comp.getName()` to `comp.getId()` if name is not set?
> 
> Unfortunately this is not sufficient, because the stripped symbols are lacking the `real=True` assumption (I understand that this is necessary for PySB) which the symbols coming from the SBML importer instead have.

Okay, really need to push that we get this fixed in pysb ...

dweindl:
> Unfortunately this is not sufficient, because the stripped symbols are lacking the `real=True` assumption

Couldn't we just re-apply the assumptions on the stripped symbols?

lcontento:
> > Unfortunately this is not sufficient, because the stripped symbols are lacking the `real=True` assumption
> 
> 
> 
> Couldn't we just re-apply the assumptions on the stripped symbols?

It may not work if the parameter has both a SBML ID and a SBML name and they differ. In that case the stripped symbols would have correct assumptions but non-matching names, so I am afraid that derivatives would end up zero again. I will check tomorrow to be sure.

FFroehlich:
> > > Unfortunately this is not sufficient, because the stripped symbols are lacking the `real=True` assumption
> > 
> > 
> > Couldn't we just re-apply the assumptions on the stripped symbols?
> 
> It may not work if the parameter has both a SBML ID and a SBML name and they differ. In that case the stripped symbols would have correct assumptions but non-matching names, so I am afraid that derivatives would end up zero again. I will check tomorrow to be sure.

Thinking about this again, I think one should always use. 

```
self._strippedsyms[name] = sp.Matrix([
     sp.Symbol(str(comp.get_id()), real=True)
    for comp in getattr(self, component)
])
```

SBML names should only affect the return of getParameterNames and shouldn't be reflected anywhere in the equations.

lcontento:
> Thinking about this again, I think this should work and is actually the right thing to do here:
> 
> ```
> self._strippedsyms[name] = sp.Matrix([
>     sp.Symbol(comp.get_id().name)  if hasattr(comp.get_id(),'name') else comp.get_id()
>     for comp in getattr(self, component)
> ])
> ```
> 
> SBML names should only affect the return of getParameterNames and shouldn't be used anywhere in the equations.

With the current PR SBML names should never be used inside the equations.

As for your code proposal, isn't `comp.get_id()` always a SymPy symbol and as such always has a `name` attribute?
In general, I don't see any way to distinguish whether a parameter comes from SBML or PySB just by looking at it at this point of the pipeline...

dweindl:
@lcontento : Let's merge as is to have #1230 fixed.

FFroehlich:
> > Thinking about this again, I think this should work and is actually the right thing to do here:
> > ```
> > self._strippedsyms[name] = sp.Matrix([
> >     sp.Symbol(comp.get_id().name)  if hasattr(comp.get_id(),'name') else comp.get_id()
> >     for comp in getattr(self, component)
> > ])
> > ```
> > 
> > 
> > SBML names should only affect the return of getParameterNames and shouldn't be used anywhere in the equations.
> 
> With the current PR SBML names should never be used inside the equations.
> 
> As for your code proposal, isn't `comp.get_id()` always a SymPy symbol and as such always has a `name` attribute?
> In general, I don't see any way to distinguish whether a parameter comes from SBML or PySB just by looking at it at this point of the pipeline...

plain `sp.Symbols` shouldn't have a `name` attribute, only if they are coming from pysb and are derived `pysb.component`s. most straightforward check would be to check for `isinstance(comp, pysb.component)`, but that would introduce a pysb dependency in ode_export.

dweindl:
> most straightforward check would be to check for `isinstance(comp, pysb.component)`, but that would introduce a pysb dependency in ode_export.

Not beautiful, but could be avoided by just string-checking class names on the object. 

but can't we get that whole distinction out of ode_exporter by preprocessing the relevant symbols in pysb_importer?

FFroehlich:
> > most straightforward check would be to check for `isinstance(comp, pysb.component)`, but that would introduce a pysb dependency in ode_export.
> 
> Not beautiful, but could be avoided by just string-checking class names on the object.

I wasn't aware `sympy.Symbols` also have `name` attributes.

```
self._strippedsyms[name] = sp.Matrix([
    comp.get_id() if not hasattr(comp.get_id(), 'model')
    else sp.Symbol(comp.get_id().name, **comp.get_id().assumptions0)
    for comp in getattr(self, component)
])
```
should actually work and also preserve assumptions from pysb that were recently introduced.

> but can't we get that whole distinction out of ode_exporter by preprocessing the relevant symbols in pysb_importer?

We can, but that would require substituting all pysb symbols during export which is quite expensive for larger models, so we are likely to take a big performance hit.

dweindl:
> We can, but that would require substituting all pysb symbols in the importer which is quite expensive for larger models, so we are likely to take a big performance hit.

Do you have a feeling for how big the impact is? Which fraction do we (not) have to strip? Is the stripping currently happening only once per symbol, or doesn't that even occur multiple times for the one in the current code?

FFroehlich:
replacing with stripped symbols would have to be done everywhere. The last time I tried applying all substitutions was comparable or higher than regular model compilation. We could remove `_strippedsyms` if all pysb variants would consistently use stripped/nonstripped symbols (as discussed, will hopefully happen at some point).

PR #1240, opened 2020-08-29
Better sparse multiplication

FFroehlich:
This computes sparsity pattern of sparse matrix-matrix products on the fly meaning we no longer have to compute them symbolically reducing the need for symbolic processing. Memory allocation is performed dynamically (in the hope that the sparsity pattern doesn't change too often during simulation).

🚀

### DIFF
--- a/include/amici/sundials_matrix_wrapper.h
+++ b/include/amici/sundials_matrix_wrapper.h
@@ -129,10 +129,16 @@ class SUNMatrixWrapper {
     sunindextype columns() const;
 
     /**
-     * @brief Get the number of non-zero elements (sparse matrices only)
+     * @brief Get the number of specified non-zero elements (sparse matrices only)
      * @return number
      */
     sunindextype nonzeros() const;
+    
+    /**
+     * @brief Get the number of allocated non-zero elements (sparse matrices only)
+     * @return number
+     */
+    sunindextype nonzero_space() const;
 
     /**
      * @brief Get the index values of a sparse matrix
@@ -212,15 +218,15 @@ class SUNMatrixWrapper {
     
     /**
      * @brief x = x + beta * A(:,j), where x is a dense vector and A(:,j) is sparse, and construct the pattern
-     * for C(:,icol)
+     * for C(:,j)
      * @param j column index
      * @param beta scaling factor
      * @param w temporary index workspace, this keeps track of the sparsity pattern in C
      * @param x temporary data workspace, this keeps track of the data in C
      * @param mark marker for w to indicate nonzero pattern
      * @param C  output matrix
-     * @param nz number of nonzeros
-     * @return updated number of nonzeros
+     * @param nz number of nonzeros that were already written to C
+     * @return updated number of nonzeros in C
      */
     sunindextype scatter(const sunindextype j, const realtype beta,
                          sunindextype *w, realtype *x, const sunindextype mark,
--- a/include/amici/sundials_matrix_wrapper.h
+++ b/include/amici/sundials_matrix_wrapper.h
@@ -133,13 +133,13 @@ class SUNMatrixWrapper {
      * @note values will be unininitialized before indexptrs are set.
      * @return number
      */
-    sunindextype nonzeros() const;
+    sunindextype num_nonzeros() const;
     
     /**
      * @brief Get the number of allocated non-zero elements (sparse matrices only)
      * @return number
      */
-    sunindextype nonzero_space() const;
+    sunindextype capacity() const;
 
     /**
      * @brief Get the index values of a sparse matrix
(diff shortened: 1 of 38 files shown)


include/amici/sundials_matrix_wrapper.h
```diff
@@ -117,10 +129,16 @@ class SUNMatrixWrapper {
...
+     * @brief Get the number of specified non-zero elements (sparse matrices only)
      * @return number
      */
     sunindextype nonzeros() const;
+    
+    /**
+     * @brief Get the number of allocated non-zero elements (sparse matrices only)
+     * @return number
+     */
+    sunindextype nonzero_space() const;
```
dweindl, line 141:
```suggestion
    sunindextype capacity() const;
```

To be coherent with std::vector?

include/amici/sundials_matrix_wrapper.h
```diff
@@ -190,14 +208,29 @@ class SUNMatrixWrapper {
                   bool transpose) const;
 
     /**
-     * @brief Perform matrix matrix multiplication
-              C[:, :] += A * B
+     * @brief Perform matrix matrix multiplication A * B
               for sparse A, B, C
-     * @param C output matrix, may already contain values
+     * @param C output matrix, may not contain values but may be preallocated
```
dweindl, line 213: > may not contain values

"will be overwritten"?

> may be preallocated

"will be reallocated to fit" ?

include/amici/sundials_matrix_wrapper.h
```diff
@@ -190,14 +208,29 @@ class SUNMatrixWrapper {
...
+    
+    /**
+     * @brief x = x + beta * A(:,j), where x is a dense vector and A(:,j) is sparse, and construct the pattern
+     * for C(:,j)
+     * @param j column index
+     * @param beta scaling factor
+     * @param w temporary index workspace, this keeps track of the sparsity pattern in C
+     * @param x temporary data workspace, this keeps track of the data in C
+     * @param mark marker for w to indicate nonzero pattern
+     * @param C  output matrix
```
dweindl, line 227:
```suggestion
     * @param C output matrix
```

include/amici/sundials_matrix_wrapper.h
```diff
@@ -190,14 +208,29 @@ class SUNMatrixWrapper {
...
+     * @param nz number of nonzeros that were already written to C
```
dweindl, line 228:
```suggestion
     * @param nnz number of nonzeros that were already written to C
```

For consistency?

src/sundials_matrix_wrapper.cpp
```diff
@@ -100,6 +100,31 @@ operator=(SUNMatrixWrapper &&other) {
     return *this;
 }
 
+void SUNMatrixWrapper::reallocate(int NNZ) {
```
dweindl, line 103: having both `realloc` and `reallocate` with different purposes seems to be asking for trouble... not sure about better naming though...

src/sundials_matrix_wrapper.cpp
```diff
@@ -349,37 +387,101 @@ void SUNMatrixWrapper::sparse_multiply(SUNMatrixWrapper *C,
...
+    {
+        Cp[j] = nz;                          /* column j of C starts here */
+        // CSparse reallocation happend when it _might_ have been necessary.
+        // We are a little bit more conservative here as we allow more
+        // reallocations on the first call but try to avoid them on later calls.
+
+        if ((Bp[j+1] > Bp[j]) && (nz + m > C->nonzero_space()))
+        {
+            C->reallocate(2*C->nonzero_space() + m);
+            Cx = C->data(); Ci = C->indexvals(); Cp = C->indexptrs();
```
dweindl, line 445: one statement per line?

src/sundials_matrix_wrapper.cpp
```diff
@@ -349,37 +387,101 @@ void SUNMatrixWrapper::sparse_multiply(SUNMatrixWrapper *C,
...
+        }
+        for (p = Bp[j]; p < Bp[j+1]; p++) {
+            nz = scatter(Bi[p], Bx[p], w.data(), x.data(), j+1, C, nz);
+        }
+        for (p = Cp[j]; p < nz; p++)
+            Cx[p] = x.at(Ci[p]); // copy data to C
```
dweindl, line 451: check dimensions before, and don't use `at`? otherwise no vectorization possible.

dweindl (approved):
Looks good as far as I can tell. What's the impact on import and simulation time?

FFroehlich:
> Looks good as far as I can tell. What's the impact on import and simulation time?

Should hopefully have negligible impact on simulation time dynamic allocation should only be done on first function call.

src/sundials_matrix_wrapper.cpp
FFroehlich: thats the sundials naming scheme ... since realloc has no int argument, I would hope confusion shouldn't be too big.

PR #1241, opened 2020-08-30
Matlab documentation to RTD

dweindl:


dweindl:
Works locally, but for unknown reasons it doesn't build on RTD. See https://github.com/readthedocs/readthedocs.org/issues/7435.

FFroehlich:
> Works locally, but for unknown reasons it doesn't build on RTD. See [readthedocs/readthedocs.org#7435](https://github.com/readthedocs/readthedocs.org/issues/7435).

Same problem for other PRs, so this doesn't seem to be related to any changes you applied here.

PR #1244, opened 2020-08-31
Feature sparse addition

FFroehlich:


paulstapor:
Will look into this tomorrow more closely. The first two questions/remarks arising are:
 * What has happened to the performance test we used to have? And linked to that: How does this affect performance for adjoints in large-scale models?
 * Would be good to have more documentation on what the new functions are doing. Just by reading quickly over the new scatter method, there's no way to understand what even its purpose is, not talking about documentation of the logic. Refs to docs are nice and a plus. Yet, I should be able of understanding at least roughly what the method does by looking at the code and doc, without having to open my browser.

FFroehlich:
> Will look into this tomorrow more closely. The first two questions/remarks arising are:
> 
> * What has happened to the performance test we used to have? And linked to that: How does this affect performance for adjoints in large-scale models?

Currently on run on PRs to master, would hope this has marginal influence on simulation time. Adding these functions will however be helpful with future extensions.

> * Would be good to have more documentation on what the new functions are doing. Just by reading quickly over the new scatter method, there's no way to understand what even its purpose is, not talking about documentation of the logic. Refs to docs are nice and a plus (thumbs up here). Yet, it should be able to understand at least roughly what a method does by looking at the code and doc, without having to open a browser.

Correct. Added some more documentation.

include/amici/sundials_matrix_wrapper.h
```diff
@@ -209,35 +209,60 @@ class SUNMatrixWrapper {
...
     /**
-     * @brief Perform matrix matrix multiplication A * B
+     * @brief Perform matrix matrix multiplication C = A * B
               for sparse A, B, C
      * @param C output matrix,
      * @param B multiplication matrix
-     * @note will overwrite existing data, indexptrs, indexvals, but will use preallocated space for these vars
+     * @note will overwrite existing data, indexptrs, indexvals for C, but will use preallocated space for these vars
      */
     void sparse_multiply(SUNMatrixWrapper *C,
```
dweindl, line 218: is nullptr allowed? if so, document what happens then. if not, use reference

include/amici/sundials_matrix_wrapper.h
```diff
@@ -209,35 +209,60 @@ class SUNMatrixWrapper {
...
+     * @param alpha scalar A
+     * @param B addition matrix
+     * @param beta scalar B
+     * @note will overwrite existing data, indexptrs, indexvals for C, but will use preallocated space for these vars
+     */
+    void sparse_add(SUNMatrixWrapper *A, realtype alpha,
+                    SUNMatrixWrapper *B, realtype beta);
+    
+    /**
+     * @brief x = x + beta * A(:,j), where x is a dense vector and A(:,j) is sparse, and constructs the
```
dweindl, line 233:
```suggestion
     * @brief Compute x = x + beta * A(:,j), where x is a dense vector and A(:,j) is sparse, and constructs the
```

src/sundials_matrix_wrapper.cpp
```diff
@@ -363,47 +369,19 @@ void SUNMatrixWrapper::multiply(gsl::span<realtype> c,
 
 void SUNMatrixWrapper::sparse_multiply(SUNMatrixWrapper *C,
                                        SUNMatrixWrapper *B) const {
-    if (!matrix_)
+    if (!matrix_ || !B->matrix_ || !C->matrix_)
```
dweindl, line 372: Must check for nullptr before dereferencing (or better pass reference)

src/sundials_matrix_wrapper.cpp
```diff
@@ -453,47 +431,120 @@ void SUNMatrixWrapper::sparse_multiply(SUNMatrixWrapper *C,
...
     /*
      * do not reallocate here since we rather keep a matrix that is a bit
      * bigger than repeatedly resizing this matrix.
      */
 }
 
+void SUNMatrixWrapper::sparse_add(SUNMatrixWrapper *A, realtype alpha,
+                                  SUNMatrixWrapper *B, realtype beta) {
+    // matrix_ == nullptr is allowed on the first call
+    if (!A->matrix_ || !B->matrix_)
```
dweindl, line 452: Must check for nullptr before dereferencing (A and B, not only their matrix_ member). Pass A,B as reference?

include/amici/sundials_matrix_wrapper.h
```diff
@@ -209,35 +209,59 @@ class SUNMatrixWrapper {
...
+     *   - perform parts of sparse matrix-matrix multiplication C(:,j)=A(:,j)*B(j,k)
+     *    enabled by passing beta=B(j,k), x=C(:,j), C=C, w=sparsity of C(:,j) from B(j,0...k-1), nnz=nnz(C(:,0...j-1)
+     *   - add the jth column of the sparse matrix A to the dense vector x.
+     *    enabled by passing beta=1.0, x=x, C=nullptr, w=nullptr, nnz=0
+     *
      * @param j column index
      * @param beta scaling factor
-     * @param w temporary index workspace, this keeps track of the sparsity pattern in C
-     * @param x temporary data workspace, this keeps track of the data in C
+     * @param w index workspace, (w[i]<mark) indicates non-zeroness of C(i,j)
```
dweindl, line 243: What's the expected size? What happens in nullptr?

dweindl (approved):

include/amici/sundials_matrix_wrapper.h
```diff
@@ -209,35 +209,60 @@ class SUNMatrixWrapper {
...
+     */
+    void sparse_add(const SUNMatrixWrapper &A, realtype alpha,
+                    const SUNMatrixWrapper &B, realtype beta);
+    
+    /**
+     * @brief Compute x = x + beta * A(:,j), where x is a dense vector and A(:,j) is sparse, and update
+     * the sparsity pattern for C(:,j) if applicable
+     *
+     * This function currently has two purposes:
+     *   - perform parts of sparse matrix-matrix multiplication C(:,j)=A(:,j)*B(j,k)
```
paulstapor, line 236: That's C(:,k)

include/amici/sundials_matrix_wrapper.h
```diff
@@ -209,35 +209,60 @@ class SUNMatrixWrapper {
...
+     * @note will overwrite existing data, indexptrs, indexvals for C, but will use preallocated space for these vars
+     */
+    void sparse_add(const SUNMatrixWrapper &A, realtype alpha,
+                    const SUNMatrixWrapper &B, realtype beta);
+    
+    /**
+     * @brief Compute x = x + beta * A(:,j), where x is a dense vector and A(:,j) is sparse, and update
+     * the sparsity pattern for C(:,j) if applicable
+     *
+     * This function currently has two purposes:
```
paulstapor, line 235: :+1:

src/newton_solver.cpp
```diff
@@ -98,26 +98,9 @@ void NewtonSolver::computeNewtonSensis(AmiVectorArray &sx) {
...
-                auto col = model_->dxdotdp_implicit.indexptrs();
-                auto row = model_->dxdotdp_implicit.indexvals();
-                auto data_ptr = model_->dxdotdp_implicit.data();
-                for (sunindextype iCol = col[model_->plist(ip)];
-                     iCol < col[model_->plist(ip) + 1]; ++iCol)
-                    sx.at(static_cast<int>(row[iCol]), ip) -= data_ptr[iCol];
-            }
+            model_->dxdotdp_full.scatter(model_->plist(ip), -1.0, nullptr,
+                                         gsl::make_span(sx.getNVector(ip)),
+                                         0, nullptr, 0);
```
paulstapor, line 103: :+1:

src/sundials_matrix_wrapper.cpp
```diff
@@ -419,81 +399,154 @@ void SUNMatrixWrapper::sparse_multiply(SUNMatrixWrapper *C,
...
+
+    /* see https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/master/CSparse/Source/cs_add.c
+     * modified such that we don't need to use CSparse memory structure and can
+     * work with preallocated C. This should minimize number of necessary
+     * reallocations as we can assume that C doesn't change size.
+     */
+    
+    sunindextype nnz = 0; // this keeps track of the nonzero index in C
+    
+    sunindextype j;
```
paulstapor, line 482: Not urgent to be addressed, but general remark:
I personally prefer use looping variables such as "iCol" or "iData", making it clear right away what is looped over, rather than "i", "j", "k". Would be happy, if we could adhere to something like that when writing new code. :)

src/sundials_matrix_wrapper.cpp
```diff
@@ -227,6 +238,32 @@ void SUNMatrixWrapper::multiply(N_Vector c, const_N_Vector b) const {
              gsl::make_span<const realtype>(NV_DATA_S(b), NV_LENGTH_S(b)));
 }
 
+static void check_dim(sunindextype n, sunindextype m, const std::string &name_n,
```
paulstapor, line 241: Good to have that!

src/model.cpp
```diff
@@ -135,7 +135,10 @@ Model::Model(const int nx_rdata, const int nxtrue_rdata, const int nx_solver,
                              ndxdotdp_explicit, CSC_MAT);
         dxdotdp_implicit =
             SUNMatrixWrapper(nx_solver, static_cast<int>(p.size()),
-                             ndwdp + ndxdotdw, CSC_MAT);
+                             ndwdp + ndxdotdw, CSC_MAT); // guess size
```
paulstapor, line 138: Does this need to be an upper bound, as it is used for capacity?

paulstapor (approved):
A only have minor things here. Some comments:
 * I think the new dxdotdp implementation will make it a little easier to reimplement DAEs at some point, if needed. That's a plus.
 * I like the new functionality of sparse add. This is likely to be helpful when implementing a relaxed Newton/Richardson method in the Steadystate solver, which is planned at some not too far point.
 * Thanks for additional documentation
 * Preformance to be checked either before merging into master, but can also do that right away for this PR, if you want.
 * do I see it correctly, that the capacity of dxdotdp is guessed in the beginning and then capacity and sparsity pattern will be fixed by the sparse multiplication?

src/model.cpp
FFroehlich: No, reallocate will allocate more space if necessary.

src/sundials_matrix_wrapper.cpp
FFroehlich: I generally agree with you. Most of this code is copy pasted though so the tradeoff is between being able to compare the implemenation to the original code and following this guidelines and I opted with having the code more comparable to the original implementation.

include/amici/sundials_matrix_wrapper.h
FFroehlich: 👍  correct still a bit confusing with outer and inner `j` being called the same.

Issue #1198, opened 2020-08-10
How to encode splines in the python or C++ interface?


paulstapor:
Matlab had a hand-crafted way of encoding splines in the symbolic interface. However, SBML does not explicitly support splines, for pySB I'm not sure...
I would like to discuss how to best address this issue. Maybe there's an SBML extension. maybe there's some other way... Splines are a useful tool and it would be good if Amici could still support them...

paulstapor:
Just looked into current SBML extensions. There seems to be a planned extension called SBMLmath, but that seems not yet usable. So, I'd be happy about any suggestion here...

Everything that comes to my mind are hacks like "Creating a script, which computes the interpolation polynomial of the spline and modifies the SBML file prior to Amici Import"... That would obviously work, but is not the thing I would really like to go for in Amici...

paulstapor:
As this is currently important in a project of @lcontento , we discussed the following ideas:
Technically, one can encode a spline directly in the SBML by hacking in the formula using polynomials and piecewise in a state assignment rule (good choice would be cubic Hermite splines, due to their simple formulas, but in principle, also classical interpolating splines would work) . However, splines with many nodes will produce monsters of SBML and later then cpp-files, with hours of symbolic processing, which are also inefficient to evaluate numerically.

Another way would be going via some state assignment rule, which contains a specific type of annotation, which AMICI will parse automatically into something which can be turned into C++ code much more elegantly. Roughly, this would be similar to the old Matlab spline implementation. Except, that the old implementation was pretty unsafe C, with some pitfalls due to old, badly documented code, and that it was quite inefficient, as it fully recomputed the interpolation polynomial at each call to the function.

So here's what we came up with:
Let's indeed advice the user to encode a spline in full SBML (if possible and user willing), but at the same time add an annotation. If AMICI finds this annotation, it will discard the state assignemt rule and use its own spline-parser instead. If the spline encoded in SBML is of the same type as requested in the annotation, the AMICI implementation will however produce something which should give exactly the same simulation results as the original SBML file - just more elegant and numerically more efficient.
On the C++-side, I would like to write a spline-class. Each spline used will be an instance of this class, which will replace the species with the state assignment rules and spline annotation. Splines are initialized once before the beginning of the simulation (if it's e.g. an interpolating spline, which isn't our first choice, but would be easy to implement). At this point, the interpolation is carried out once and the spline coefficients are stored as members of the spline object. This will make later calls to splines very efficient.

[...]

PR #1245, opened 2020-08-31
Python splines

paulstapor:
This thing is gonna to be spline support in Amici on the python side, therefore fixing #1198 .
However, before this becomes a valid PR, there are quite some things to be sorted out...
(everyone, feel free to add things to the list or discuss about the items, i.e., "I don't think that point xyz is important for merging" is also an appreciated contribution :) )
 - [ ] Adapt boundary condition and extrapolation in C++ code, according to python side
 - [ ] change handling in spline_constructors file (not returning a vector, but passing a vector of AbstractSpline pointers)
 - [x] Bugfix current extrapolation implementation on C++ side
 - [ ] Implement periodic boundary condition on C++ side
 - [ ] Add a "final value member" to spline class, to make it play nicely with steadystate solver (if spline is to be evaluated at `t = inf`, the "final value" member should simply be read.
 - [ ] Implement spline reinitialization, to make it work with preequilibration and presimulation
 - [x] Double-check example from lcontento which is still failing
 - [ ] Add docstrings (looks like hell at least on C++ side atm :imp: )
 - [ ] Allow splines also for species assignment rules (not only for parameters) and integrate with x_rdata in C++
 - [ ] Allow also formulae for spline slopes, i.e., don't force finite differences for Hermite splines
 - [ ] Implement tension parameter for Hermite splines? (i.e., allowing genera cardinal splines?)
 - [ ] Thinking carefully of which data members should be private, protected, public in splne class
 - [ ] Address sonarcloud code review 
 - [ ] Implementing C2-splines as other subclass?
 - [ ] Switching Matlab implementation to new spline class, 
 - [ ] Maybe reimplement JakStat-Swameye model with new spline class and add to benchmark collection?
 - [x] Implement natural boundary conditions in C++
 - [ ] Add a very extensive test suite, which checks:
     - [ ] Multiple splines with different length
     - [ ] Arbitrary formulae in spline nodes (may depend on p, k, and hard coded numeric values
     - [ ] Check splines  for species assignment rules
     - [ ] Formulae for spline slopes
     - [ ] Basically all this for all types of boundary conditions and extrapolations
     - [ ] All this with checking of `x`, `sx`, `y`, `sy`, `sllh`
     - [ ] Ideally all checked via parsing an Amici anotation, running through the full formula, and checking with analytically computed value


include/amici/symbolic_functions.h
```diff
@@ -3,32 +3,242 @@
...
+ * negative values
+ *
+ * @param x argument
+ * @return if(x>0) then log(x) else -Inf
+ *
+ */
 double log(double x);
+
+/**
+ * C implementation of matlab function dirac
```
yannikschaelte, line 17: I would not write "matlab" for these functions, as this is just an implementation of that function, not specific to the matlab side (same for heaviside, min, max, ...).

paulstapor: Was just cut-copy-pasting old docstrings. Yep, would also be a good possibility to update them ;)

dweindl:
Superseded by #1515

PR #1246, opened 2020-09-01
Compute Hierarchical Derivatives Numerically

FFroehlich:


src/model.cpp
```diff
@@ -130,23 +130,42 @@ Model::Model(const int nx_rdata, const int nxtrue_rdata, const int nx_solver,
...
+        dwdw_ = SUNMatrixWrapper(nw, nw, ndwdw, CSC_MAT);
+        // size dynamically adapted for dwdx_ and dwdp_
+        dwdx_ = SUNMatrixWrapper(nw, nx_solver, 0, CSC_MAT);
+        dwdp_ = SUNMatrixWrapper(nw, static_cast<int>(p.size()), 0, CSC_MAT);
+
+        for (int irec = 0; irec <= w_recursion_depth_; ++irec) {
+            dwdp_hierarchical_.emplace_back(
+                SUNMatrixWrapper(nw, static_cast<int>(p.size()), ndwdw + ndwdp,
+                                 CSC_MAT)); // guessed size
+            dwdx_hierarchical_.emplace_back(SUNMatrixWrapper(
```
paulstapor, line 143: If I get it correctly, this is computationally more heavy than the current implementation, but will be more efficient for large models during simulation, right?
Can we quantify this somehow?

python/amici/ode_export.py
```diff
@@ -1717,37 +1730,17 @@ def _derivative(self, eq: str, var: str, name: str = None) -> None:
...
+
+        # compute recursion depth based on nilpotency of jacobian. computing
+        # nilpotency can be done more efficiently on numerical sparsity pattern
+        if name == 'dwdw':
+            nonzeros = np.asarray(
+                self._eqs[name].applyfunc(lambda x: int(not x.is_zero))
+            ).astype(np.int64)
+            while nonzeros.max():
+                nonzeros = nonzeros.dot(nonzeros)
+                self._w_recursion_depth += 1
```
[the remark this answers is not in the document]
paulstapor: We will need to do a new release soon...

paulstapor (approved):
Looks pretty good overall. Would be interested in how this affects performance and how dwdp and dwdw will play together with splines in the near future...
If I get it correct: This kind of redoes what symbolic optimization in Matlab did long ago, right?

src/model.cpp
```diff
@@ -130,23 +130,42 @@ Model::Model(const int nx_rdata, const int nxtrue_rdata, const int nx_solver,
...
-                             ndxdotdp_explicit, CSC_MAT);
+
+        dwdw_ = SUNMatrixWrapper(nw, nw, ndwdw, CSC_MAT);
+        // size dynamically adapted for dwdx_ and dwdp_
+        dwdx_ = SUNMatrixWrapper(nw, nx_solver, 0, CSC_MAT);
+        dwdp_ = SUNMatrixWrapper(nw, static_cast<int>(p.size()), 0, CSC_MAT);
+
+        for (int irec = 0; irec <= w_recursion_depth_; ++irec) {
+            dwdp_hierarchical_.emplace_back(
+                SUNMatrixWrapper(nw, static_cast<int>(p.size()), ndwdw + ndwdp,
```
dweindl, line 141: We can also add an additional constructor using the std::vector size type to get rid of all the static casts...

dweindl (approved):
Good, as far as I can tell.

include/amici/model.h
```diff
@@ -200,6 +203,34 @@ class Model : public AbstractModel {
     using AbstractModel::fx0_fixedParameters;
     using AbstractModel::fy;
     using AbstractModel::fz;
+    
+    /**
+     * @brief Model specific implementation of fdwdw_explicit, no w chainrule (Py)
+     * @param dxdotdp_explicit partial derivative xdot wrt p
```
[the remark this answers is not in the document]
FFroehlich: 👍

python/amici/ode_export.py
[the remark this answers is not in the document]
FFroehlich: correct see #1224 for details

src/model.cpp
FFroehlich: this only applies if `w` is recursive, otherwise (which is the case for sbml models with conservation laws) there should be 0 overhead.

FFroehlich: > this only applies if `w` is recursive, otherwise (which is the case for sbml models with conservation laws) there should be 0 overhead.

i.e. if this turns out to be a problem, an easy fix would be to just flatten `w` in ode_export, in which case w_recursion_depth_ would be 0.

scripts/run-valgrind-py.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Test python model wrapping inside virtual environment
+
+script_path=$(dirname $BASH_SOURCE)
+amici_path=$(cd "$script_path"/.. && pwd)
+
+set -e
+
+if [[ -z "${BNGPATH}" ]]; then
+    export BNGPATH=${amici_path}/ThirdParty/BioNetGen-2.3.2
```
dweindl, line 10: Unrelated to this PR, but I guess we could upgrade BNG at some point...

src/model.cpp
```diff
@@ -131,26 +130,44 @@ Model::Model(const int nx_rdata, const int nxtrue_rdata, const int nx_solver,
...
-                "Number of elements in ndJydy is not equal "
-                " nytrue.");
+
+        dwdw_ = SUNMatrixWrapper(nw, nw, ndwdw, CSC_MAT);
+        // size dynamically adapted for dwdx_ and dwdp_
+        dwdx_ = SUNMatrixWrapper(nw, nx_solver, 0, CSC_MAT);
+        dwdp_ = SUNMatrixWrapper(nw, p.size(), 0, CSC_MAT);
+
+        for (int irec = 0; irec <= w_recursion_depth_; ++irec) {
+            /* for the first element we know the exact size, while for all we
```
dweindl, line 140:
```suggestion
            /* for the first element we know the exact size, while for all others we
```
?

src/model.cpp
```diff
@@ -1265,9 +1281,17 @@ void Model::fdydp(const realtype t, const AmiVector &x) {
 
     /* get dydp slice (ny) for current time and parameter */
     for (int ip = 0; ip < nplist(); ip++)
-        fdydp(&dydp_.at(ip * ny), t, x.data(), state_.unscaledParameters.data(),
-              state_.fixedParameters.data(), state_.h.data(), plist(ip), w_.data(),
-              dwdp_.data());
+        if (pythonGenerated) {
```
dweindl, line 1284: Looking forward to the moment we can remove that distinction :)

src/model.cpp
```diff
@@ -1265,9 +1281,17 @@ void Model::fdydp(const realtype t, const AmiVector &x) {
...
+            fdydp(&dydp_.at(ip * ny), t, x.data(),
```
dweindl, line 1285: At some point we may want to change those signatures to take spans... Slightly better readable without all those `.data()`...

src/sundials_matrix_wrapper.cpp
```diff
@@ -399,50 +457,36 @@ void SUNMatrixWrapper::sparse_multiply(SUNMatrixWrapper &C,
...
-    
-    sunindextype bCol;
-    sunindextype bcolptr;
-    sunindextype ccolptr;
-    auto Cx = C.data();
-    auto Ci = C.indexvals();
-    auto Cp = C.indexptrs();
-    
-    sunindextype m = nrows;
-    sunindextype n = B.columns();
```
dweindl, line 414: That was on purpose, to avoid the call to `B.columns()` in every iteration, because afaik it wasn't as cheap as expected.

dweindl (approved):
Phew, that's quite some changes. Looks good as far as I can tell.

dweindl:
SBML test suite runs are taking about half an hour longer here. Is that related to the changes here, or just temporary GHA issues? :thinking:

FFroehlich:
> SBML test suite runs are taking about half an hour longer here. Is that related to the changes here, or just temporary GHA issues? 🤔

Might be due to changes here, these changes here could introduce some overhead for smaller models, 25% isn't really something I would worry too much about as long as bigger models are not impacted in the same way. Will have to look at the performance test as soon as this is merged.

PR #1247, opened 2020-09-01
Remove J/JB/JSparse/JSparseB duplication, fixes #468

### LINKED ISSUE #468: reduce J, JSparse, JB, JSparseB code duplication
J, JSparse, JB, JSparseB all contain the same entries but are currently compiled multiple times, which makes up most of the compilation time when compiling from pysb. 

Would be nice to simplify them to a single function such that they only need to compiled once.

FFroehlich:


FFroehlich:
I hope the "only" thing still missing is implementation of sparse blockwise transpose required for second order sensitivities.

FFroehlich:
> I hope the "only" thing still missing is implementation of sparse blockwise transpose required for second order adjoint sensitivities.

FFroehlich:
image

image

include/amici/sundials_matrix_wrapper.h
```diff
@@ -252,6 +252,28 @@ class SUNMatrixWrapper {
                          sunindextype *w, gsl::span<realtype> x,
                          const sunindextype mark,
                          SUNMatrixWrapper *C, sunindextype nnz) const;
+    
+    /**
+     * @brief Computes transpose A' of sparse matrix A and writes is to the matrix C = alpha * A'
```
dweindl, line 257:
```suggestion
     * @brief Computes transpose A' of sparse matrix A and writes it to the matrix C = alpha * A'
```

include/amici/sundials_matrix_wrapper.h
```diff
@@ -252,6 +252,28 @@ class SUNMatrixWrapper {
...
+     *
+     * @param C output matrix (sparse or dense)
+     * @param alpha scalar multiplier
+     */
+    void transpose(SUNMatrix C, const realtype alpha) const;
```
dweindl, line 262: Don't we want to use `SUNMatrixWrapper`?

include/amici/sundials_matrix_wrapper.h
```diff
@@ -252,6 +252,28 @@ class SUNMatrixWrapper {
...
+    
+    /**
+     * @brief Writes a sparse matrix A to a dense matrix D.
+     *
+     * @param D dense output matrix
+     */
+    void to_dense(SUNMatrix D) const;
```
dweindl, line 269: Don't we want to use `SUNMatrixWrapper`?

dweindl (approved):
Neat cleanup :+1:

include/amici/sundials_matrix_wrapper.h
FFroehlich: We use this to directly write to the raw SUNMatrix that comes from sundials, so this needs to be SUNMatrix. Could add additional overload for `SUNMatrixWrapper`, but would currently be unused.

PR #1267, opened 2020-09-14
fix setting of maximum number of steps, fixes #726

### LINKED ISSUE #726: solver.setMaxSteps() is not working correctly
I managed to simulate a model past (as indicated by `rdata['numsteps'])` the specified number of maximal steps specified by `solver.setMaxSteps()` . Seems to be dependent on specifying simulation timepoints prior to to the timepoint where the maximum number of steps is reached. Current hypothesis is that only number of steps between timepoints in penalized and not total.

FFroehlich:


include/amici/solver.h
```diff
@@ -1130,6 +1130,7 @@ class Solver {
      * problem
      *
      * @param mxsteps number of steps
+     * @note in contrast to the SUNDIALS method, this sets the overall maximu, not the maximum between output times.
```
dweindl, line 1133:
```suggestion
     * @note in contrast to the SUNDIALS method, this sets the overall maximum, not the maximum between output times.
```

dweindl (approved):
Looks good. Want to address `maxstepsB_` here as well?

src/solver.cpp
```diff
@@ -36,10 +36,20 @@ Solver::Solver(const Solver &other)
       maxstepsB_(other.maxstepsB_), sensi_(other.sensi_)
 {}
 
+void Solver::apply_max_num_steps() const {
+    // set remaining steps, setMaxNumSteps only applies to a single call of solve
+    long int cursteps;
+    getNumSteps(solver_memory_.get(), &cursteps);
+    setMaxNumSteps(maxsteps_-cursteps);
```
dweindl: Wait, we need to be careful here: in the (probably rare) case, where the solver just reaches `tout` with `maxsteps` steps, this passes `0` to SUNDIALS, which is interpreted as default value (500). In the following step, this will get negative which is interpreted as "no limit".

FFroehlich:
> Looks good. Want to address `maxstepsB_` here as well?

👍  done.

src/solver.cpp
dweindl: Corner-case, but that would be fun to debug :D

FFroehlich: good catch!

src/solver.cpp
```diff
@@ -36,10 +36,39 @@ Solver::Solver(const Solver &other)
       maxstepsB_(other.maxstepsB_), sensi_(other.sensi_)
 {}
 
+void Solver::apply_max_num_steps() const {
+    // set remaining steps, setMaxNumSteps only applies to a single call of solve
+    long int cursteps;
+    getNumSteps(solver_memory_.get(), &cursteps);
+    if (maxsteps_ <= cursteps)
+        throw AmiException("Reached maximum number of steps %i before reaching "
+                           "tout at t=%d.", maxsteps_, t_);
```
dweindl, line 45:
```suggestion
                           "tout at t=%g.", maxsteps_, t_);
```

src/solver.cpp
```diff
@@ -36,10 +36,39 @@ Solver::Solver(const Solver &other)
...
+void Solver::apply_max_num_steps_B() const {
+    // set remaining steps, setMaxNumSteps only applies to a single call of solve
+    long int curstepsB;
+    auto maxstepsB = (maxstepsB_ == 0) ? maxsteps_ * 100 : maxstepsB_;
+    for (int i_mem_b = 0; i_mem_b < (int)solver_memory_B_.size(); ++i_mem_b) {
+        if (solver_memory_B_.at(i_mem_b)) {
+            getNumSteps(solver_memory_B_.at(i_mem_b).get(), &curstepsB);
+            if (maxstepsB <= curstepsB)
+                throw AmiException("Reached maximum number of steps %i before "
+                                   "reaching tout at t=%d     in backward "
```
dweindl, line 58:
```suggestion
                                   "reaching tout at t=%g in backward "
```

src/solver.cpp
```diff
@@ -36,10 +36,39 @@ Solver::Solver(const Solver &other)
...
+
+void Solver::apply_max_num_steps_B() const {
+    // set remaining steps, setMaxNumSteps only applies to a single call of solve
+    long int curstepsB;
+    auto maxstepsB = (maxstepsB_ == 0) ? maxsteps_ * 100 : maxstepsB_;
+    for (int i_mem_b = 0; i_mem_b < (int)solver_memory_B_.size(); ++i_mem_b) {
+        if (solver_memory_B_.at(i_mem_b)) {
+            getNumSteps(solver_memory_B_.at(i_mem_b).get(), &curstepsB);
+            if (maxstepsB <= curstepsB)
+                throw AmiException("Reached maximum number of steps %i before "
```
dweindl, line 57:
```suggestion
                throw AmiException("Reached maximum number of steps %ld before "
```

src/solver.cpp
```diff
@@ -36,10 +36,39 @@ Solver::Solver(const Solver &other)
       maxstepsB_(other.maxstepsB_), sensi_(other.sensi_)
 {}
 
+void Solver::apply_max_num_steps() const {
+    // set remaining steps, setMaxNumSteps only applies to a single call of solve
+    long int cursteps;
+    getNumSteps(solver_memory_.get(), &cursteps);
+    if (maxsteps_ <= cursteps)
+        throw AmiException("Reached maximum number of steps %i before reaching "
```
dweindl, line 44:
```suggestion
        throw AmiException("Reached maximum number of steps %ld before reaching "
```

dweindl (approved):

Issue #1279, opened 2020-10-06
Installing amici package on Windows failed


feddersmichael:
I wanted to install the package by following the instructions from [here](https://github.com/AMICI-dev/AMICI/blob/master/INSTALL.md#windows-using-gcc-mingw) and [here](https://github.com/AMICI-dev/AMICI/issues/425).

Unfortunately I received an error after executing the very last comand 

`pip install --global-option="build_clib" --global-option="--compiler=mingw32" \ --global-option="build_ext" --global-option="--compiler=mingw32" \ amici --no-cache-dir --verbose`

I followed every step given, only exception is that I used the SWIG version 4.0.2 and also changed the PATH variable accordingly.
To not overload the box I copied the error message into a seperate file. Bachelor-Thesis is the name of my anaconda environment. I am very welcome for any help and ofcours willing to clarify my setup if necessary.

[issue.pdf](https://github.com/AMICI-dev/AMICI/files/5334198/issue.pdf)

FFroehlich:
This might be a formatting error for the installation instructions.

Does 

`pip install --global-option="build_clib" --global-option="--compiler=mingw32"  --global-option="build_ext" --global-option="--compiler=mingw32" amici --no-cache-dir --verbose`

work for you?

feddersmichael:
It did run through withoiut an error so thanks for your adjusted code. I still don't see amici as an installed package under my Bachelor-Thesis environment though, so there is still an issue or am I getting something wrong here?

PR #1281, opened 2020-10-06
Update INSTALL.md

FFroehlich:
follow up on #1279


dweindl (commented):
Let's wait for the RTD build, but I think this will mess up copy and pasting... Not sure why the `develop` version would not work. Windows `cmd.exe` issue? :thinking:

dweindl (approved):
Ah, sorry. This is the separate MD file, not RTD. Need to clean that up...

PR #1282, opened 2020-10-06
Add python 3.9 tests

FFroehlich:
3.9 was released today:

https://docs.python.org/3.9/whatsnew/3.9.html

### DIFF
--- a/.github/workflows/test_python_ver_matrix.yml
+++ b/.github/workflows/test_python_ver_matrix.yml
@@ -20,7 +20,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


dweindl (approved):

dweindl:
Our agreement was supporting the two most recent minor versions of Python, right? Can we allowing failing 
builds for 3.6, 3.7? (Not necessary to address here)

dweindl:
> Our agreement was supporting the two most recent stable minor versions of Python, right? Can we allowing failing
> builds for 3.6, 3.7? (Not necessary to address here)

Ah, yes, all entries listed separately in the branch protection settings.

PR #1283, opened 2020-10-06
Feature synthetic data

dilpath:
Ideas for tests?

To use this, the `feature_synthetic_data` branch of AMICI and PEtab (and pyPESTO for the example) should be used together.
Implementation in AMICI: [PR](https://github.com/AMICI-dev/AMICI/pull/1283)
Example optimisation with synthetic data: [notebook](https://github.com/ICB-DCM/pyPESTO/blob/feature_synthetic_data/doc/example/synthetic_data.ipynb), [PR](https://github.com/ICB-DCM/pyPESTO/pull/482)
PEtab base class: [code](https://github.com/PEtab-dev/PEtab/blob/feature_synthetic_data/petab/simulate.py), [PR](https://github.com/PEtab-dev/PEtab/pull/472)


paulstapor:
> Ideas for tests?

Given a random seed, the synthetic data should reproducible, right? At least, if noise is added to the simulation, it should be the same...
Beyond that, one could store a reference trajectory for a test model, simulate, and test the residuals to the reference via a chi2 test. Would be a probabilistic test. But one could run the same simulation 100 times and see, whether or not the overall chi2-value-distribution over the runs makes sense...

python/amici/petab_simulate.py
```diff
@@ -0,0 +1,75 @@
+from amici import AmiciModel
+from amici.petab_import import import_petab_problem
+from amici.petab_objective import simulate_petab, rdatas_to_measurement_df
+from petab.simulate import Simulator
+
+
+class PetabSimulator(Simulator):
+    """
```
dweindl, line 8: Add class description

python/amici/petab_simulate.py
```diff
@@ -0,0 +1,75 @@
...
+    amici_model:
+        Instance of an `amici.amici.ModelPtr` object (TODO confirm).
```
dweindl, line 10: Redundant or conflicting with typehint. "Model to simulate" ?

python/amici/petab_simulate.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    def _simulate_without_noise(
+            self,
+            amici_model: AmiciModel = None,
+            **kwargs
+    ):
+        """
+        See `petab.simulate.Simulator.simulate()` docstring.
```
dweindl, line 18:
```suggestion
        See :py:func:`petab.simulate.Simulator.simulate()` docstring.
```

python/amici/petab_simulate.py
FFroehlich, line 10:
```suggestion
        Amici model instance.
```

dweindl (commented):
The PEtab test problem here is from the test suite? The respective repo is cloned for the other PEtab test suite runs already. Would be great to use the data from there instead of adding the files here.

python/tests/petab_test_models/conversion_reaction/measurements.tsv
```diff
@@ -0,0 +1,11 @@
+observableId	simulationConditionId	time	measurement	noiseParameters
```
FFroehlich, line 1: gitignore

python/tests/petab_test_models/conversion_reaction/observables.tsv
```diff
@@ -0,0 +1,2 @@
+observableId	observableFormula	noiseFormula
```
FFroehlich, line 1: gitignore

python/tests/petab_test_models/conversion_reaction/parameters.tsv
```diff
@@ -0,0 +1,3 @@
+parameterId	parameterScale	lowerBound	upperBound	nominalValue	estimate
```
FFroehlich, line 1: gitignore

FFroehlich:
> The PEtab test problem here is from the test suite? The respective repo is cloned for the other PEtab test suite runs already. Would be great to use the data from there instead of adding the files here.

agreed

python/amici/petab_simulate.py
```diff
@@ -0,0 +1,75 @@
...
+        result = simulate_petab(
+            self.petab_problem,
+            self.amici_model,
+            **{k: v
+               for k, v in kwargs.items()
+               if k in relevant_kwargs_simulate},
+        )
+
+        # TODO use `rdatas_to_simulation_df` instead?
+        simulation_df = rdatas_to_measurement_df(
```
FFroehlich, line 69:
```suggestion
        measurement_df = rdatas_to_measurement_df(
```
as this function is supposed to generate data, not simulations we should stick to `measurement_df` here.

FFroehlich:
> > Ideas for tests?
> 
> Given a random seed, the synthetic data should reproducible, right? At least, if noise is added to the simulation, it should be the same...
> Beyond that, one could store a reference trajectory for a test model, simulate, and test the residuals to the reference via a chi2 test. Would be a probabilistic test. But one could run the same simulation 100 times and see, whether or not the overall chi2-value-distribution over the runs makes sense...

I agree this test makes sense, but I don't think it should be implemented in amici since amici doesn't implement the noise generation (should probably be done in pypesto?).

I think tests here should focus on repeated simulations and passing of kwargs to the right function.

paulstapor:
> I agree this test makes sense, but I don't think it should be implemented in amici since amici doesn't implement the noise generation (should probably be done in pypesto?).

Wasn't aware which package does actually what in this new feature.
The test obviously makes most sense to be implemented in the package which finally generates the artificial data (and the noise)...
If this generation is done by pyPESTO, then obviously pyPESTO is the better choice, fully agreed.

dilpath:
Thanks for the feedback. There are now tests for repeated calls without noise, and handling of `kwargs`.

Tests on noise generation have been added to the [PEtab PR](https://github.com/PEtab-dev/PEtab/pull/472).

Also switched to using the already cloned PEtab test suite (won't check this is working until the PEtab PR is merged).

dilpath:
Now ready for review (the PEtab version in this PR is `develop`, and will be changed to the next PEtab PyPI version once the next PEtab version is released, before this PR is merged).

dweindl (approved):

dweindl:
> Now ready for review (the PEtab version in this PR is `develop`, and will be changed to the next PEtab PyPI version once the next PEtab version is released, before this PR is merged).

petab 0.1.11 is now available on pypi

python/sdist/setup.py
```diff
@@ -209,7 +209,7 @@ def main():
         setup_requires=['setuptools>=40.6.3'],
         python_requires='>=3.6',
         extras_require={
-            'petab': ['petab==0.1.7'],
+            'petab': ['petab==0.1.11'],
```
dweindl, line 212:
```suggestion
            'petab': ['petab>=0.1.11'],
```

Would only set a minimum version requirement (assuming no major backwards incompatibilities in the near future...)

FFroehlich (approved):

Issue #1149, opened 2020-07-13
Refactor generation of symbolic expressions during SBML import


dweindl:
As mentioned in https://github.com/ICB-DCM/AMICI/pull/1145#issuecomment-653117048, it would be good to refactor SBML import to create symbolic expressions for model entities in a central place and avoid to repeatedly iterate over SBML entities, which is somewhat slow for large models (1). Should clean up the code and speed up import a little. 

(1)
```
2020-07-13 12:57:17.395 - amici.sbml_import - INFO - Finished processing SBML parameters          + (5.52E-01s)
2020-07-13 12:57:17.397 - amici.sbml_import - INFO - Finished processing SBML compartments        + (2.49E-03s)
2020-07-13 12:57:29.107 - amici.sbml_import - INFO - Finished processing SBML species             + (1.17E+01s)
2020-07-13 13:18:11.813 - amici.sbml_import - INFO - Finished processing SBML reactions           + (1.24E+03s) <----
2020-07-13 13:19:52.319 - amici.sbml_import - INFO - Finished processing SBML rules               + (1.00E+02s) <----
2020-07-13 13:20:03.240 - amici.sbml_import - INFO - Finished processing SBML observables         + (1.03E+01s)
```

FFroehlich:
High processing time for reactions is also one of the main reason why we have a seperate import for pysb. If that goes aways we could look into removing that.

dweindl:
@dilpath : assigned you, as you mentioned you were working on this. if wrong, just unassign yourself.

PR #1284, opened 2020-10-06
refactor sbml

FFroehlich:
refactor sbml import, no logic changes


python/amici/sbml_import.py
```diff
@@ -1438,8 +1344,8 @@ def _replace_in_all_expressions(self,
         """
         # Avoid replacing variables with rates
         if include_rate_rule_targets \
-                or old not in set((*self.compartment_rate_rules.keys(),
-                           *self.species_rate_rules.keys())):
+                or old not in {*self.compartment_rate_rules.keys(),
+                               *self.species_rate_rules.keys()}:
```
dilpath, line 1347: Can also be without `.keys()` i.e. `{*self.compartment_rate_rules, *self.species_rate_rules}`

FFroehlich: 👍

FFroehlich:
ended up needing to change some logic after all ...
- adds support for initial parameter assignments
- refactor observable code such that validation of tests is simpler and comprehensive 
- extend generated locals for sympification

tests/testSBMLSuite.py
```diff
@@ -73,93 +72,68 @@ def test_sbml_testsuite_case(test_number, result_path):
...
+        ] and species in [str(s) for s in wrapper.species_rate_rules]
+    ]
+    amounts_to_concentrations(concentration_species, wrapper,
+                              simulated, requested_concentrations)
+
+    concentrations_to_amounts(amount_species, wrapper, simulated,
+                              requested_concentrations)
+    for variable in variables:
+        # some values in expected will be int64, so we cast everything as float
+        # to avoid errors from downcasting float64 to int64
```
dweindl, line 124: Not converted automatically? :flushed:

dweindl (approved):
Looks good!

Closes #1149, I'd say.

dweindl:
Oh, can handle an additional 25 test cases now. :+1: :rocket:

tests/testSBMLSuite.py
FFroehlich: simulated has dtype `object` for which np.isclose doesnt work and expected is mixed `np.int64`, `np.float64` so you have to explicitely cast. Not sure what happens when comparing `int64` vs `float64`, but that isn't easy to implement, but comment may indeed be a bit misleading.

python/amici/sbml_import.py
```diff
@@ -1295,16 +1171,102 @@ def replace_assignments(formula: str) -> sp.Basic:
...
+                          for ass in self.sbml.getListOfInitialAssignments()]
+        rulevars = [rule.getVariable()
+                    for rule in self.sbml.getListOfRules()
+                    if rule.getFormula() != '']
+        # doesnt look like there is a better way to get hold of those lists:
+        species_references = next((
+            list(element)
+            for element in self.sbml.all_elements
+            if isinstance(element, sbml.ListOfSpeciesReferences)
+        ), [])
```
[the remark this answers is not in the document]
FFroehlich: :+1:

FFroehlich:
> Looks good!
> 
> Closes #1149, I'd say.

I would argue that there still is some repeated processing, would hope this can be addressed in a couple of additional PRs.

python/amici/sbml_import.py
```diff
@@ -2020,6 +1922,131 @@ def _get_str_symbol_identifiers(str_symbol: str) -> tuple:
...
+            conc = conc_from_conc
+
+    else:
+        if species.isSetInitialConcentration():
+            conc = conc_from_conc
+
+        if species.isSetInitialAmount() and not math.isnan(amount):
+            conc = conc_from_amount
+
+    return conc
```
dilpath, line 2027:
```suggestion
    if species.isSetInitialConcentration():
        return sp.sympify(species.getInitialConcentration())
    
    if species.isSetInitialAmount():
        amt = species.getInitialAmount()
        if not math.isnan(amt):
            return sp.sympify(amt) / _get_species_compartment_symbol(species)

    return _get_identifier_symbol(species)
```

Not sure I got the logic, but the code that is evaluated for both conditions of the `if-else` statement looks the same to me? Also, SBML spec says `initialAmount` and `initialConcentration` should not both be specified, so order shouldn't matter right?

dilpath (approved):
Nice!!

python/amici/sbml_import.py
```diff
@@ -1295,16 +1169,99 @@ def replace_assignments(formula: str) -> sp.Basic:
...
+        """
+        Transforms an expression to its value at the initial timepoint by
+        replacing species by their initial values.
+
+        :param sym_math:
+            symbolic expression
+        :return:
+            transformed expression
+        """
+        if not isinstance(sym_math, sp.Expr):
```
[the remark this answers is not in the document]
FFroehlich: Correct, but `if not isinstance(sym_math, sp.Expr)` is more forgiving if we happen to pass a float instead.

python/amici/sbml_import.py
```diff
@@ -111,6 +119,23 @@ class SbmlImporter:
         model symbols for sympy to consider during sympification
         see `locals`argument in `sympy.sympify`
 
+    :ivar species_assignment_rules:
```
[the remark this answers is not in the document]
FFroehlich: 👍

python/amici/sbml_import.py
FFroehlich: Yes, was also wondering why this code was necessary, and I think I also tried to simplify but it didn't work, which was probably due to other issues. Just tried locally and eveything passed. Thanks!

PR #1304, opened 2020-10-14
make stoichiometries and reaction ids local symbols

FFroehlich:


python/amici/ode_export.py
```diff
@@ -3228,3 +3228,17 @@ def generate_measurement_symbol(observable_id: Union[str, sp.Symbol]):
     if not isinstance(observable_id, str):
         observable_id = strip_pysb(observable_id)
     return sp.Symbol(f'm{observable_id}', real=True)
+
+
+def generate_flux_symbol(reaction_index: int) -> sp.Symbol:
```
dweindl, line 3233:
```suggestion
def generate_flux_symbol(reaction_index: int) -> sp.Symbol:
```

python/amici/sbml_import.py
```diff
@@ -473,15 +475,26 @@ def _gather_locals(self) -> None:
...
-
         # SBML time symbol + constants
         self.local_symbols['time'] = sp.Symbol('time', real=True)
         self.local_symbols['avogadro'] = sp.Symbol('avogadro', real=True)
 
+        # definitions below already use self.local_symbols and shouldn't be
+        # reordered
+        for r in self.sbml.getListOfReactions():
+            for e in list(r.getListOfReactants()) + \
+                     list(r.getListOfProducts()):
```
dweindl, line 486:
```suggestion
            for e in chain(r.getListOfReactants(), r.getListOfProducts()):
```
`from itertools import chain`
?

dweindl (approved):

python/amici/sbml_import.py
FFroehlich: 👍

FFroehlich:
slightly increased limit for model import since there seems to be a slowdown in generation of cpp code although nothing on that end changed, so I assume there is some performance degradation on the GHA end.

PR #1305, opened 2020-10-15
Steadystate export

paulstapor:
As this has been asked for by people working with some large-scale models (and will make sense for people working with kinetic models of metabolism):
Here's some little script, which exports steady states to a dataframe. Same for fluxes will follow.


FFroehlich (changes requested):
Isn't this duplicating the functionality of https://github.com/AMICI-dev/AMICI/blob/d16078e8b6ac8b4ae31d4e7ff01738bc0291ccdd/python/amici/pandas.py#L171
?

Also, in the interest of modularity, I think features that do not explicitly depend on petab should be implemented independently of petab, i.e., this function should be built around amici.runAmiciSimulations, not simulate_petab. You could then add a wrapper that makes this function more accessible for petab problems.

paulstapor:
> Isn't this duplicating the functionality of
> 
> https://github.com/AMICI-dev/AMICI/blob/d16078e8b6ac8b4ae31d4e7ff01738bc0291ccdd/python/amici/pandas.py#L171
> 
> ?
> 
> Also, in the interest of modularity, I think features that do not explicitly depend on petab should be implemented independently of petab, i.e., this function should be built around amici.runAmiciSimulations, not simulate_petab. You could then add a wrapper that makes this function more accessible for petab problems.

I would say that most of the functionality here is due to checking which conditions from a given PEtab problem have steady state measurements, and accounting for that in the case of user-specified inputs. It's not really duplicating things, except for maybe 5 lines. Also, this routine simulates, which getSimulationStatesAsDataFrame doesn't do. Hoewever, one can argue about whether it should simulate or not. So this remains open I would say... I would vote *pro* simulation, as this allows to pass a simple PEtab problem and not do any simulations which won't be used in the end...

I integrated it with simulate_petab as the main use case will be PEtab-based. However, I can also integrate it with runSimulations(). That would mostly mean *moving* the code, however, not removing it... In any case, changing it to work with `runSimulations()` would make it usable also without a PEtab problem...

I'm fine with any code change requests...

paulstapor:
One more question on "how-to-unittest":
The PEtab test suite is not by default installed before running unit tests (probably because this would make unit tests take longer).
However, the steady state export needs a PEtab test case to run... Shall I create a new PEtab test case based on one of the python examples, or rather use the one I use now from the PEtab test suite? What would be the best way to have an easy test case for this?

FFroehlich:
> One more question on "how-to-unittest":
> The PEtab test suite is not by default installed before running unit tests (probably because this would make unit tests take longer).
> However, the steady state export needs a PEtab test case to run... Shall I create a new PEtab test case based on one of the python examples, or rather use the one I use now from the PEtab test suite? What would be the best way to have an easy test case for this?

Well I guess this nicely highlights the underlying problem. This function does not implement anything related to the PEtab spec, so I don't really see why it should be implemented around simulate_petab.

Two options:
- This is implemented for runAmiciSimulations and there is some way that makes it simple to interface if from simulate_petab, then I think its fine to have this functionality in AMICI
- This is implemented only for petab, in which case it probably makes more sense to implement it for https://github.com/PEtab-dev/PEtab/blob/master/petab/simulate.py#L12 and AMICI is probably not the best place to harbor this function

paulstapor:
> Two options:
> 
>     * This is implemented for runAmiciSimulations and there is some way that makes it simple to interface if from simulate_petab, then I think its fine to have this functionality in AMICI
> 
>     * This is implemented only for petab, in which case it probably makes more sense to implement it for https://github.com/PEtab-dev/PEtab/blob/master/petab/simulate.py#L12 and AMICI is probably not the best place to harbor this function

Good point. Not yet sure... Need to think about it. Might add it to AMICI, and keep a simple script as example.

I might however also add this in a slightly altered form to the project specifc repo or even to my personal github repo... 
I see your point: If we add all functionality to AMICI that somebody might want to have at some point, AMICI will just be littered with many corner-case functions which are rarely used. 
This would make it hard to keep AMICI in a good state, as unit testing all functionality of AMICI already takes long and would take way too long if we inflate the code base with such functions...
And I don't want that...

It might be good to keep this separate from AMICI, just as yaml2sbml. Maybe that's the easiest way and yet the best for the AMICI code base. Might be then meaningful to have a list of repos with some convenience or helper functions, which can be used with AMICI, but are not part of it... What do you think?

FFroehlich:
> > Two options:
> > ```
> > * This is implemented for runAmiciSimulations and there is some way that makes it simple to interface if from simulate_petab, then I think its fine to have this functionality in AMICI
> > 
> > * This is implemented only for petab, in which case it probably makes more sense to implement it for https://github.com/PEtab-dev/PEtab/blob/master/petab/simulate.py#L12 and AMICI is probably not the best place to harbor this function
> > ```
> 
> Good point. Not yet sure... Need to think about it. Might add it to AMICI, and keep a simple script as example.
> 
> I might however also add this in a slightly altered form to the project specifc repo or even to my personal github repo...
> I see your point: If we add all functionality to AMICI that somebody might want to have at some point, AMICI will just be littered with many corner-case functions which are rarely used.
> This would make it hard to keep AMICI in a good state, as unit testing all functionality of AMICI already takes long and would take way too long if we inflate the code base with such functions...
> And I don't want that...
> 
> It might be good to keep this separate from AMICI, just as yaml2sbml. Maybe that's the easiest way and yet the best for the AMICI code base. Might be then meaningful to have a list of repos with some convenience or helper functions, which can be used with AMICI, but are not part of it... What do you think?

Correct. However, I think it actually makes sense to have this functionality, as extracting the steadystate of a systems might be something that more than the one-odd user is interested in. From my perspective, the tipping point is whether this functionality is limited to petab only, or also applicable to other ways of simulating the system. Say for example the pysb simulator PR is finally approved. Adding steadystate extraction to that simulator could be a worthwhile improvement, but with the functionality limited to simulate_petab, this wouldn't be possible easily.

paulstapor:
Will try to integrate this this week within `amici.runSimulations()` and maybe add an example script for the PEtab case...

python/amici/petab_objective.py
```diff
@@ -76,6 +77,8 @@ def simulate_petab(
         are assumed to be in linear scale.
     :param log_level:
         Log level, see :mod:`amici.logging` module.
+    :num_threads:
+        Number of cores to be used for simulations when multi-threading
```
dweindl, line 81:
```suggestion
        Number of threads to be used for simulations if multi-threading is enabled
```

python/amici/petab_objective.py
```diff
@@ -155,6 +159,114 @@ def simulate_petab(
...
+    # get the steady states (postequilibration only, use last timepoint)
+    steady_states = np.array([rdata['x'][-1,:]
+                              for rdata in sim_results['rdatas']])
+    # create dataframe
+    steady_state_df = pd.DataFrame(columns=amici_model.getStateIds(),
+        index=condition_list, data=steady_states)
+
+    # export to file (if requested)
+    if export_file is not None:
+        steady_state_df.to_csv(export_file, sep=sep, index=False)
```
dweindl, line 265: I'd leave that out of this function. Adds unnecessary complexity to function body and signature. Saving it outside is not that big a deal.

python/amici/petab_objective.py
```diff
@@ -155,6 +159,114 @@ def simulate_petab(
...
+                          if any(np.isinf(edata.getTimepoints()))]
+        # filter edata objects for steadystate simulations
+        edatas = [edata for edata in edatas
+                  if any(np.isinf(edata.getTimepoints()))]
+
+    if not condition_list:
+        if export_file is not None:
+            logging.warning('No output file could be written, as no matching '
+                            'preequilibration and simulation condition with '
+                            'steady state data was found.')
```
dweindl, line 240: If you return an empty DataFrame it would be more coherent to also write that to file. But preferred to move file I/O out of here (see below).

python/amici/petab_objective.py
```diff
@@ -155,6 +159,114 @@ def simulate_petab(
     }
 
 
+def export_steadystates(
```
dweindl, line 162: why "export"? would prefer get/compute/...

python/amici/petab_objective.py
```diff
@@ -628,6 +740,106 @@ def _get_measurements_and_sigmas(
...
+                                  steadystate_conditions: pd.DataFrame,
+    ) -> pd.DataFrame:
+    """
+    Filter simulation conditions passed by user for steadystate data
+
+    :param simulation_conditions:
+        dataframe with simulation conditions passed by user
+
+    :param steadystate_conditions:
+        ddataframe with simulation conditions with all steadystate measurements
```
dweindl, line 791:
```suggestion
        Dataframe with simulation conditions with all steadystate measurements
```

python/amici/petab_objective.py
```diff
@@ -155,6 +159,114 @@ def simulate_petab(
...
+        simulation_conditions: Union[pd.DataFrame, Dict] = None,
+        edatas: List[AmiciExpData] = None,
+        parameter_mapping: ParameterMapping = None,
+        scaled_parameters: Optional[bool] = False,
+        log_level: int = logging.WARNING,
+        num_threads: int = 1,
+        export_file: str = None,
+        sep: str = '\t'
+) -> pd.DataFrame:
+    """Simulate PEtab model.
```
dweindl, line 176: This could be made more precise...

python/amici/petab_objective.py
```diff
@@ -155,6 +159,114 @@ def simulate_petab(
...
+    :export_file:
+        Path to (csv) file to which the steady state value should be written
+    :sep:
+        Separator for csv file for export
+    :return:
+        Pandas Dataframe containing
+        * the names of the state variables as row indices,
+        * the indices of the amici edata objects as columns,
+          (**NOTE**: only edatas with timepoints INF will be considered)
+        * the steady state values of the state variables as table content`.
```
dweindl, line 213:
```suggestion
        * the steady state values of the state variables as table content.
```

python/amici/petab_objective.py
```diff
@@ -155,6 +159,114 @@ def simulate_petab(
...
+    """
+
+    # if steady state measurements are available: get the simulation conditions
+    steadystate_conditions = _find_steadystate_conditions(petab_problem)
+
+    if simulation_conditions is not None:
+        # filter simulation conditions for steadystate simulations
+        simulation_conditions = _filter_simulation_conditions(
+            simulation_conditions, steadystate_conditions)
```
dweindl, line 222: I am not completely sure if understand the overall purpose of the function, but why not provide steadystates for all simulation conditions, independently of whether there are steadystate measurements?

python/amici/petab_objective.py
```diff
@@ -628,6 +740,106 @@ def _get_measurements_and_sigmas(
...
+def _create_condition_list(simulation_conditions: pd.DataFrame) -> List[str]:
+    """
+    Create a list identifying the preequilibration and simulation conditions
+
+    :param simulation_conditions:
+        dataframe with simulation conditions passed by user
+
+    :return:
+        list with the preequilibration and simulation conditions
+    """
```
dweindl, line 828: Hm. Without reading the code, it's not really clear to me what goes in and what comes out. Required / provided columns, etc.

paulstapor:
I think I will take this branch down next week and make a little independent thing out of it (which is completely satisfactory for its current purpose).
Many thanks for the code revision! Maybe I will integrate this into AMICI in a midterm future in the way suggested by @FFroehlich . But right now, there are more urgent things and I don't see myself spending some hours on this in the next month(s). Heavisides and events are more important, and the Python splines as well.
If nobody objects, I will take this PR down, as I don't really see a good reason for having a PR open which won't be merged in the next month...

PR #1306, opened 2020-10-16
Fix namespace when replacing pow by amici::pos_pow

dweindl:
Probably broke during change to C++ codeprinter


paulstapor (approved):
Fine if tests pass

FFroehlich:
> Fine if tests pass

Well I guess we should add a test to make sure this doesn't break in the future ;)

Issue #1139, opened 2020-06-21
SymPy's `sympify` and floating point arithmetic


lcontento:
AMICI uses SymPy for symbolic computation and mathematical formulas from SBML files are converted to strings and loaded with `sympy.sympify`. This function does not produce a faithful representation of the given string, but applies some transformations. Some of them are very reasonable, such as `sympy.sympify('x - x')` returning zero. Others are quite problematic. For example, consider the following expression for a sigmoid function:
```python
sympy.sympify('1 / (1 + exp(800.0 - x))')
```
While the original formula can be evaluated numerically without any problem, `sympify` splits the exponential in `exp(800.0) * exp(-x)`. SymPy has arbitrary precision arithmetic so it has no trouble with the very large number `exp(800.0)`, but the resulting literal is interpreted as `inf` in floating point arithmetic (and completely rejected by `libsbml` MathML parser).

SymPy has an `evaluate=False` option for `sympify` to reduce the number of simplifications it applies. However, this is at the moment not propagated to functions like `exp` but only to the basic arithmetic operations. Using the `evaluate` context manager (marked as experimental though) is the way to go at the moment:
```python
import sympy
from sympy.core.parameters import evaluate
with evaluate(False):
    print(sympy.sympify('1 / (1 + exp(800.0 - x))'))
```

Personally, I would convert all instances of `sympy.sympify` reading in user-written expressions to a custom one disabling evaluation. This should not be a difficult or particularly dangerous change (I think). We may lose some small opportunity for simplification, but in general I believe user-written code should be followed as accurately as possible and their intentions not second-guessed.

As a final note, I guess similar floating point arithmetic mistakes may occur when creating the code for the derivatives. I am not sure whether such errors would be possible to avoid without ditching SymPy completely, but hopefully they are not common enough to worry...

FFroehlich:
I would argue that the anticipated result depends on context. In particular, when programmatically generating model code, we generally do want simplifications. We also want simplifications for larger models where size of generated code can become an issue. I do see the point though that simplifications are not always wanted.

We already have the infrastructure set up to perform simplification, so I don't see any issues with changing the code such that no simplification are performed by default by using `with evaluate(False):` as you proposed. Yet, I would keep something like `sp.sympify` as default for simplification. Completely avoiding simplification could be achieved by passing `lambda x:x` as simplification function.

How does that sound?

JanHasenauer:
[...]

PR #1315, opened 2020-10-21
Refactor SBML rule processing

FFroehlich:
No longer flattens sbml rule assignment. uses toposort to obtain correct ordering of interdependent assignments. Turns out we were still simulating species that had assignment rules, which is now fixed (which includes a fix for model without species).


FFroehlich:
Note that this doesn't really solve https://github.com/AMICI-dev/AMICI/issues/1139, should be remove from description when merging.

dilpath (approved):
:+1:

python/amici/sbml_import.py
```diff
@@ -974,47 +908,72 @@ def replace_assignments(formula: str) -> sp.Expr:
...
+        }
 
-            # Assignment rules take precedence over compartment volume
-            # definitions, so they need to be evaluated first.
-            # Species assignment rules always overwrite
-            for variable, formula in (
+        # Assignment rules take precedence over compartment volume
+        # definitions, so they need to be evaluated first.
+        # Species assignment rules always overwrite
+        for variable, formula in (
```
dweindl, line 939:
```suggestion
        for variable, formula in itt.chain(
```
?

python/amici/sbml_import.py
```diff
@@ -1270,45 +1219,41 @@ def _replace_special_constants(self) -> None:
...
-                        for symbol in self.symbols.keys()
-                        if 'identifier' in self.symbols[symbol].keys()]):
-                self._replace_in_all_expressions(constant, value)
-            else:
-                # yes sbml supports this but we wont. Are you really expecting
-                # to be saved if you are trying to shoot yourself in the foot?
-                raise SBMLException(
-                    f'Encountered currently unsupported element id {constant}!'
-                )
+            self._replace_in_all_expressions(constant, value)
```
dweindl, line 1222: So far it's only one constant, but in general might be faster for `_replace_in_all_expressions` to handle multiple (independent) substitutions at once.

python/amici/sbml_import.py
```diff
@@ -1330,68 +1275,37 @@ def _get_element_from_assignment(self, element_id: str) -> sp.Expr:
...
+                sym = self._sympy_from_sbml_math(ele.getStoichiometry())
         elif ele.isSetStoichiometry():
-            sym = sp.sympify(ele.getStoichiometry(),
-                             locals=self.local_symbols)
+            sym = self._sympy_from_sbml_math(ele.getStoichiometry())
         else:
-            return sp.sympify(1.0)
-        sym = _parse_special_functions(sym)
-        _check_unsupported_functions(sym, 'Stoichiometry')
+            sym = sp.sympify(1.0)
```
dweindl, line 1301:
```suggestion
            sym = sp.Float(1.0)
```

dweindl (approved):
:+1:

python/amici/sbml_import.py
```diff
@@ -807,88 +790,57 @@ def _process_rules(self) -> None:
...
-                    assignments[assignment] = assignments[assignment].subs(
-                        sym_id, formula
-                    )
-
-        # do this at the very end to ensure we have flattened all recursive
-        # rules
-        for sym_id in assignments.keys():
-            self._replace_in_all_expressions(symbol_with_assumptions(sym_id),
-                                             assignments[sym_id])
+                        contains_rule_assignment = True
```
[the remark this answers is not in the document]
FFroehlich: No, while loop has to stay since we currently do not allow `w` to occur in `x0`.

python/amici/sbml_import.py
FFroehlich: True, also applies to other occurrences. Requires rewriting of both `self._replace_in_all_expressions` and `_make_initial`, so let's not do this here.

python/amici/sbml_import.py
```diff
@@ -1267,48 +1225,45 @@ def _replace_special_constants(self) -> None:
...
-                        for symbol in self.symbols.keys()
-                        if 'identifier' in self.symbols[symbol].keys()]):
-                self._replace_in_all_expressions(constant, value)
-            else:
-                # yes sbml supports this but we wont. Are you really expecting
-                # to be saved if you are trying to shoot yourself in the foot?
-                raise SBMLException(
-                    f'Encountered currently unsupported element id {constant}!'
-                )
+            self._replace_in_all_expressions(constants, value)
```
dilpath, line 1231:
```suggestion
            self._replace_in_all_expressions(constant, value)
```

tests/testSBMLSuite.py
```diff
@@ -169,21 +171,25 @@ def concentrations_to_amounts(
...
+        if not isinstance(wrapper.sbml.getElementBySId(species), sbml.Species):
+            continue
+
+        if species_id in wrapper.symbols[SymbolId.SPECIES]:
+            if wrapper.symbols[SymbolId.SPECIES][species_id]['amount']:
+                continue
+            comp = wrapper.symbols[SymbolId.SPECIES][species_id].get(
+                'compartment', None
+            )
+            if comp is None:
```
dilpath, line 187:
```suggestion
            # Compartments and parameters that are treated as species are not
            # contained within a compartment.
            if comp is None:
```

python/amici/sbml_import.py
```diff
@@ -452,8 +454,15 @@ def _gather_base_locals(self):
             if r.isSetId():
                 self.local_symbols[r.getId()] = _get_identifier_symbol(r)
 
+        # oo is sympy infinity
+        self.local_symbols['INF'] = sp.sympify('oo')
+        self.local_symbols['NaN'] = sp.nan
+
+        # rem function
+        self.local_symbols['rem'] = sp.Mod
```
dilpath, line 462: Should an exception be raised if these symbols exist in the model? e.g.:
```python
reserved = {
    'INF': sp.oo,
    'NaN': sp.nan,
    'rem': sp.Mod,
}

if reserved.keys() & list_of_sbml_ids:
    raise ...
# or if it would have already been added to local_symbols, can do
if reserved.keys() & local_symbols.keys():
    raise ...

local_symbols.update(reserved)
```

dilpath (approved):
:+1:

python/amici/sbml_import.py
```diff
@@ -1011,21 +1009,40 @@ def _process_initial_assignments(self):
...
+        while contains_ia:
+            contains_ia = False
+            for sym_id, sym_math in self.initial_assignments.items():
+                for s in sym_math.free_symbols:
+                    if s in self.initial_assignments:
+                        contains_ia = True
+                        sym_math = smart_subs(
+                            sym_math, s, self.initial_assignments[s]
+                        )
+                self.initial_assignments[sym_id] = sym_math
```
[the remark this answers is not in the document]
FFroehlich: 👍

python/amici/sbml_import.py
FFroehlich: Good call, unfortunately SBML seems to indeed allow such silliness ...

python/amici/sbml_import.py
```diff
@@ -1226,9 +1248,10 @@ def _replace_special_constants(self) -> None:
         """
         constants = [
             (symbol_with_assumptions('avogadro'), sp.Float(6.02214179e23)),
+            (symbol_with_assumptions('exponentiale'), sp.E),
```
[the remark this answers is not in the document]
FFroehlich: Bit of a philosophical question. To my understanding, the SBML spec defines the number as exactly this value with not less or more precision:

> In other words, the value of this csymbol is equivalent to the following:
>(6.02214179 · 10E23) · dimensionless
> If the value of the constant is revised by international standards-setting organizations in the future, a future Version > of the SBML Level 3 specification may stipulate a new value to be used for this csymbol constant

Note that this definition is different than the SI definition used in sympy (6.02214076e23)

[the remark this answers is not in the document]
dilpath: Ah good point, should keep as is then :+1:

PR #1320, opened 2020-10-24
Monkeypatch sympy pow derivative

FFroehlich:


lcontento (commented):
I am sorry, but I cannot approve in any good faith the use of monkey patching inside a library. We cannot be 100% sure that doing so will not break another part of Sympy and we are going to change Sympy's behaviour for everyone else who is importing the package during the same session. Monkey patch by the end-user is sometimes unavoidable but at least the user knows that it happened. Doing it in a library without the user knowing about it does not sound a good idea to me.

A couple of possible alternative solutions I can think of at the moment (both would take a bit more work I am afraid):
* do the monkey patching inside a context manager.This will protect the user from the changes in Sympy, but does not protect us from possible behaviour broken by the monkey patch (probably fine since the monkey patch does no really big change)
* define a custom SymPy function that behaves exactly like `Pow` but has the custom derivative expression. Then, replace `Pow` with the custom version (using standard SymPy functions) before derivatives are taken.

FFroehlich:
Defined custom Pow function. Currently dysfunctional due to what I believe is a bug in sympy: https://github.com/sympy/sympy/issues/20339

python/amici/ode_export.py
```diff
@@ -44,6 +44,28 @@
...
+            # first piece never applies or is zero anyways
+            return part1 + part2
+
+        return part1 + sp.Piecewise(
+            (sp.sympify(0.0), sp.And(sp.Eq(self.base, 0), sp.Eq(dbase, 0))),
+            (part2, True)
+        )
+
+    def __repr__(self):
+        return 'AmiPow'
```
lcontento, line 66:
```suggestion
class CustomPow(sp.Function):
    nargs = (2, )  # function must take 2 arguments

    @classmethod
    def eval(cls, *args):
        return None  # means leave unevaluated

    def _eval_derivative(self, s):
        base, exp = self.args
        dbase = base.diff(s)
        dexp = exp.diff(s)
        part1 = sp.Pow(base, exp - 1) * exp * dbase
        part2 = self * dexp * sp.log(base)
        if dbase.is_nonzero or base.is_nonzero or dbase.is_nonzero or \
                part2.is_zero:
            # first piece never applies or is zero anyways
            return part1 + part2

        return part1 + sp.Piecewise(
            (sp.sympify(0), sp.And(sp.Eq(base, 0), sp.Eq(dbase, 0))),
            (part2, True)
        )
    
    def fdiff(self, argindex=1):
        raise Exception(
            'All derivative calculations should be routed through _eval_derivative()'
        )

    def _eval_is_real(self):
        return True
```

lcontento:
> Defined custom Pow function. Currently dysfunctional due to what I believe is a bug in sympy: [sympy/sympy#20339](https://github.com/sympy/sympy/issues/20339)

Unfortunately I don't think it is a bug... SymPy is just organized in a way that makes extending expressions/symbols impossible.

As Oscar mentioned, the best way is to implement a custom function/expression. One possibility is to copy and adapt the code for `Pow` from https://github.com/sympy/sympy/blob/master/sympy/core/power.py, but that is a bit longish. Another alternative is to extend `Function` instead of `Expr` (I have put a suggestion for a possible way to do it).

I am not sure which approach would be better (the latter is certainly more compact though). And I am afraid the resulting custom power function will not play nice with simplification/evaluation in either case. So it may be better to substitute it into symbolic expression just before taking the derivative and substituting it back in the derivative before the call to the simplification function.

python/amici/ode_export.py
FFroehlich: This would prevent all kinds of simplifications, such as `powsimp` from working.

FFroehlich:
> > Defined custom Pow function. Currently dysfunctional due to what I believe is a bug in sympy: [sympy/sympy#20339](https://github.com/sympy/sympy/issues/20339)
> 
> Unfortunately I don't think it is a bug... SymPy is just organized in a way that makes extending expressions/symbols impossible.

It's not a bug, its a feature!? Still think it's a bug, that isn't really straightforward to solve so I understand it if they won't fix it.

> As Oscar mentioned, the best way is to implement a custom function/expression. One possibility is to copy and adapt the code for `Pow` from https://github.com/sympy/sympy/blob/master/sympy/core/power.py, but that is a bit longish. Another alternative is to extend `Function` instead of `Expr` (I have put a suggestion for a possible way to do it).
> 
> I am not sure which approach would be better (the latter is certainly more compact though). And I am afraid the resulting custom power function will not play nice with simplification/evaluation in either case. So it may be better to substitute it into each symbolic expression just before taking the derivative and then substitute it back in the derivative before the call to the simplification function.

Replacing after taking the derivative is likely to be computationally prohibitively expensive.

So I guess the best way to make this work is monkeypatch sympy in context manager in `ODEModel.generate_model_code`

dweindl:
> So I guess the best way to make this work is monkeypatch sympy in context manager in ODEModel.generate_model_code

Sounds good to me.

lcontento:
> It's not a bug, its a feature!? Still think it's a bug, that isn't really straightforward to solve so I understand it if they won't fix it.

I don't think it is fair to call it a bug since the user is not expected to subclass from SympPy's functions and symbols: it is not *supposed* to work. But I wouldn't go as far as calling it a "feature" :wink:, rather a "limitation" of SymPy's current approach. I guess that in order to allow for such subclassing they would need to rewrite half of SymPy, so this won't change any time soon...

> Replacing after taking the derivative is likely to be computationally prohibitively expensive.

Replacements wrapped in a `evaluate=False` context are usually much faster, so I would not exclude it.

> So I guess the best way to make this work is monkeypatch sympy in context manager in `ODEModel.generate_model_code`

But this is probably good enough, so maybe we can start from here.

python/amici/ode_export.py
```diff
@@ -3314,6 +3320,7 @@ def cast_to_sym(value: Union[SupportsFloat, sp.Expr, BooleanAtom],
         raise TypeError(f"Couldn't cast {input_name} to sympy.Expr, was "
                         f"{type(value)}")
 
+    # use Pow function with custom derivative
```
dweindl, line 3323: hmm... that comment looks a bit lost here..

python/amici/ode_export.py
```diff
@@ -3344,3 +3351,54 @@ def smart_subs_dict(sym: sp.Expr,
         for eid, expr in subs.items()
         if eid in sym.free_symbols
     })
+
+
+# https://gist.github.com/rectangletangle/0a0d5a2e84dd3178d348
```
dweindl, line 3356: Do github gists come with a default license? :thinking:

python/tests/test_misc.py
```diff
@@ -85,3 +86,21 @@ def test_cmake_compilation(sbml_example_presimulation_module):
...
+
+
+def test_monkeypatch():
+    t = sp.Symbol('t')
+    n = sp.Symbol('n')
+    vals = [(t, 0),
+            (n, 1)]
+
+    # check that the removable singularity still exists
+    assert (t**n).diff(t).subs(vals) is sp.nan
```
dweindl, line 98: :+1:

python/tests/test_misc.py
```diff
@@ -85,3 +86,21 @@ def test_cmake_compilation(sbml_example_presimulation_module):
...
+
+    # check that we can monkeypatch it out
+    with _monkeypatched(sp.Pow, '_eval_derivative',
+                        _custom_pow_eval_derivative):
+        assert (t ** n).diff(t).subs(vals) is not sp.nan
+
+    # check that the monkeypatch is transiet
```
dweindl, line 105:
```suggestion
    # check that the monkeypatch is transient
```

python/tests/test_sbml_import.py
```diff
@@ -310,6 +312,41 @@ def test_units(model_units_module):
     assert rdata['status'] == amici.AMICI_SUCCESS
 
 
+def test_sympy_exp_monkeypatch():
+    url = 'https://www.ebi.ac.uk/biomodels/model/download/BIOMD0000000529.2?' \
```
dweindl, line 316: Add one-liner why this model is chosen?

src/model.cpp
```diff
@@ -1934,6 +1934,10 @@ void Model::fdwdw(const realtype t, const realtype *x) {
     fdwdw(dwdw_.data(), t, x, state_.unscaledParameters.data(),
           state_.fixedParameters.data(), state_.h.data(), w_.data(),
           state_.total_cl.data());
+    
+    if (always_check_finite_) {
+        app->checkFinite(gsl::make_span(dwdw_.get()), "dwdx");
```
dweindl, line 1939:
```suggestion
        app->checkFinite(gsl::make_span(dwdw_.get()), "dwdw");
```
?

dweindl (approved):

python/amici/ode_export.py
FFroehlich: Yes, and its no reuse 🤦 

https://stackoverflow.com/questions/61026720/the-default-license-of-a-gist

Unlikely to hold up though due to triviality as well as prior art.

python/amici/ode_export.py
FFroehlich: indeed!

src/model.cpp
FFroehlich: 👍

python/amici/ode_export.py
```diff
@@ -3344,3 +3350,54 @@ def smart_subs_dict(sym: sp.Expr,
...
+        name of the attribute to be patched
+
+    :param patch:
+        patched value
+
+    """
+    pre_patched_value = getattr(obj, name)
+    setattr(obj, name, patch)
+    yield object
+    setattr(obj, name, pre_patched_value)
```
lcontento, line 3374:
```suggestion
    try:
        yield object
    finally:
        setattr(obj, name, pre_patched_value)
```

python/amici/ode_export.py
```diff
@@ -3344,3 +3350,54 @@ def smart_subs_dict(sym: sp.Expr,
...
+    part1 = sp.Pow(self.base, self.exp - 1) * self.exp * dbase
+    part2 = self * dexp * sp.log(self.base)
+    if dbase.is_nonzero or self.base.is_nonzero or dbase.is_nonzero or \
+            part2.is_zero:
+        # first piece never applies or is zero anyways
+        return part1 + part2
+
+    return part1 + sp.Piecewise(
+        (
+            sp.sympify(0.0), sp.And(sp.Eq(self.base, 0), sp.Eq(dbase, 0))),
```
lcontento, line 3401:
```suggestion
            sp.sympify(0), sp.And(sp.Eq(self.base, 0), sp.Eq(dbase, 0))),
```

lcontento (approved):

python/amici/ode_export.py
```diff
@@ -3398,6 +3398,6 @@ def _custom_pow_eval_derivative(self, s):
 
     return part1 + sp.Piecewise(
         (
-            sp.sympify(0.0), sp.And(sp.Eq(self.base, 0), sp.Eq(dbase, 0))),
+            sp.Float(0.0), sp.And(sp.Eq(self.base, 0), sp.Eq(dbase, 0))),
```
lcontento, line 3401: Just nitpicking :wink: but since we are still at the symbolic stage why do you think float is better here?

python/amici/ode_export.py
dweindl: > triviality

True... but maybe then directly leave the reference out

python/amici/ode_export.py
FFroehlich: I generally dislike integers, bad habit from dividing them too often in C.

python/amici/ode_export.py
```diff
@@ -3344,3 +3350,54 @@ def smart_subs_dict(sym: sp.Expr,
...
+        name of the attribute to be patched
+
+    :param patch:
+        patched value
+
+    """
+    pre_patched_value = getattr(obj, name)
+    setattr(obj, name, patch)
+    try:
+        yield object
```
dilpath, line 3373:
```suggestion
        yield obj
```

python/amici/ode_export.py
```diff
@@ -3344,3 +3350,54 @@ def smart_subs_dict(sym: sp.Expr,
...
+
+    :param s:
+        variable with respect to which the derivative will be computed
+
+    """
+    dbase = self.base.diff(s)
+    dexp = self.exp.diff(s)
+    part1 = sp.Pow(self.base, self.exp - 1) * self.exp * dbase
+    part2 = self * dexp * sp.log(self.base)
+    if dbase.is_nonzero or self.base.is_nonzero or dbase.is_nonzero or \
```
dilpath, line 3395: `dbase.is_nonzero` is here twice?

python/amici/ode_export.py
```diff
@@ -3344,3 +3350,54 @@ def smart_subs_dict(sym: sp.Expr,
...
+            part2.is_zero:
+        # first piece never applies or is zero anyways
+        return part1 + part2
+
+    return part1 + sp.Piecewise(
+        (self.base, sp.And(sp.Eq(self.base, 0), sp.Eq(dbase, 0))),
```
dilpath, line 3401:
```suggestion
        (sp.Float(0), sp.And(sp.Eq(self.base, 0), sp.Eq(dbase, 0))),
```
Perhaps more readable?

dilpath (approved):

python/amici/ode_export.py
FFroehlich: True, but we do not know whether `self.base` is `sp.Integer(0)` or `sp.Float(0)` or some other variable that is equal zero and for which dbase also happens to be zero, so I think returning `self.base` is the implementation that makes the least possible assumptions.

python/amici/ode_export.py
FFroehlich: 👍

python/amici/ode_export.py
lcontento: The only downside of using `base` I see is a bit more code bloat (since it may be a *long* expression), but performance-wise the compiler should optimize it away.

PR #1327, opened 2020-10-28
Fix: Invalidate ReturnData::sllh on backwards integration failure

### LINKED ISSUE #1325: ReturnData::sllh not set to NaN upon backwards integration failures
Upon backwards integration failures, `ReturnData::sllh` should be set to NaN instead of being left in an undefined state. This seems to be broken.

Invalidating sllh seems to rely on `BackwardProblem*` being `nullptr` in `ReturnData::processSimulationObjects`. Not sure in which situations that would be the case, but not when running into `IntegrationFailureB`.

dweindl:
Closes #1325

### DIFF
--- a/src/amici.cpp
+++ b/src/amici.cpp
@@ -193,7 +193,7 @@ AmiciApplication::runAmiciSimulation(Solver& solver,
 
     rdata->processSimulationObjects(
         preeq.get(), fwd.get(),
-        bwd_success?bwd.get():nullptr,
+        bwd_success ? bwd.get() : nullptr,
         posteq.get(), model, solver, edata);
     return rdata;
 }
--- a/src/amici.cpp
+++ b/src/amici.cpp
@@ -139,7 +139,6 @@ AmiciApplication::runAmiciSimulation(Solver& solver,
 
 
         if (edata && solver.computingASA()) {
-            bwd_success = false;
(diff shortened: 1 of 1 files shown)


src/amici.cpp
```diff
@@ -187,8 +191,10 @@ AmiciApplication::runAmiciSimulation(Solver& solver,
                  ex.getBacktrace());
     }
 
-    rdata->processSimulationObjects(preeq.get(), fwd.get(), bwd.get(),
-                                    posteq.get(), model, solver, edata);
+    rdata->processSimulationObjects(
+        preeq.get(), fwd.get(),
+        bwd_success?bwd.get():nullptr,
```
FFroehlich, line 196:
```suggestion
        bwd_success ? bwd.get() : nullptr,
```

src/amici.cpp
```diff
@@ -155,6 +158,7 @@ AmiciApplication::runAmiciSimulation(Solver& solver,
                 preeq->workSteadyStateBackwardProblem(&solver, &model,
                                                       bwd.get());
             }
+            bwd_success = true;
```
FFroehlich, line 161: that's `posteq` and not `bwd` though

FFroehlich (approved):

src/amici.cpp
dweindl: Changed.

PR #1329, opened 2020-10-30
Fix infinite loop & identifier substitution

FFroehlich:


python/amici/ode_export.py
```diff
@@ -1749,6 +1749,13 @@ def _derivative(self, eq: str, var: str, name: str = None) -> None:
                 while nonzeros.max():
                     nonzeros = nonzeros.dot(nonzeros)
                     self._w_recursion_depth += 1
+                    if self._w_recursion_depth > len(sym_eq):
+                        raise RuntimeError(
+                            'dwdw is not nilpotent. Something, somewhere went '
+                            'terribly wrong. Please file a bug report at '
+                            'https://github.com/AMICI-dev/AMICI/issues and '
+                            'attach this model.'
+                        )
```
paulstapor, line 1758: Looks good. Like the error message. ;)

paulstapor (approved):
Can't judge to 100%, as I'm not deeply in the code here, but it seems to make sense as far as I can tell. Thanks!

python/amici/sbml_import.py
```diff
@@ -1202,21 +1227,22 @@ def _replace_in_all_expressions(self,
             for species in self.symbols[SymbolId.SPECIES].values():
                 species['init'] = smart_subs(species['init'],
                                              old, self._make_initial(new))
-                if 'dt' in species:
-                    species['dt'] = smart_subs(species['dt'], old, new)
+                for field in ['dt', 'compartment']:
```
dilpath, line 1230: I guess some errors in the test suite are due to this, as the compartment symbol is substituted with some value or expression. If this is kept, then maybe the [line in the test script that determines a species' compartment](https://github.com/AMICI-dev/AMICI/blob/fix_infinite_loop/tests/testSBMLSuite.py#L183) can be replaced with [getCompartment](https://github.com/AMICI-dev/AMICI/blob/fix_infinite_loop/tests/testSBMLSuite.py#L189).

dilpath (approved):
Looks good :+1: Sorry for the delay.

python/amici/sbml_import.py
FFroehlich: indeed, that code could and should easily be simplified anyways.

PR #1334, opened 2020-11-14
Update OpenBLAS from v0.3.10 to v0.3.12 for Windows

thomassligon:
Update OpenBLAS from v0.3.10 to v0.3.12, among other things because it supports building for double-precision only (v0.3.10 worked, but caused errors when trying to link utest).


scripts/installOpenBLAS.ps1
```diff
@@ -2,16 +2,16 @@ Write-Host 'script installOpenBLAS.ps1 started'
...
 $webclient.DownloadFile($uri,"$output")
-Expand-Archive -Path 'C:\BLAS\v0.3.10.zip' -DestinationPath 'C:\BLAS\OpenBLAS-v0.3.10' -Force # expand zip file
+Expand-Archive -Path 'C:\BLAS\v0.3.12.zip' -DestinationPath 'C:\BLAS\OpenBLAS-v0.3.12' -Force # expand zip file
 C:\Users\travis\build\AMICI\scripts\compileBLAS.cmd
 New-Item -Path 'C:\BLAS\lib' -ItemType Directory -Force # create directory
-Copy-Item "C:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10\lib\Release\openblas.lib" -Destination "C:\BLAS\lib" -Recurse
+Copy-Item "C:\BLAS\OpenBLAS-v0.3.12\OpenBLAS-0.3.12\lib\Release\openblas.lib" -Destination "C:\BLAS\lib" -Recurse
 New-Item -Path 'C:\BLAS\bin' -ItemType Directory -Force # create directory
-Copy-Item "C:\BLAS\OpenBLAS-v0.3.10\OpenBLAS-0.3.10\lib\openblas.dll" -Destination "C:\BLAS\bin" -Recurse
+Copy-Item "C:\BLAS\OpenBLAS-v0.3.12\OpenBLAS-0.3.12\lib\openblas.dll" -Destination "C:\BLAS\bin" -Recurse
```
dweindl, line 14: Might be good to introduce an `$openblas_version` variable in all the scripts. Will simplify future changes.

thomassligon:
This line in travis
`  - export BLAS_LIBS="/LIBPATH:C:\\BLAS\\lib openblas.lib" BLAS_CFLAGS="-IC:\\BLAS\\OpenBLAS-v$openBLAS_version\\OpenBLAS-$openBLAS_version"`
produced this in the compilation of cblas.cpp
`-IC:BLASOpenBLAS-v0.3.12OpenBLAS-0.3.12`
I'll try using forward slashes.

dweindl:
>  I'll try using forward slashes.

Alternatively another pair of backslashes or changing double quotes to single quotes

thomassligon:
Now the environment variable with the forward slashes look OK, but I don't think that the openBLAS installation script worked. What does
`Switched to branch 'develop'`
mean?
```
```$ powershell -File 'C:\Users\travis\build\AMICI\scripts\installOpenBLAS.ps1'
script installOpenBLAS.ps1 started
    Directory: C:\
Mode                LastWriteTime         Length Name                                                                  
----                -------------         ------ ----                                                                  
d-----       11/19/2020   9:46 AM                BLAS                                                                  
From https://github.com/AMICI-dev/AMICI
 * [new ref]           refs/pull/1334/head -> develop
Switched to branch 'develop'
script installOpenBLAS.ps1 completed

**Comment by thomassligon** [comment]
The double quotes are necessary so that the variable is expanded.
The double backslashes are a pain. The first version of MS-DOS introduced the backslash in paths, even though there was a lot of talk about upgrading MS-DOS to Xenix (which never happened). Ever since then it is a pain. But, as far as I can see, Windows can deal with forward slashes in path names, so that might always work for us, but you never know if some application somewhere still has problems. In summary, using backslashes because the command is intended for Windows causes problems with strings in BASH, python, and and and, so forward slashes is probably better.

**Comment by thomassligon** [comment]
OK, the problem is to be found in the raw log, since STDOUT and STDERR are not correlated in travis logs.

**Comment by thomassligon** [comment]
I don't understand the most recent failure. The script
`installOpenBLAS.ps1`
contains a copy statement
`Copy-Item "C:\BLAS\OpenBLAS-v$version\OpenBLAS-$version\lib\Release\openblas.lib" -Destination "C:\BLAS\lib" -Recurse`
and the result should show up in the ls/dir statement
`Get-ChildItem -Path "C:\BLAS" -Include "openblas.lib" -Recurse # check for file`
but it doesn't. That leads to the error in linking AMICI because openblas.lib couldn't be found.

**Comment by thomassligon** [comment]
It looks to me like this task is finished.

**Review by FFroehlich — APPROVED** [review]
Great, looks good. Thank You!

File: INSTALL.md [hunk 2]
```diff
@@ -221,7 +221,8 @@ The first script needs to be called in Powershell, and it needs to call `compile
 
     C:\Users\travis\build\AMICI\scripts\compileBLAS.cmd
```
dweindl, line 222: @thomassligon : Can you please update the documentation there, taking into account that that script now requires the OpenBLAS version (without "v") as argument. Maybe mentioning that 0.3.12 is known to work.

thomassligon:
The last changes, a document update and removal of the "v" from some names, is now working.

dweindl (approved):

FFroehlich:
Not sure why, but it looks like this broke again  :( https://travis-ci.com/github/AMICI-dev/AMICI/builds/204845595

thomassligon:
It looks like we have a mixed version, unless I am looking at the wrong branch. The develop branch seems to be using 0.3.12 in both scripts and the environment variable in travis.yml, and the master branch (relevant for this message?) seems to be using the 0.3.10 in the 2 scripts, but the environment variable created in travis.yml micht be using 0.3.12 because the link job contains 
`-IC:/BLAS/OpenBLAS-0.3.12/OpenBLAS-0.3.12 `

dweindl:
> Not sure why, but it looks like this broke again :( https://travis-ci.com/github/AMICI-dev/AMICI/builds/204845595

As it passes for the branch run and only fails for the PR run, I'd say this is fine.

PR #1338, opened 2020-11-20
New fim

### LINKED ISSUE #1335: Double check FIM approximation with parameter dependent sigma
Numerical optimization experiments suggest that we may not be doing the right there.

paulstapor:
If I didn't mess up signs, this might fix  #1335.

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -979,22 +979,23 @@ class SbmlImporter:
             for obs_id, obs in self.symbols[SymbolId.OBSERVABLE].items()
         }
 
-        self.symbols[SymbolId.LLHY] = {
-            symbol_with_assumptions(f'J{obs_id}'): {
-                'name': f'J{obs["name"]}',
-                'value': sp.sympify(noise_distribution_to_cost_function(
-                    noise_distributions.get(str(obs_id), 'normal')
-                )(obs_id), locals=dict(zip(
-                    _get_str_symbol_identifiers(obs_id),
-                    (obs_id, obs['measurement_symbol'], sigma_id)
-                ))),
-                'dist': noise_distributions.get(str(obs_id), 'normal'),
-            }
-            for (obs_id, obs), (sigma_id, sigma) in zip(
+        self.symbols[SymbolId.LLHY] = {}
+        for (obs_id, obs), (sigma_id, sigma) in zip(
                 self.symbols[SymbolId.OBSERVABLE].items(),
                 self.symbols[SymbolId.SIGMAY].items()
-            )
-        }
+        ):
+            symbol = symbol_with_assumptions(f'J{obs_id}')
+            dist = noise_distributions.get(str(obs_id), 'normal')
+            cost_fun = noise_distribution_to_cost_function(dist)(obs_id)
+            value = sp.sympify(cost_fun, locals=dict(zip(
+                _get_str_symbol_identifiers(obs_id),
+                (obs_id, obs['measurement_symbol'], sigma_id)
+            )))
+            self.symbols[SymbolId.LLHY][symbol] = {
+                    'name': f'J{obs["name"]}',
+                    'value': value,
+                    'dist': dist,
+                }
 
     @log_execution_time('processing SBML initial assignments', logger)
     def _process_initial_assignments(self):
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -987,7 +987,8 @@ class SbmlImporter:
                 )(obs_id), locals=dict(zip(
                     _get_str_symbol_identifiers(obs_id),
                     (obs_id, obs['measurement_symbol'], sigma_id)
-                )))
+                ))),
+                'dist': noise_distributions.get(str(obs_id), 'normal'),
             }
             for (obs_id, obs), (sigma_id, sigma) in zip(
                 self.symbols[SymbolId.OBSERVABLE].items(),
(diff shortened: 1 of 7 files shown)


paulstapor:
not yet doing the right thing, I'd say...

src/rdata.cpp
```diff
@@ -809,6 +809,7 @@ void ReturnData::fFIM(int it, Model &model, const ExpData &edata) {
     for (int iy = 0; iy < nytrue; ++iy) {
         if (!edata.isSetObservedData(it, iy))
             continue;
+        int iyt = iy + it * ny;
```
FFroehlich, line 812: just want to note this is not used anywhere.

src/rdata.cpp
```diff
@@ -818,7 +819,15 @@ void ReturnData::fFIM(int it, Model &model, const ExpData &edata) {
                     *
                     amici::fsres(y_it.at(iy), sy_it.at(iy + ny * jp),
                                  observedData[iy], sigmay_it.at(iy),
-                                 ssigmay_it.at(iy + ny * jp));
+                                 ssigmay_it.at(iy + ny * jp))
+                    -
+                    ssigmay_it.at(iy + ny * ip) * ssigmay_it.at(iy + ny * jp)
```
FFroehlich, line 824: Note that this term is already accounted for in `https://github.com/AMICI-dev/AMICI/blob/a4324d02c9b1c47c770b48f8427360fe61d5a1ff/src/rdata.cpp#L729`, but the sign may be wrong?

src/rdata.cpp
paulstapor: yes, that's part of the code from my old implementation of this, which is a while ago...
will remove, thanks

src/rdata.cpp
paulstapor: Ah, okay, good to konw! Then the only term which we needed is the $\sigma^4$ thing.
And yes, the sign was wrong...
So, the FIM I think we want is basically:
(sy / sigma)^T * (sy / sigma) + (ssigma / sigma)^T * (ssigma / sigma) - ((data - y) *ssigmy / sigma^2)^T * ((data - y)  *ssigmy / sigma^2)^T

FFroehlich:
@paulstapor Okay reworked everything again and added some documentation. Implementation should now be equivalent what `lsqnonlin` is doing with augmented residual. Added some doc that explains whats done there. However, this formulation is missing some `1/s^2` and `1/s^4` terms that could still be added.

python/amici/sbml_import.py
```diff
@@ -987,7 +987,8 @@ def _process_log_likelihood(self,
                 )(obs_id), locals=dict(zip(
                     _get_str_symbol_identifiers(obs_id),
                     (obs_id, obs['measurement_symbol'], sigma_id)
-                )))
+                ))),
+                'dist': noise_distributions.get(str(obs_id), 'normal'),
```
dweindl, line 991: `noise_distributions.get(str(obs_id), 'normal')` occurs already a few lines above. reuse result?

src/rdata.cpp
```diff
@@ -41,51 +42,53 @@ ReturnData::ReturnData(std::vector<realtype> ts, int np, int nk, int nx,
...
     chi2 = getNaN();
     if (sensi >= SensitivityOrder::first) {
         sllh.resize(nplist, getNaN());
         if (sensi >= SensitivityOrder::second)
             s2llh.resize(nplist * (nJ - 1), getNaN());
 
-        if (sensi_meth == SensitivityMethod::forward ||
-            sensi >= SensitivityOrder::second)
+        if ((sensi_meth == SensitivityMethod::forward ||
+            sensi >= SensitivityOrder::second) & enable_fim)
```
dweindl, line 67:
```suggestion
            sensi >= SensitivityOrder::second) && enable_fim)
```

Should work either way, but let's stick to logical operators.

src/rdata.cpp
```diff
@@ -804,21 +804,56 @@ void ReturnData::fFIM(int it, Model &model, const ExpData &edata) {
...
+     * -3*(s_dv*y_du + s_du*y_dv)*r/s^3 is missing from 2222 and
+     * -2*s_du*s_dv*r^2/s^4 is missing from 3333
+     */
+    
     auto observedData = edata.getObservedDataPtr(it);
+
     for (int iy = 0; iy < nytrue; ++iy) {
+        auto y = y_it.at(iy);
+        auto m = observedData[iy];
+        auto s = sigmay.at(iy);
```
dweindl, line 844: Unused if no data set for `iy`. Move below the following `if` block?

dweindl (approved):
Can't judge the actual FIM part. Rest looks good.

src/rdata.cpp
FFroehlich: :+1

python/amici/sbml_import.py
FFroehlich: would require unpacking the dict comprehension.

dweindl: true. okay. not expecting `noise_distributions` to grow too large...

FFroehlich: Decided to unpacking that monster of a comprehension, probably more readable like that anyways.

paulstapor (commented):
Looks good, except the return true thing...

src/model.cpp
```diff
@@ -546,6 +546,10 @@ std::vector<std::string> Model::getObservableIds() const {
     return std::vector<std::string>();
 }
 
+bool Model::hasQuadraticLLH() const {
+    return true;
```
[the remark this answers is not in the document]
FFroehlich: It's a virtual method that can be overridden by subclasses in case they don't implement a gaussian llh. Setting this to `true` by default ensure backwards compatibility.

PR #1352, opened 2020-12-02
Piecewise to heaviside

paulstapor:
Piecewise functions were so far parsed into if-else-statements, which made the model files poorly readable. Much more important though, the problem was that there might have been a discontinuity in the right hand side, but the solver didn't notice this and ran over it. This could lead to poor accuracy of the solver.

However, back from Matlab times, AMICI had it's own interface for Heaviside functions. If a piecewise function gets translated into an AMICI-Heaviside function (using the existing C++ code), both problems are solved. Moreover, this implementation should be a first step into the direction of implementing evens on the Python side.

I know this is a comparably large pull-request, sorry. But there's no way to split this into smaller parts, except including substantially less functionality, which is however urgently needed...

This implementation tries to stay as close to the existing event implementation as possible. It is supposed to be working for forward sensitivities and Heaviside functions in full generality, i.e., also supporting parameter and state-dependent arguments of HEaviside functions.


python/amici/ode_export.py
```diff
@@ -1465,6 +1584,70 @@ def generate_basic_variables(self, *, from_sbml: bool = False) -> None:
...
+                tmp_new, roots = _make_unique(tmp_old, roots)
+                _, roots = _make_unique(sp.sympify(-1 * tmp_old), roots)
+                heavisides.append((sp.Heaviside(tmp_old), tmp_new))
+
+            return roots, heavisides
+
+        for state in self._states:
+            roots, heavisides = _extract_heavisides(state._dt, roots)
+            for heaviside_sympy, heaviside_amici in heavisides:
+                state._dt = state._dt.subs(heaviside_sympy, heaviside_amici)
```
FFroehlich, line 1645: Isn't it possible to already perform the substitution while performing `extract_heavisides`? Applying subs can be quite costly especially for larger models.

python/amici/ode_export.py
```diff
@@ -1653,12 +1837,41 @@ def _compute_equation(self, name: str) -> None:
             # force symbols
             self._derivative('xdot', 'p', name=name)
 
+        elif name == 'drootdt':
+            self._eqs[name] = self.eq('root').jacobian(time_symbol)
+
+        elif name == 'drootdt_total':
+            self._eqs[name] = self.eq('drootdt') * self._eqs['xdot'] + \
```
FFroehlich, line 1844: Shouldn't the first `drootdt` here be `drootdx`?

Note that using `self._eqs['xdot']` means `drootdt` may depend on `w`, which was not the case in the matlab interface where `model.fun.xdot.sym_noopt` was used. To avoid further breaking with matlab generated code here, I would simply substitute in all expressions here, which is not the most efficient, but should hopefully be fine for most practical purposes.

python/amici/ode_export.py
```diff
@@ -1667,7 +1880,11 @@ def _compute_equation(self, name: str) -> None:
 
         if self._simplify:
             dec = log_execution_time(f'simplifying {name}', logger)
-            self._eqs[name] = dec(self._eqs[name].applyfunc)(self._simplify)
+            if name in ('stau', 'deltasx'):
```
FFroehlich, line 1883:
```suggestion
            if isinstance(self._eqs[name], list):
```

FFroehlich (changes requested):

python/amici/sbml_import.py
```diff
@@ -1473,6 +1468,40 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+                trigger = root.args[0] - root.args[1]
+                if type(root) in (sp.core.relational.StrictLessThan,
+                                  sp.core.relational.LessThan):
+                    trigger = -trigger
+            else:
+                raise SBMLException('AMICI can not parse peicewise functions '
+                                    'with non-relational arguments.')
+            tmp = sp.Heaviside(trigger)
+            formula += coeff * tmp
+            lastroot -= tmp
```
FFroehlich, line 1499: Not sure whether this formula is actually correct. https://docs.sympy.org/latest/modules/functions/elementary.html#piecewise states that **only the first** expression that is true will be returned. I think applying `sp.piecewise_fold` and then translating the generated `ITE` expression may have a better mileage.

don't like the naming `lastroot` here, can we have something more descriptive?

python/amici/ode_export.py
```diff
@@ -626,6 +676,49 @@ def __init__(self,
...
+
+        :param value:
+            formula for the root / trigger function
+
+        :param rhs_update:
+            formula for the update of the rigth hand side (for Heaviside
+            functions this is just the coefficient)
+
+        :param state_update:
+            formula for the bolus function (0 for Heaviside functions)
```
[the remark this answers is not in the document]
paulstapor: Depends... for an actual event, yes. But putting None in here gives us a nice possibility to discern between events and heavisides... although one could simply add a flag for this case...

paulstapor:
Thanks for the code review! I'll implement the suggestions as soon as the test cases are there!

python/amici/ode_export.py
```diff
@@ -626,6 +676,48 @@ def __init__(self,
         """
         super(LogLikelihood, self).__init__(identifier, name, value)
 
+class Event(ModelQuantity):
+    """
+    A Event defines the either a trigger of an event or a root of the argument
```
dweindl, line 681:
```suggestion
    An Event defines either a trigger of an event or a root of the argument
```

python/amici/ode_export.py
```diff
@@ -1223,6 +1324,15 @@ def num_expr(self) -> int:
         """
         return len(self.sym('w'))
 
+    def num_events(self) -> int:
+        """
+        Number of Expressions.
```
dweindl, line 1329:
```suggestion
        Number of Events.
```

python/amici/ode_export.py
```diff
@@ -1223,6 +1324,15 @@ def num_expr(self) -> int:
...
+
+        :return:
+            number of expression symbols
```
dweindl, line 1332:
```suggestion
            number of event symbols
```

python/amici/ode_export.py
```diff
@@ -1465,6 +1583,70 @@ def generate_basic_variables(self, *, from_sbml: bool = False) -> None:
...
+                roots.append(Event(
+                    identifier=sp.Symbol(root_symstr),
+                    name=root_symstr,
+                    value=root_found,
+                    state_update=None,
+                    event_observable=None
+                ))
+                return roots[-1]._identifier, roots
+
+        def _extract_heavisides(dxdt, roots):
```
dweindl, line 1624: Please add docstring

python/amici/ode_export.py
```diff
@@ -1653,12 +1836,41 @@ def _compute_equation(self, name: str) -> None:
             # force symbols
             self._derivative('xdot', 'p', name=name)
 
+        elif name == 'drootdt':
+            self._eqs[name] = self.eq('root').jacobian(time_symbol)
```
dweindl, line 1840:
```suggestion
            self._eqs[name] = self.eq('root').smart_jacobian(time_symbol)
```
?

python/amici/ode_export.py
```diff
@@ -1653,12 +1836,41 @@ def _compute_equation(self, name: str) -> None:
...
+
+        elif name == 'drootdt_total':
+            self._eqs[name] = self.eq('drootdt') * self._eqs['xdot'] + \
```
dweindl, line 1843: `smart_multiply` ?

python/amici/ode_export.py
```diff
@@ -1653,12 +1836,41 @@ def _compute_equation(self, name: str) -> None:
...
+
+        elif name == 'stau':
+            self._eqs[name] = [
+                -self.eq('sroot')[ie, :] / self.eq('drootdt_total')[ie]
+                for ie in range(self.num_events())
+            ]
+
+        elif name == 'deltasx':
+            self._eqs[name] = [
+                (self.eq('xdot_old') - self.eq('xdot')) * self.eq('stau')[ie]
```
dweindl, line 1854: `smart_multiply` ?

python/amici/ode_export.py
```diff
@@ -1653,12 +1836,41 @@ def _compute_equation(self, name: str) -> None:
...
+
+
         elif match_deriv:
             self._derivative(match_deriv.group(1), match_deriv.group(2))
 
         else:
             raise ValueError(f'Unknown equation {name}')
 
+        if name == 'root':
+            # Events et processed after the ODE model has been set up.
```
dweindl, line 1870: "et"? "are"? "need to be"?

python/amici/ode_export.py
```diff
@@ -2209,9 +2425,12 @@ def __init__(
         self.model_path: str = os.path.abspath(output_dir)
         self.model_swig_path: str = os.path.join(self.model_path, 'swig')
 
+        # We need to process events and Heaviside functions in the ODE Model,
+        # before adding it to ODEExporter
+        ode_model.get_events()
```
dweindl, line 2430: calling `get_events` without using the return value feels wrong. rename?

python/amici/ode_export.py
```diff
@@ -2643,6 +2862,22 @@ def _get_function_body(self,
                 lines.append(f'{function}[{index}] = '
                              f'{_print_with_exception(formula)};')
 
+        elif function in event_functions:
+            outer_cases = {}
+            for ie in range(self.model.num_events()):
+                inner_equations = equations[ie]
```
dweindl, line 2868:
```suggestion
            for ie, inner_equations in enumerate(equations):
```
?

python/amici/ode_export.py
```diff
@@ -2643,6 +2862,22 @@ def _get_function_body(self,
...
+                inner_lines = []
+                if not smart_is_zero_matrix(inner_equations):
+                    inner_cases = {
+                        ipar: _get_sym_lines_array(inner_equations[:, ipar],
+                                                   function, 0)
+                        for ipar in range(self.model.num_par())
+                        if not smart_is_zero_matrix(inner_equations[:, ipar])}
```
dweindl, line 2875: `inner_equations` is checked for zero twice, first the complete one, then element-wise. not sure how costly. could do element-wise at first occurrence and save result.

python/amici/sbml_import.py
```diff
@@ -1473,6 +1468,40 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+        symbolic expressions for arguments of the piecewise function
+    """
+    # how many condition-expression pairs will we have?
+    formula = sp.sympify('0')
+    lastroot = sp.Float(1.0)
+    conditions = list(grouper(args, 2, False))
+    for coeff, root in conditions:
+        if root != sp.false:
+            # Sympy doesn't assume by default that the expression is real...
+            root._assumptions['extended_real'] = True
```
dweindl, line 1486:
```suggestion
            root._assumptions['extended_real'] = True
```

Is there a cleaner way for modifying assumptions. Checked sometime ago, but didn't find any. Maybe somebody else knows.

python/amici/sbml_import.py
```diff
@@ -1473,6 +1468,40 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+            # we now need to convert the relational >, >=, ... expression into
+            # a trigger function which will be a Heaviside argument
+            if root.is_Relational:
+                trigger = root.args[0] - root.args[1]
+                if type(root) in (sp.core.relational.StrictLessThan,
+                                  sp.core.relational.LessThan):
+                    trigger = -trigger
+            else:
+                raise SBMLException('AMICI can not parse peicewise functions '
```
dweindl, line 1495:
```suggestion
                raise SBMLException('AMICI can not parse piecewise functions '
```

python/amici/sbml_import.py
```diff
@@ -1473,6 +1468,40 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+                                    'with non-relational arguments.')
```
dweindl, line 1496: Show ID / formula of the problematic function

python/amici/ode_export.py
paulstapor: It's already done state by state right after parsing, so the whole right hand side is not checked more often than necessary, I hope. Replacing right at checking for occurrence will not work well, as this will/must lead to redundantly saved roots. One could go for parsing string instead of symbols, but I'm somewhat sure that sympy does exactly this...
I tried to optimize it a little further by moving subs into the _extract_heavisides function, hence only those roots, which were found in a certain line of the rhs actually get replaced, and no others will be attempted (which will save time).
But beyond this, I think thereÄs no further optimization which will not break some functionality...

python/amici/ode_export.py
paulstapor: totally right for the drootdx, changed it. Will now do the non-optimized rhs thing.

python/amici/sbml_import.py
paulstapor: I know what you mean. I looked into this a little, but didn't find a good API doc for this... If there is one, it's not well findable. As the current implementation works and is also clear about it's purpose (although not super nice, as opeating on protected variables), I wouldn't bother with searching longer than a few minutes for an API...

python/amici/ode_export.py
paulstapor: I was hoping that checking the whole thing is somewhat faster... But I removed the outer checking now.

python/amici/ode_export.py
paulstapor: Feels like you're right. Renamed to parse_events()

python/amici/sbml_import.py
paulstapor: You're right, but I will refactor this anyway now...

python/amici/sbml_import.py
```diff
@@ -1473,6 +1468,40 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+            tmp = sp.Heaviside(trigger)
```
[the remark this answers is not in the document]
paulstapor: I tried it. The answer is no, we don't want that.... :D
It does not seem to always do the best job, as it sometimes inflates the expression quite a bit, making it much harder to read.

python/amici/sbml_import.py
```diff
@@ -1473,6 +1468,40 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+    conditions = list(grouper(args, 2, False))
+    for coeff, root in conditions:
+        if root != sp.false:
+            # Sympy doesn't assume by default that the expression is real...
+            root._assumptions['extended_real'] = True
+            # we now need to convert the relational >, >=, ... expression into
+            # a trigger function which will be a Heaviside argument
+            if root.is_Relational:
+                trigger = root.args[0] - root.args[1]
+                if type(root) in (sp.core.relational.StrictLessThan,
```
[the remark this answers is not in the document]
paulstapor: Will have to rewrite that anyway, in order to get something which also survives boolean operations...

python/amici/sbml_import.py
paulstapor: Eh... won't work. Piecewise_fold goes the other way round. It converts ITE into piecewise, not away from it...
I am writing a parser here... yay,,,

FFroehlich: Hmm, I thought piecewise_fold goes both ways:

https://docs.sympy.org/latest/modules/functions/elementary.html

python/amici/ode_export.py
```diff
@@ -626,6 +677,46 @@ def __init__(self,
         """
         super(LogLikelihood, self).__init__(identifier, name, value)
 
+class Event(ModelQuantity):
+    """
+    An Event defines either an SBML event or a root of the argument of a
```
FFroehlich, line 682:
```suggestion
    An Event defines either a SBML event or a root of the argument of a
```

FFroehlich (approved):
Good to go from my end

python/amici/ode_export.py
```diff
@@ -1459,12 +1568,34 @@ def generate_basic_variables(self, *, from_sbml: bool = False) -> None:
...
+        # before adding it to ODEExporter
+        self.parse_events()
+
         for var in self._variable_prototype:
             if var not in self._syms:
                 self._generate_symbol(var, from_sbml=from_sbml)
 
         self._generate_symbol('x', from_sbml=from_sbml)
 
+    def parse_events(self) -> None:
```
dweindl, line 1582:
```suggestion
    def _parse_events(self) -> None:
```
?

python/amici/ode_export.py
```diff
@@ -3430,3 +3618,117 @@ def _custom_pow_eval_derivative(self, s):
...
+
+def _process_heavisides(dxdt: sp.Expr, roots: List[Event]) -> sp.Expr:
+    """
+    Parses the RHS of a state variable, checks for Heaviside functions,
+    collects unique roots functions that can be tracked by SUNDIALS and
+    replaces Heaviside Functions by amici helper variables that will be
+    updated based on SUNDIALS root tracking.
+
+    :param dxdt:
+        right hand sid eof state variable
```
dweindl, line 3706:
```suggestion
        right hand side of state variable
```

python/amici/ode_export.py
```diff
@@ -2051,7 +2221,12 @@ def _print_with_exception(math: sp.Expr) -> str:
     user_functions = {fun['sympy']: fun['c++'] for fun in CUSTOM_FUNCTIONS}
 
     try:
-        ret = cxxcode(math, standard='c++11', user_functions=user_functions)
+        with _monkeypatched(_CXXCodePrinterBase, '_print_Max',
```
dweindl, line 2224:
```suggestion
        # Required until https://github.com/sympy/sympy/pull/20558 is released
        with _monkeypatched(_CXXCodePrinterBase, '_print_Max',
```

python/amici/ode_export.py
```diff
@@ -3430,3 +3618,117 @@ def _custom_pow_eval_derivative(self, s):
...
+    Custom Min printing function, see https://github.com/sympy/sympy/pull/20558
+    """
+    from sympy import Min
+    if len(expr.args) == 1:
+        return self._print(expr.args[0])
+    return "%smin(%s, %s)" % (self._ns, self._print(expr.args[0]),
+                              self._print(Min(*expr.args[1:])))
+
+
+def _collect_heaviside_roots(args):
```
dweindl, line 3645: Add argument and return type hint

python/amici/ode_export.py
```diff
@@ -3430,3 +3618,117 @@ def _custom_pow_eval_derivative(self, s):
...
+    """
+    Recursively checks an expression for the occurrence of Heaviside
+    functions and return all roots found
```
dweindl, line 3648: document argument / return value

python/amici/sbml_import.py
```diff
@@ -1473,6 +1482,105 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+def _denest_piecewise(args):
+    """
+    Denest piecewise functions that contain piecewise as condition
+
+    :param args:
+        Arguments to the piecewise function
+
+    :return:
+        Arguments where conditions no longer contain piecewise functions and
+        the coditional dependency is flattend out
```
dweindl, line 1494:
```suggestion
        the conditional dependency is flattened out
```

python/amici/sbml_import.py
```diff
@@ -1473,6 +1482,105 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
     return sym
 
 
+def _denest_piecewise(args):
```
dweindl, line 1485: please add typehints (input + return)

python/amici/sbml_import.py
```diff
@@ -1473,6 +1482,105 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+
+        else:
+            args_out.extend([coeff, cond])
+    # cut off last condition as that's the default
+    return tuple(args_out[:-1])
+
+def _parse_piecewise_to_heaviside(args: Iterable[sp.Expr]) -> sp.Expr:
+    """
+   Piecewise functions cannot be transformed into C++ right away, but AMICI has
+   a special interface for Heaviside function, so we transform them
```
dweindl, line 1530: fix indendation

python/amici/sbml_import.py
```diff
@@ -1473,6 +1482,105 @@ def _parse_special_functions(sym: sp.Expr, toplevel: bool = True) -> sp.Expr:
...
+        elif isinstance(trigger, sp.Or):
+            return sp.Max(*[_parse_trigger(arg)
+                            for arg in trigger.args])
+        elif isinstance(trigger, sp.And):
+            return sp.Min(*[_parse_trigger(arg)
+                            for arg in trigger.args])
+        else:
+            raise SBMLException('AMICI can not parse piecewise functions '
+                                f'with argument {trigger}.')
+        return root
```
dweindl, line 1561: Seems overly complex. Simplify?

```suggestion
            return root
        
        if isinstance(trigger, sp.Or):
            return sp.Max(*[_parse_trigger(arg)
                            for arg in trigger.args])
        
        if isinstance(trigger, sp.And):
            return sp.Min(*[_parse_trigger(arg)
                            for arg in trigger.args])
        
        raise SBMLException('AMICI can not parse piecewise functions '
                                f'with argument {trigger}.')
        
```

dweindl (approved):

documentation/python_interface.rst
```diff
@@ -25,7 +25,7 @@ AMICI can import :term:`SBML` models via the
 Status of SBML support in Python-AMICI
 ++++++++++++++++++++++++++++++++++++++
 
-Python-AMICI currently **passes 850 out of the 1780 (~48%) test cases** from
+Python-AMICI currently **passes 805 out of the 1780 (~45%) test cases** from
```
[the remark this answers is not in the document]
FFroehlich: Not sure what happened, looking at the tests that are skipped I couldn't really identify those 45 that are not longer passing.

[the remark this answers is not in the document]
dweindl: Ah, this is unrelated to this PR. (Here it increases from 804 to 805). This is where it happened: https://github.com/AMICI-dev/AMICI/actions/runs/331448378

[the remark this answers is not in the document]
dweindl: Used to work but skipped now:
```
969
970
971
974
1104
1105
1107
1109
1121
1434
1436
1438
1440
1442
1465
1498
1552
1554
1574
1638
1639
1651
1652
1653
1654
1655
1656
1657
1722
1724
1725
1726
1733
1734
1735
1744
1746
1748
1750
1753
1764
1765
1766
1767
1768
1773
1774
```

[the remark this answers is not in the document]
FFroehlich: I suspect thats due to the skippings with msgs like
```
SKIPPED [2] tests/testSBMLSuite.py:87: AMICI does not support SBML models containing variables with SId k0.
3677
SKIPPED [3] tests/testSBMLSuite.py:87: AMICI does not support SBML models containing variables with SId A1_sr.
3678
SKIPPED [2] tests/testSBMLSuite.py:87: AMICI does not support SBML models containing variables with SId A_sr1.
```
Those are a little over 50 cases, so lets have a more detailed look in a different PR. Also looks like there are still a handful of low hanging fruits like ceil/min/max applied to numerical values, which we should be able to pass.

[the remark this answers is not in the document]
dweindl: Yes, separate PR.

paulstapor:
Just also had a look at the refactored test. Looks fine for me as well!

PR #1360, opened 2020-12-10
Add support for different noise models for PySB import

### LINKED ISSUE #1176: Support of different noise models for PySB import
PySB import currently assumes Gaussian noise for all observables and there is no convenient way to change that. Would be great if one could also specify observable-specific noise distributions as already implemented for SBML import.

dweindl:
.. as already the case for SBML import.

Closes #1176

### DIFF
--- a/python/amici/pysb_import.py
+++ b/python/amici/pysb_import.py
@@ -201,7 +201,7 @@ def ode_model_from_pysb_importer(
                               noise_distributions)
     _process_pysb_expressions(model, ode, observables, sigmas,
                               noise_distributions)
-    ode._has_quadratic_nllh = all(
+    ode._has_quadratic_nllh = not noise_distributions or all(
         noise_distr in ['normal', 'lin-normal']
         for noise_distr in noise_distributions.values()
     )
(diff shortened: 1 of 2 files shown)


python/amici/pysb_import.py
```diff
@@ -9,7 +9,8 @@
     ODEExporter, ODEModel, State, Constant, Parameter, Observable, SigmaY,
     Expression, LogLikelihood, generate_measurement_symbol
 )
-
+# TODO move away from sbml import
+from .sbml_import import noise_distribution_to_cost_function, _get_str_symbol_identifiers
```
dweindl, line 13: @FFroehlich : I'd like to move those functions out of `sbml_import`, as they are not sbml-specific. Would not add them to ode_export, as it's rather on the import side (and that file is already too huge). What about introducing something like `import_utils` for general import functions?

FFroehlich: sounds good!

python/amici/pysb_import.py
```diff
@@ -102,6 +116,7 @@ def pysb2amici(model: pysb.Model,
     ode_model = ode_model_from_pysb_importer(
         model, constant_parameters=constant_parameters,
         observables=observables, sigmas=sigmas,
+        noise_distributions=noise_distributions,
```
dweindl, line 119: At some point we might want to keep those together in a single dict...

FFroehlich (approved):
pysb import also needs to set `ODEModel._has_quadratic_nllh`

PR #1361, opened 2020-12-11
Refactor: introduce import_utils.py

dweindl:
To collect format-independent functions for model import.

Just moving code. No changes in implementation.


FFroehlich (approved):

FFroehlich:
If toposort symbols is also moved, this file should be added to the triggers for running SBML testsuite

dweindl:
> If toposort symbols is also moved, this file should be added to the triggers for running SBML testsuite

Thanks, good point.

PR #1362, opened 2020-12-11
CI: Show all compiler warnings and treat as error

### LINKED ISSUE #1167: Build AMICI with `-Werror` during CI runs
As recent changes caused a number of compiler warnings...

(Also for swigging)

dweindl:
Closes #1167

... and pass `NVector`s as pointers to const where possible (as opposed to unintended const pointers to non-const before).


FFroehlich (approved):
👍

PR #1372, opened 2021-01-05
Fix species assignment rules in reactions

PaulJonasJost:
Hey, i encountered an error with an assignment rule in my sbml model. Here is a fix, although credits belong to @dilpath

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -787,6 +787,8 @@ class SbmlImporter:
                     species_id = _get_identifier_symbol(
                         self.sbml.getSpecies(element.getSpecies())
                     )
+                    if self.is_assignment_rule_target(self.sbml.getSpecies(element.getSpecies())):
+                        continue
                     species = self.symbols[SymbolId.SPECIES][species_id]
 
                     if species['constant']:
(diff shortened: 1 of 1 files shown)


dilpath:
Somehow after requesting a review from Daniel, your username has disappeared from the list of possible reviewers, @FFroehlich :see_no_evil: F5 doesn't help either...

Would be good for you to check too!

python/amici/sbml_import.py
```diff
@@ -787,6 +787,8 @@ def _process_reactions(self):
                     species_id = _get_identifier_symbol(
                         self.sbml.getSpecies(element.getSpecies())
                     )
+                    if self.is_assignment_rule_target(self.sbml.getSpecies(element.getSpecies())):
+                        continue
```
dilpath, line 791:
```suggestion
                    sbml_species = self.sbml.getSpecies(element.getSpecies())
                    if self.is_assignment_rule_target(sbml_species):
                        continue
                    species_id = _get_identifier_symbol(sbml_species)
```

FFroehlich: 👍

FFroehlich (approved):

PR #1374, opened 2021-01-14
Add expression IDs and names to generated models

### LINKED ISSUE #1373: Include expression IDs in generated models
So far, observable/parameter/state IDs/names are available in AMICI models. Expressions IDs/names are not. Would be convenient to have them available too (e.g. for debugging, and they would be handy for more informative error messages). Does anything speak against adding those?

dweindl:
Closes #1373

### DIFF
--- a/include/amici/model.h
+++ b/include/amici/model.h
@@ -547,6 +547,19 @@ class Model : public AbstractModel {
     virtual std::vector<std::string> getObservableNames() const;
 
     /**
+     * @brief Report whether the model has expression names set.
+     * @return Boolean indicating whether expression names were set. Also
+     * returns `true` if the number of corresponding variables is just zero.
+     */
+    virtual bool hasExpressionNames() const;
+
+    /**
+     * @brief Get names of the expressions.
+     * @return Expression names
+     */
+    virtual std::vector<std::string> getExpressionNames() const;
+
+    /**
      * @brief Report whether the model has parameter IDs set.
      * @return Boolean indicating whether parameter IDs were set. Also returns
      * `true` if the number of corresponding variables is just zero.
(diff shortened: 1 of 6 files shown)


include/amici/model.h
```diff
@@ -599,7 +612,21 @@ class Model : public AbstractModel {
...
+     * @return Boolean indicating whether expression ids were set. Also returns
+     * `true` if the number of corresponding variables is just zero.
+     */
+    virtual bool hasExpressionIds() const;
+
+    /**
+     * @brief Get IDs of the expression.
+     * @return Expression IDs
+     */
+    virtual std::vector<std::string> getExpressionIds() const;
```
dweindl, line 625: We could consider changing that at some point to `getIdsFor({observables,states,expressions})` etc. That would save quite a bit of more or less replicated code...

FFroehlich (approved):

FFroehlich:
add tests?

dweindl:
> add tests?

Hm... Somewhat boring to unit test. It's exercised via `test_swig_interface.py` (which would not fail if the functions were removed). A test for correct IDs would have to be adapted if naming or handling of expressions changes, probably more annoying than helpful?

FFroehlich:
> > add tests?
> 
> Hm... Somewhat boring to unit test. It's exercised via `test_swig_interface.py` (which would not fail if the functions were removed). A test for correct IDs would have to be adapted if naming or handling of expressions changes, probably more annoying than helpful?

Hmm don't see any tests that check the functionality there. What about a test on a small pysb model that simply checks that name/ids are not `''`, i.e., actually set during model construction.

dweindl:
> Hmm don't see any tests that check the functionality there. 

Wouldn't call it tested, but used (`test_copy_constructors`).

> What about a test on a small pysb model that simply checks that name/ids are not `''`, i.e., actually set during model construction.

Added test for the pysb presimulation example to avoid importing an additional model. Checking full names/IDs now, hoping they won't change too often.

Issue #464, opened 2018-11-14
DAE integration fails


yannikschaelte:
I have a problem with a simple published model (5 states, 8 parameters): I would like to have it as a DAE, but AMICI fails to do the simulation then, giving a status -99 and just nan values. Using the matlab solver ODE15S works and gives the expected result. Also simulating the system as an ODE works. Computing with or without gradients changes nothing about this.

The output I get from the script contained in the enclosed zip is:

```
Solved with AMICI as ODE:

ans =

     0


ans =

    0.4405    4.6192    0.8595

Warning: AMICI ERROR: in module IDAS in function IDACalcIC : Newton/Linesearch algorithm failed to converge.  
> In simulate_model_kuntz_dae (line 232)
  In example_kuntz (line 58) 
Warning: AMICI simulation failed:
AMICI setup failed:
(IDA routine IDACalcIC failed with error code -4)
Error occured in:
stacktrace not available on windows platforms
 
> In simulate_model_kuntz_dae (line 232)
  In example_kuntz (line 58) 
Solved with AMICI as DAE:

ans =

   -99


ans =

   NaN   NaN   NaN

Solved with ODE15S as ODE:

ans =

    0.4405    4.6192    0.8595    0.8636    2.4634

Solved with ODE15S as DAE:

ans =

    0.4405    4.7000    0.8696    0.8696    2.4633
```

The test script: [example.zip](https://github.com/ICB-DCM/AMICI/files/2580126/example.zip)

Tested on Windows with latest amici and Matlab 2017b. Any hints on what is wrong here?

FFroehlich:
Sorry, I have very limited knowledge of DAEs, so I am probably not going to be of much help.

What I can tell you:

IDAs fails to compute consistent initial conditions (`IDACalcIC : Newton/Linesearch algorithm failed to converge.`). I remember that we had some trouble with this a while back and I would currently guess that this is more of an implementation problem on the AMICI side (as in either the options or input functions to IDACalcIC are incorrect).

FFroehlich:
The only DAE testcase we have is the robertson mode, so there is overall very limited testing of DAEs.

What I imagine could be wrong: 
- The functions to compute the mass matrix and/or the system jacobian.
- The options that are passed to `IDACalcIC`. If I remember correctly, the key parameter is whether the initial conditions should be computed based on the initial conditions for the state or the derivative of the state.
- something entirely different I have not thought of so far.

JanHasenauer:
[...]

PR #1376, opened 2021-01-15
fix id vector for DAEs

FFroehlich:
Following a post on the SUNDIALS mailing list, I realized that the ID vector is incorrectly implemented. The previous implementation computed the differential *equations* (which was also done incorrectly) and not the *components*.


matlab/@amimodel/parseModel.m
```diff
@@ -184,7 +184,7 @@ function parseModel(this)
 if(strcmp(this.wtype,'iw'))
     if(isfield(this.fun,'M'))
         this.getFun([],'M');
-        this.id = double(logical(sum(this.fun.M.sym,2)~=0));
+        this.id = double(logical(any(this.fun.M.sym)));
```
paulstapor, line 187: Shouldn't it be `any(this.fun.M.sym, 2)` ?

dweindl: That would yield the same result, not? (With `any` potentially being a bit faster - not that it really matters here)

dweindl (commented):
As this went unnoticed for quite a while, it might be good to have it tested somewhere. Although I'd rather not add much new matlab code.

matlab/@amimodel/parseModel.m
FFroehlich: From my understanding, checking for differential *components* can be done by checking whether any of the respective columns of the mass matrix are nonzero, to see whether the time derivative of the respective component appears *anywhere* in the rhs.

paulstapor: Ah, okay. I see my error. Indeed, your code is fine!

paulstapor (approved):

matlab/@amimodel/parseModel.m
```diff
@@ -184,7 +184,7 @@ function parseModel(this)
 if(strcmp(this.wtype,'iw'))
     if(isfield(this.fun,'M'))
         this.getFun([],'M');
```
dweindl, line 186:
```suggestion
if(strcmp(this.wtype, 'iw'))
    if(isfield(this.fun, 'M'))
        this.getFun([], 'M');
```

Just to trigger GHA...

FFroehlich:
existing examples work, #464 remains unfixed by this.

PR #1382, opened 2021-01-26
account for heaviside functions in expressions

FFroehlich:


paulstapor (approved):
should hopefully do the job...

dilpath (approved):
Should have a test at some point I guess

FFroehlich:
> Should have a test at some point I guess

will add Chen_MSB2009 which should now be working.

src/backwardproblem.cpp
```diff
@@ -174,9 +176,10 @@ void BackwardProblem::handleDataPointB(const int it) {
 }
 
 realtype BackwardProblem::getTnext(const int it) {
-    if (discs_.size() > 0 && discs_.back() > model_->getTimepoint(it)) {
+    assert (it >= 0 || discs_.size() > 0);
```
dweindl, line 179: What about keeping cheap assertions also active in release builds?

python/amici/sbml_import.py
```diff
@@ -1625,16 +1625,16 @@ def _parse_trigger(trigger: sp.Expr) -> sp.Expr:
         # translating into aheaviside function that is implemented as unit step
         if isinstance(trigger, sp.core.relational.StrictLessThan):
             # x < y => x - y < 0 => r < 0
-            return sp.Heaviside(-root)
+            return 1 - sp.Heaviside(root)
         if isinstance(trigger, sp.core.relational.LessThan):
             # x <= y => not(y < x) => not(y - x < 0) => not -r < 0
-            return 1 - sp.Heaviside(root)
+            return sp.Heaviside(-root)
```
dilpath: What's the difference between `1-H(x)` and `H(-x)`? Should the difference between `< ` and `<=` be the value that `H(x)` is defined to take at `x==0`? Similarly for `>` and `>=` below?

dilpath: Ah, nevermind https://github.com/AMICI-dev/AMICI/blob/a5c679b0cce90e192bacf9461d43825de8f675ce/src/symbolic_functions.cpp#L90-L95

dilpath: Hm, given this definition, `x < 0` would correspond to `H(-x)`?

paulstapor: Actually, `1-H(x)` and `H(-x)` should be equivalent... For the difference between `<` and `<=` I have to say: Hard to tell what will happen if x is a floating point number...

dilpath: > `1-H(x)` and `H(-x)` should be equivalent

Not equivalent for this implementation of the Heaviside function but agreed :+1: (equivalent everywhere except 0)

FFroehlich: This implementation addresses the situation when the simulation starts at a point where some of the arguments of heaviside functions evaluates to 0 (which can happen when translating SBML piecewise functions), which led to errors previously. It looks like the cpp implementation needs to be updated as its not consistent with the internal implementation. Will add a short blurb about that to the documentation.

FFroehlich:
> > Should have a test at some point I guess
> 
> will add Chen_MSB2009 which should now be working.

Actually just disabling the model agains since that model is a mess (as reported before). I'm confident that implementation is correct, but llh values don't match. Also functionality is tested in Weber_BMC2015.

PR #1383, opened 2021-01-28
Allow loading of existing models in import_petab_problem

lcontento:
`import_petab_problem` will fail if a model module has already been generated. This PR makes it possible.

Is there any reason why a pre-existent model should not be loaded when `force_compile` is false? If so then maybe we could add a keyword argument to only turn on this behaviour when needed.

### DIFF
--- a/python/amici/petab_import.py
+++ b/python/amici/petab_import.py
@@ -283,7 +283,7 @@ def import_petab_problem(
         os.makedirs(model_output_dir)
 
     # check if compilation necessary
-    if not _can_import_model(model_name) or force_compile:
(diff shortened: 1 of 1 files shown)


dweindl (approved):

dweindl:
> Is there any reason why a pre-existent model should not be loaded

The only reason is that it's hard to ensure that an existing model matches the possibly changed petab.Problem. But that is independent of the proposed changes...

dweindl:
@ failing doxygen:

 > error: Problems running gs. Check your installation!
 
Not sure what's going on there again. Can be ignored here.

yannikschaelte (approved):

PR #1384, opened 2021-01-29
Cleanup package building

dweindl:
* Use PEP 517 pyproject.toml (taking proper care of dependencies of setup.py itself)
* Use setup.cfg
* NOTE: creating a source distribution now requires https://github.com/pypa/build (or manual installation of the build requirements)

### DIFF
--- a/python/sdist/pyproject.toml
+++ b/python/sdist/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools>=40.6.3", "wheel"]
+requires = ["setuptools>=40.6.3", "wheel", "numpy"]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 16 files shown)


python/sdist/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools>=40.6.3", "wheel", "numpy"]
```
FFroehlich, line 2: shouldn't this allow us to remove a couple direct numpy installations in the CI workflow?

FFroehlich (approved):
nice 👍

python/sdist/pyproject.toml
dweindl: For setuptools/wheel yes, those I removed. Numpy was so far installed from within `setup`, which worked unreliably and is probably not so great practice. But I found two cases from ancient times - now removed.

PR #1392, opened 2021-01-30
Upgrade to sundials 5.7.0

dweindl:
* Upgrade to sundials 5.7.0
* Remove a load of unused sundials components


dweindl:
@FFroehlich Any idea what could be wrong with `test_compare_conservation_laws_sbml` (https://github.com/AMICI-dev/AMICI/runs/1797159532?check_suite_focus=true)?

ThirdParty/sundials/include/sundials/sundials_linearsolver.h
```diff
@@ -181,13 +183,15 @@ SUNDIALS_EXPORT int SUNLinSolFree(SUNLinearSolver S);
...
-#define SUNLS_VECTOROP_ERR       -810   /* vector operation error        */
+#define SUNLS_ATIMES_NULL        -804   /* atimes function is NULL       */
+#define SUNLS_ATIMES_FAIL_UNREC  -805   /* atimes unrecoverable failure  */
+#define SUNLS_PSET_FAIL_UNREC    -806   /* pset unrecoverable failure    */
+#define SUNLS_PSOLVE_NULL        -807   /* psolve function is NULL       */
+#define SUNLS_PSOLVE_FAIL_UNREC  -808   /* psolve unrecoverable failure  */
+#define SUNLS_PACKAGE_FAIL_UNREC -809   /* external package unrec. fail  */
+#define SUNLS_GS_FAIL            -810   /* Gram-Schmidt failure          */
+#define SUNLS_QRSOL_FAIL         -811   /* QRsol found singular R        */
+#define SUNLS_VECTOROP_ERR       -812   /* vector operation error        */
```
dweindl, line 194: Return codes for linear solvers have changed here, that is why `test_compare_conservation_laws_sbml` fails. @paulstapor or @FFroehlich , could you please have a look how `SteadystateProblem` and `defines.h` needs to be updated?

We should either get rid of the `AMICI_*` `define`s copying these values or adding respective static assertions somewhere, as done for other values.

FFroehlich (approved):
Nice 👍 , was deleting `python/tests/solverSettings.hdf5` intentional?

ThirdParty/sundials/include/sundials/sundials_linearsolver.h
FFroehlich: will do

dweindl:
> was deleting `python/tests/solverSettings.hdf5` intentional?

I wasn't aware of deleting it. But that file never should have been added to version control. So yes, intentional :)

ThirdParty/sundials/include/sundials/sundials_linearsolver.h
FFroehlich: done

Issue #1396, opened 2021-02-01
AMICI simulation fails when having initial states defined as NaN in the condition table


elbaraim:
I provide here a minimal example of code with a PEtab model in which when I set both initial conditions to numerical values in the condition table, it simulates and returns the expected value (the initial condition of the input).

However, if I set one or both to NaN (or empty string) it fails instead of simulation without doing assignments.

The minimal example reproduces the error scenario. Executing `test.py`.
[test.zip](https://github.com/AMICI-dev/AMICI/files/5905432/test.zip)

I have latest `develop` version.

dilpath:
Could you try the above PR with the following script (and the PEtab you supplied -- thanks for that)?

```python
import petab
from amici.petab_import import import_petab_problem
from amici.petab_objective import simulate_petab


petab_problem = petab.Problem.from_yaml("test/Pom1p_Problem.yml")
amici_model = import_petab_problem(petab_problem)
result = simulate_petab(petab_problem, amici_model)
for rdata in result['rdatas']:
    print(rdata['y'])

```

dweindl:
Related to #1319, #1345 and #1344.

dweindl:
Should be fixed in 0.11.13

Issue #1345, opened 2020-11-30
Allow enabling/disabling state reinitialization for individual states


dweindl:
So far, state-reinitialization can is only be switched on/off for all states (whose initial state depends on fixedParameters) at once.
Would be convenient, it this could be handled more dynamically by ignoring any reinitialization for the states for which x0 evaluates to NaN.

dweindl:
Turns out my original plan doesn't work out as smoothly as hoped for (started on #1344). The inconvenience is, that one would have to add one additional constant to the model for each state (or group of states) that one may want to conditionally reinitialize or not.

Not completely sure how the ideal interface for that would look like. Would be good if one could enable/disable state reinitialization for individual states via ExpData, and without adding additional model parameters/constants. Any ideas comments on how to best handle that?

paulstapor:
The problem is how the functions x0_fixedParameters and sx0_fixedParameters work, right?

dweindl:
> The problem is how the functions x0_fixedParameters and sx0_fixedParameters work, right?

Yes, the overall problem/motivation is that you can't use `x0_fixedParameters` to reinitialize varying sets of states. Either you reinitialize all states from `x0_fixedParameters` (with `reinitializeFixedParameterInitialStates=true`) or none (with `reinitializeFixedParameterInitialStates=false`), but nothing inbetween.

paulstapor:
well, obviously one could hack this by saving the old state and wrapping around the current code in C++. But this is cumbersome and error-prone. Better would be to add an "ix" variable to the headers of those two functions/files, implementing a switch statement in the files themselves, like this is already done for sensitivity function files with ip or event function files now with ie. So, we already have a very comparable concept for similar cases...
Obviously, this would break Matlab vs Python version and we would need another "if (python_generated)" in AMICI... But that would be fine from my side... Implementable in 2 roughly days, I'd say, including tests and code review. Pure coding time should be less...

Btw.: This functionality would also be highly welcome when dealing with adjoint preequilibration...

dweindl:
Oh, somehow I overlooked your reply @paulstapor . Sounds good, either some index list, or a boolean array. Maybe the former is better, as it's more in line with `plist` as you pointed out.

@FFroehlich - works for you? Any other suggestions?

FFroehlich:
Sounds good!

dweindl:
Just to make sure we are talking about the same thing. We would:

[...]

PR #1344, opened 2020-11-30
Extended PEtab test suite / cleanup PEtab state reinitialization

dweindl:
To include cases from https://github.com/PEtab-dev/petab_test_suite/pull/35

Related to #1319

PR #1397, opened 2021-02-01
Unspecified initial states in PEtab conditions table default to SBML initial value

dilpath:
https://github.com/AMICI-dev/AMICI/issues/1396

The warning added to "petab_import.py" is untested, just seems reasonable to me.

The warning added to "petab_objective.py" can result in many warnings, by default.

`float()` is used in `petab_objective.py` since any SBML initial assignment (for states in the PEtab conditions table) would have been overwritten in "petab_import.py" anyway, so I guess those overwritten SBML initial assignments should be ignored everywhere, hence only float-type values (initial amounts or concentrations) would be expected.

Not a nice solution I guess, since a private method from "sbml_import.py" is used.


FFroehlich (approved):
Is there an appropriate test case in the petab testsuite?

FFroehlich:
> Not a nice solution I guess, since a private method from "sbml_import.py" is used.

I don't think thats a big issue, feel free to make this a public method. I generally make all functions private by default as this slims the API and makes it easier to navigate.

dilpath:
> Is there an appropriate test case in the petab testsuite?

Not sure. @elbaraim could your PEtab problem from #1396 be added there?

python/amici/petab_objective.py
```diff
@@ -343,6 +344,17 @@ def _set_initial_concentration(condition_id, species_id, init_par_id,
                                        par_map, scale_map):
             value = petab.to_float_if_float(
                 petab_problem.condition_df.loc[condition_id, species_id])
+            if pd.isna(value):
+                value = float(
+                    get_species_initial(
+                        petab_problem.sbml_model.getSpecies(species_id)
+                    )
+                )
+                logger.warning(f'The species {species_id} has no initial value'
```
dweindl, line 353: Not sure I'd go for warnings, as this is explicitly a feature in PEtab. Might be confusing. Maybe change to info?

dilpath: Ah, maybe not worth logging then. Switched to debug but could also remove.

dweindl: I think debug can be helpful

python/amici/petab_import.py
```diff
@@ -532,6 +532,13 @@ def import_model_sbml(
         if assignment is None:
             assignment = sbml_model.createInitialAssignment()
             assignment.setSymbol(assignee_id)
+        else:
+            logger.warning('The SBML model has an initial assignment defined '
```
dweindl, line 536: same for this one, would also prefer info/debug

dweindl (approved):

elbaraim:
> > Is there an appropriate test case in the petab testsuite?
> 
> Not sure. @elbaraim could your PEtab problem from #1396 be added there?

The PEtab example I provided in the issue? Sure. Although I think an even easier model could be used with the same characteristics as in that one. What do you think is more appropriate?

elbaraim:
Hi @dilpath @dweindl , 
There is still an open issue in this PR:
If a preequilibration is defined and the initial state is left empty (or NaN) in the condition table, one would expect that the value that would be taken as initial would be the one obtained from the preequilibration and not from the SBML model.

For debugging: In the example I provided in #1396 one can see this unwanted behavior happening. The condition `unperturbed` is used as preequilibration condition but then the initial value (after preequilibration) is set to the initial value in the SBML (which is zero) instead of the last value in `sol['rdatas'].x_ss`.

```
print('before reinitialization (preeq)')
print(sol['rdatas'][0].x_ss[15])
print(sol['rdatas'][1].x_ss[15])
```
447796069.1255652
447796069.1255652
```
print('after reinitialization')
print(sol['rdatas'][0].x0[15])
print(sol['rdatas'][1].x0[15])
```
2.0
0.0

For the conditions left empty (in this case `unperturbed`) I would expect to get the same value in `x_ss` and `x0` (not zero in this case).
- Is this train of thoughts correct? Maybe I am just wrong 😅

dweindl:
>  If a preequilibration is defined and the initial state is left empty (or NaN) in the condition table, one would expect that the value that would be taken as initial would be the one obtained from the preequilibration and not from the SBML model.

Right. The handling of species in the condition table needs a makeover. There is meanwhile a better test case in the petab test suite and I started fixing stuff (#1344). The problem is, that #1345 needs to be addressed first to fully support those PEtab cases there. This is still some work. But we should at least raise some exception in case things are not fully supported.

dweindl:
Merging as is, even if it does not fully resolve @elbaraim 's issue. The rest will take a bit more time.

PR #1403, opened 2021-02-03
Python: Add numpy minimum version requirements

dweindl:
Should prevent errors like

```
RuntimeError: module compiled against API version 0xe but this version of numpy is 0xd
[...]
ImportError: numpy.core.multiarray failed to import
```


FFroehlich (approved):
👍

PR #1404, opened 2021-02-05
Add w to root signature

dilpath:


FFroehlich:
This will be a bit more tricky since 
* you have to touch the cpp code that calls the generated function
* the signature is now different from what is used in MATLAB so either that needs to be updated or we need another `pythonGenerated` switch

Alternatively, a simple `subs(zip(self._syms['w'], self._eqs['w']))` applied to the root function may do the trick.

FFroehlich (approved):

dilpath:
> Alternatively, a simple `subs(zip(self._syms['w'], self._eqs['w']))` applied to the root function may do the trick.

Thanks a lot for the quick fix! Works well in my case.

Issue #931, opened 2020-01-31
Allow directly setting presimulation and preequilibration parameters in model


FFroehlich:
Having to create and `Expdata` instance for is often a bit cumbersome and this may also ease the programmatic creation of ExpData using the model constructor.

dweindl:
What about just giving `Model` an `ExpData` member instead of replicating all the `ExpData` members?
This would also get rid of the need for the `ConditionContext` class, that I don't really like, due to the side effects in the destructor.

FFroehlich:
I see your point. My Primary issue is that interactive simulations are quite cumbersome at the moment. ExpData doesnt have any `ByName` setters or getters and the most of the members are immutable tuples, which effectively means that changing a single value typically involves 3-4 lines of code compared to a single call when setting variables directly in the model.

dweindl:
Yeah, could be more comfortable, agreed.

> ExpData doesnt have any ByName setters or getters and the most of the members are immutable tuples

Let's add them? Pull the ids/names and some of the dimensions out of `Model` into some `ModelInfo`. Both `Model` and `ExpData` can have one of those then. The getters/setters we can move to `ExpData`, so that you do `model->edata->setParameterById()`, or `model->edata->setMeasurementForObservable(id, values)`.
Having all those things both as part of `Model` and `ExpData` feels wrong and like a maintaince struggle.

Having names and IDs in a separate object would also be quite convenient for tying that info onto `ReturnData`. Allowing for `rdata->getResiduals(observableId, time)` could also be quite nice...

dweindl:
Maybe off-topic with respect to setting parameters, but regarding convenient interactive use of `ExpData`:
If find it currently pretty inconvenient that I am expected to decide on the number of time points / replicates during the creation of ExpData. Would be great if one could just append data for a given observable for arbitrary timepoints.

FFroehlich:
> Yeah, could be more comfortable, agreed.
> 
> > ExpData doesnt have any ByName setters or getters and the most of the members are immutable tuples
> 
> Let's add them? Pull the ids/names and some of the dimensions out of `Model` into some `ModelInfo`. Both `Model` and `ExpData` can have one of those then. The getters/setters we can move to `ExpData`, so that you do `model->edata->setParameterById()`, or `model->edata->setMeasurementForObservable(id, values)`.
> Having all those things both as part of `Model` and `ExpData` feels wrong and like a maintenance struggle.

[...]

PR #1407, opened 2021-02-10
Refactor(C++): Extract SimulationParameters from Model and ExpData

dweindl:
This allows using Model without ExpData unless there really are measurements. Also avoids having to add any new member of ExpData also to Model and slims down Model a bit. Getters and setters for SimulationParameters members are still in Model, which is to preserve backwards compatibility for ExpData, which allowed unchecked access to the respective members. Further refactoring to follow.

Partially addresses #931 (moving ID / name functions is still missing).


FFroehlich (commented):
👍

PR #1409, opened 2021-02-11
Doc: Add table with AMICI C++/Python/Matlab feature comparison

dweindl:


documentation/about.rst
```diff
@@ -57,3 +57,53 @@ simulation.
...
+     - yes
+   * - Second-order sensitivities
+     - no
+     - no
+     - yes
+   * - Events
+     - no
+     - no
+     - yes
+   * - Preequilibration
```
FFroehlich, line 102:
```suggestion
   * - :term:`preequilibration`
```

documentation/about.rst
```diff
@@ -57,3 +57,53 @@ simulation.
...
+     - yes
+     - yes
+     - yes
+   * - Presimulation
```
FFroehlich, line 106:
```suggestion
   * - :term:`presimulation`
```

documentation/about.rst
dweindl: Oh yeah, forgot that we had that already in the glossary. :+1:

FFroehlich (approved):

PR #1414, opened 2021-02-12
CI: Allow tests on unstable python version to fail

dweindl:


FFroehlich (approved):

dweindl:
For some reason, the python 3.10 job is running for hours (https://github.com/AMICI-dev/AMICI/runs/1888130393?check_suite_focus=true). Will remove for now.

Issue #1319, opened 2020-10-23
Initial concentrations not overwritten after preequilibration


LeonardSchmiester:
**What did you expect to happen?**
I have a model, where I want to reset initial concentrations for some species (in my case different drugs) after preequilibration. In the SBML file, I have an initial assignment to the species, which is set to 0 for the preeq condition in the PEtab condition file and then to e.g. 10 for the simulation condition. I would have expected `x0` in `rdata` for this species to be 10.

**What has happened instead?**
It seems like the initial concentration of the species is not overwritten with the value in the condition file. So in the above example `x0` is 0. If I delete the preeq condition, `x0` is set correctly.

**To Reproduce**
Here is a minimal example to reproduce the problem: [minimal_example.tar.gz](https://github.com/AMICI-dev/AMICI/files/5428256/minimal_example.tar.gz)

**AMICI version and system environment**
 - OS and version: Ubuntu
 - AMICI interface: Python
 - AMICI version: 0.11.8

paulstapor:
Might be that this has something to do with `model.setReinitializeFixedParameterInitialStates()`? If this flag isn't set to `True`, model simulation will show exactly the behavior you described...

LeonardSchmiester:
Thanks for the hint @paulstapor . This indeed fixed the problem for the minimal example. Will also try for the large model now.

dweindl:
Okay, the problem here is that `simulate_petab` does not handle that correctly for `initialAssignment`s, but only if you have a species name in the condition table.

dweindl:
I am not completely sure how that should be handled (-> https://github.com/PEtab-dev/PEtab/issues/481).

LeonardSchmiester:
> Okay, the problem here is that `simulate_petab` does not handle that correctly for `initialAssignment`s, but only if you have a species name in the condition table.

Using the species Id indeed gives the results I expected.

dweindl:
As this turns out to be slightly more complicated (https://github.com/PEtab-dev/PEtab/issues/481), I'd suggest to only add support for simple initial assignments for now, i.e. those with math elements consisting of a single parameter that is not a target of other rules. The rest should trigger a `NotImplementedError`. Will fix this problem here. Anything beyond can be addressed elsewhere.

dweindl:
Should be fixed in 0.11.13

PR #1417, opened 2021-02-12
More flexible state reinitialization / PEtab import reinitialization fixes

dweindl:
Adds possibility to provide state indices for selective reinitialization based on fixed parameters.

The previous `ExpData::reinitializeFixedParameterInitialStates` is still there, but will be removed in an upcoming version.

Addresses #1345, #1396, #1319

Supersedes #1344 

Include new test cases from https://github.com/PEtab-dev/petab_test_suite/pull/35


src/edata.cpp
```diff
@@ -409,6 +416,9 @@ void ConditionContext::applyCondition(const ExpData *edata,
                                  "not match ExpData (%zd).",
                                  model_->nk(), edata->fixedParameters.size());
           model_->setFixedParameters(edata->fixedParameters);
+          if(!edata->reinitializeFixedParameterInitialStates)
```
FFroehlich, line 419: I am confused, don't we want to check for `edata->reinitializeFixedParameterInitialStates` here?

FFroehlich (approved):

src/edata.cpp
dweindl: The thing is, that if `edata->reinitializeFixedParameterInitialStates == true`, then this has been set on Model a couple of lines before. This will populate the index array in Model. If we would execute the next line, this would overwrite them with an empty array.
Not great, but keeps backward compatibility for the moment and still allow selective reinitialization. To be cleaned up soon.

PR #1420, opened 2021-02-16
CI: SBML test suite: show tag summary for passed tests

dweindl:
Include supported tags (i.e. at least 1 passed test with the given tag) in sbml semantic test suite summary.
Description of the tags is available at https://github.com/sbmlteam/sbml-test-suite/blob/master/docs/tags-documentation/all-tags.txt


FFroehlich:
Pretty sure we do not support `DelayInEventAssignment`

dweindl:
> Pretty sure we do not support `DelayInEventAssignment` and others, there still is something fishy with the analysis.

I'd agree :D 

Should be fixed now.

FFroehlich (approved):
Looks good to me now!

PR #1427, opened 2021-02-18
pysb benchmark

FFroehlich:
benchmark_pysb


python/benchmark/benchmark_pysb.py
```diff
@@ -0,0 +1,104 @@
+import os
```
dweindl, line 1: Add file docstring with brief overview?

dweindl (approved):

PR #1439, opened 2021-02-26
Micro-optimize SUNMatrixWrapper::transpose

dweindl:
Quick test with a large model (nx=3665) shows up to 20% reduction in cpu_time and cpu_timeB

### DIFF
--- a/src/sundials_matrix_wrapper.cpp
+++ b/src/sundials_matrix_wrapper.cpp
@@ -629,14 +629,15 @@ void SUNMatrixWrapper::transpose(SUNMatrixWrapper &C, const realtype alpha,
     if (!matrix_ || !C.matrix_)
         return;
 
-    if (!((C.matrix_id() == SUNMATRIX_SPARSE && C.sparsetype() == CSC_MAT)
-          || C.matrix_id() == SUNMATRIX_DENSE))
+    auto C_matrix_id = C.matrix_id();
+    if (!((C_matrix_id == SUNMATRIX_SPARSE && C.sparsetype() == CSC_MAT)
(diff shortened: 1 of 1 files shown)


src/sundials_matrix_wrapper.cpp
```diff
@@ -651,40 +652,38 @@ void SUNMatrixWrapper::transpose(SUNMatrixWrapper &C, const realtype alpha,
...
-            for (aidx = get_indexptr(acol);
-                 aidx < get_indexptr(acol+1); aidx++) {
-                widx = (acol/blocksize)*blocksize + get_indexval(aidx) % blocksize;
-                w.at(widx)++;
-                assert(w[widx] <= rows());
+        for (sunindextype acol = 0; acol < nrows; acol++) { /* row counts */
+            auto next_indexptr = get_indexptr(acol+1);
+            for (sunindextype aidx = get_indexptr(acol);
+                 aidx < next_indexptr; aidx++) {
+                sunindextype widx = (acol/blocksize)*blocksize + get_indexval(aidx) % blocksize;
```
dweindl, line 663: @FFroehlich: I was wondering if all those `* blocksize / blocksize` are helpful here in any way

dweindl: Ah, all integers. Forget it.

FFroehlich (approved):
👍  wonder whether it makes sense to replace other `.at(...)` in the same way.

dweindl:
> wonder whether it makes sense to replace other `.at(...)` in the same way.

In several places it would, yes.

PR #1443, opened 2021-03-08
Event implementation

paulstapor:
Implementation of events on python side. Functionality restricted to models without conservation laws for the moment, as otherwise, also the conserved quantities must be considered during an event. Not sure whether this is really nicely captured in sbml_importer... Maybe a second test model with nested events could be implemented in test_events... O(pi)nions, please...


python/amici/sbml_import.py
```diff
@@ -869,6 +870,133 @@ def _process_time(self) -> None:
...
+                )
+            # FIXME check what AMICI supports by default
+            # True means event does not executed if triggered at time == 0
+            # False means it does.
+            if not trigger_sbml.getInitialValue():
+                raise SBMLException(
+                    f'Event {event_id} has a trigger that has an initial '
+                    'value of False. This is currently not supported in AMICI.'
+                )
+            if not trigger_sbml.getInitialValue():
```
FFroehlich, line 958: duplicate code?

python/amici/sbml_import.py
```diff
@@ -869,6 +870,133 @@ def _process_time(self) -> None:
...
+                raise SBMLException(
+                    f'Event {event_id} has a trigger that has an initial '
+                    'value of False. This is currently not supported in AMICI.'
+                )
+            trigger_sym = self._sympy_from_sbml_math(trigger_sbml)
+            trigger = _parse_event_trigger(trigger_sym)
+
+            # FIXME?: Currently, all event assignment targets must exist in
+            # self.symbols[SymbolId.SPECIES]
+            state_vector = list(self.symbols[SymbolId.SPECIES].keys())
```
FFroehlich, line 968: to make this work, we will have to convert parameters that get updated in events to species.

python/amici/sbml_import.py
```diff
@@ -869,6 +870,133 @@ def _process_time(self) -> None:
...
+
+            bolus = [sp.Symbol('0') for _ in state_vector]
+
+            event_assignments = event.getListOfEventAssignments()
+            if not len(event_assignments):
+                # ignore events with no event assignments
+                # TODO raise exception?
```
FFroehlich, line 975: I think continue is fine here.

python/amici/sbml_import.py
```diff
@@ -869,6 +870,133 @@ def _process_time(self) -> None:
...
+                    # ignore event assignments with no change in value
+                    continue
+                formula = self._sympy_from_sbml_math(event_assignment)
+                bolus[state_vector.index(variable_sym)] = \
+                    formula - variable_sym
+
+            self.symbols[SymbolId.EVENT][event_sym] = {
+                'name': event_id,
+                'value': trigger,
+                'state_update': sp.MutableDenseMatrix(bolus), # 'starts_in_event': trigger_sbml.getInitialValue(),
```
FFroehlich, line 995: remove comment?

python/amici/sbml_import.py
```diff
@@ -1640,15 +1775,56 @@ def _parse_trigger(trigger: sp.Expr) -> sp.Expr:
...
+                isinstance(trigger, sp.core.relational.StrictGreaterThan):
+            # y >= x or y > x
+            return root
+
+    # or(x,y) = not(and(not(x),not(y))
+    if isinstance(trigger, sp.Or):
+        return sp.Mul(*[_parse_event_trigger(arg) for arg in trigger.args])
+
+    if isinstance(trigger, sp.And):
+        raise SBMLException('AMICI can not parse logical AND expressions in '
```
FFroehlich, line 1821: why can't we use the same implementation for `_parse_heaviside_trigger`?

python/tests/test_events.py
```diff
@@ -0,0 +1,366 @@
...
+
+def create_amici_model(sbml_model, model_name) -> AmiciModel:
+    sbml_importer = SbmlImporter(sbml_model)
+    output_dir = sbml_test_models_output_dir / model_name
+    sbml_importer.sbml2amici(
+        model_name=model_name,
+        output_dir=str(output_dir)
+    )
+
+    sys.path.insert(0, str(output_dir.resolve()))
```
FFroehlich, line 118: please use `amici.import_model_module` instead.

python/tests/test_events.py
```diff
@@ -0,0 +1,366 @@
...
+    model_module = importlib.import_module(model_name)
+
+    model = model_module.getModel()
+    return model
+
+
+def create_sbml_model(
```
FFroehlich, line 125: I appreciate the time and effort you put into this, but what about using `https://simplesbml.readthedocs.io/en/latest/` instead of reinventing the wheel?

python/tests/test_events.py
```diff
@@ -0,0 +1,366 @@
...
+
+    # Need to return document, else AMICI throws an error.
+    # (possibly due to garbage collection?)
+    return document, model
+
+
+def get_model_definition(model_name):
+    if model_name == 'events_plus_heavisides':
+        return model_definition_events_plus_heavisides()
+#    elif model_name == 'simple_event':
```
FFroehlich, line 204: don't comment this out?

FFroehlich (changes requested):
Great work, looks likes only a bit of finetuning will be required to get this done!

FFroehlich:
> Implementation of events on python side. Functionality restricted to models without conservation laws for the moment, as otherwise, also the conserved quantities must be considered during an event. Not sure whether this is really nicely captured in sbml_importer... Maybe a second test model with nested events could be implemented in test_events... O(pi)nions, please...

Yes, I also wasn't sure how to deal with conservation laws. My thinking was that we primarily need them when computing steady-states, which won't work very well with events anyways. So it probably makes more sense to simply disable conservation laws for states that are updated by events?

python/amici/ode_export.py
```diff
@@ -1580,8 +1606,10 @@ def _generate_symbol(self, name: str, *, from_sbml: bool = False) -> None:
         elif name in sensi_functions:
             length = self.eq(name).shape[0]
         else:
-            length = len(self.eq(name))
-
+            try:
+                length = len(self.eq(name))
+            except:
+                print("Don't piss me off")
```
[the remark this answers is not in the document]
paulstapor: Okay, that was some debugging code I forgot... Whoopsie...

paulstapor:
> Yes, I also wasn't sure how to deal with conservation laws. My thinking was that we primarily need them when computing steady-states, which won't work very well with events anyways. So it probably makes more sense to simply disable conservation laws for states that are updated by events?

Currently, conservation laws are either completely switched on or off via one global flag, it's not state-wise (at least not right now). As already one conserved state will kill the Jacobian, one really wants to take care of all or none, so I think this is fine for the moment...
One could simply make the conservation_law flag dependent on the existence of events. It will already be automatically switched off, if a model only has fixed state variables (yes, the SBML test suite covers this case :roll_eyes: ).

In the long run, it might be cool to have also coverage for conservation laws, as this would allow to switch the condition context under the hood. But nor right now...

python/amici/sbml_import.py
paulstapor: That's a bit tricky... For Heavisides, we really have to discern "<" and "<=" or ">" and ">=", which will yield different Heaviside functions. For events, we only have one trigger, which doesn't care about this. Moreover, the Heaviside-parser directly clutches a Heaviside function at the trigger, which we don't want for events. So there are slight little difference. I will see, to which degree we can reuse things, but that's not guaranteed to work...

python/amici/ode_export.py
```diff
@@ -1835,19 +1867,75 @@ def _compute_equation(self, name: str) -> None:
...
+            ]
+
+        elif name == 'deltasx':
+            event_eqs = []
+            for ie in range(self.num_events()):
+                if self._events[ie]._state_update is not None:
+                    # ====== chain rule for the state variables ===============
+                    # get xdot with expressions back-substituted
+                    tmp_eq = smart_multiply(
+                        (self.eq('xdot_old') - self.sym('xdot')),
```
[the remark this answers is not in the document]
paulstapor: Yes, it's identical. Can do so, no prob.

python/amici/sbml_import.py
```diff
@@ -869,6 +870,133 @@ def _process_time(self) -> None:
...
+            event_sym = sp.Symbol(event_id)
+
+
+            # TODO move these checks to the `check_support` method?
+            delay = event.getDelay()
+            if delay is not None:
+                try:
+                    delay_time = float(self._sympy_from_sbml_math(delay))
+                    if delay_time != 0:
+                        # `TypeError` would be raised in the above `float(...)`
```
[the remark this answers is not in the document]
paulstapor: How about an SBML Exception, as this is about non-supported SBML features?

python/amici/sbml_import.py
```diff
@@ -869,6 +870,133 @@ def _process_time(self) -> None:
...
+                raise SBMLException(f'Event {event_id} has a priority '
+                                    'specified. This is currently not '
+                                    'supported in AMICI.')
+
+            trigger_sbml = event.getTrigger()
+            if trigger_sbml is None:
+                # TODO raise SBMLException? Or just ignore?
+                continue
+            if trigger_sbml.getMath() is None:
+                # TODO untested, find SBML test suite case
```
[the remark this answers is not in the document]
paulstapor: Have to look, might very well be done.

python/amici/sbml_import.py
paulstapor: Talking about parameters (with/without assignment rules), which are targets of event assignments?
If so, can do...

python/amici/sbml_import.py
paulstapor: Yes, makes sense to allow void assignments, as we sometimes will just want to get some read out.

python/tests/test_events.py
paulstapor: arg, yes

python/tests/test_events.py
paulstapor: That function was just cut-copy-pasted from the Heaviside test-case... :D
Doesn't make things better, true. But Here I really didn't put much effort. Can slim things down though by using simplesbml.

python/tests/test_events.py
paulstapor: That test case isn't yet implemented. I wonder whether we need it at all, as the above one really covers many weird cases... I might rather implement another case with nested events. This would tests some further functionality.

paulstapor:
I also have a question on logical operations in event triggers. Those are typically really nasty to convert into correct triggers... 
E.g., in the case of an "A or B", we need positive roots for each of the expressions "A" and "B". Which is tricky: Imagine the trigger just switched from negative to positive due to A, then it needs to become negative again to allow a positive crossing for B...
On could maybe implement this as two events, with the same targets and assignments, just having the triggers splitted up... Not sure whether this is the best way though...
For "and"s, we could use something like 
`sympy.Mult(sympy.sign(sympy.min(*[triggers])), sympy.add([sympy.abs(trigger) for trigger in trggers]))`
However, that's somewhat nasty...

paulstapor:
I tried to address some first review comments. I deleted SBMLExceptions for all the `persistent`, ùseValuesFromTriggerTime`, and so on... These things will have no effect as long as no delay is used (or the delay is 0), hence it makes no sense to raise if they are used but no delay is modelled...

python/amici/sbml_import.py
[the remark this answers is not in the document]
FFroehlich: Sure, also fine. I thought the idea was to pipe both exceptions into a single SBMLException to avoid duplication.

python/amici/sbml_import.py
FFroehlich: > Talking about parameters (with/without assignment rules), which are targets of event assignments?
> If so, can do...

I think any parameters that do not have rate rules will not have a `SymbolId.SPECIES` implementation. So we will need something similar to `https://github.com/AMICI-dev/AMICI/blob/b090fbb34ee64e65522cef0a2c0d639e70cfbcfe/python/amici/sbml_import.py#L650` here.

python/tests/test_events.py
FFroehlich: if its a simple copy paste, please import instead. Not necessary to replace this now, but maybe you can open an issue that this needs to be fixed in the future?

python/tests/test_events.py
FFroehlich: fine, I thought there might have already been an implementation somewhere else. Simply delete in that case.

FFroehlich:
> I also have a question on logical operations in event triggers. Those are typically really nasty to convert into correct triggers...
> E.g., in the case of an "A or B", we need positive roots for each of the expressions "A" and "B". Which is tricky: Imagine the trigger just switched from negative to positive due to A, then it needs to become negative again to allow a positive crossing for B...
> On could maybe implement this as two events, with the same targets and assignments, just having the triggers splitted up... Not sure whether this is the best way though...
> For "and"s, we could use something like
> `sympy.Mult(sympy.sign(sympy.min(*[triggers])), sympy.add([sympy.abs(trigger) for trigger in trggers]))`
> However, that's somewhat nasty...

Ah sorry, was a bit confused there. I think this should work by simply implementing implementing `or` as `sp.Min` and `and` as `sp.Max` (or `-sp.Min` in analogy to the Heaviside implementation). At least thats what I remember implementing in the MATLAB interface.

python/amici/sbml_import.py
```diff
@@ -1640,15 +1775,56 @@ def _parse_trigger(trigger: sp.Expr) -> sp.Expr:
...
+    Recursively translates a boolean trigger function into a real valued
+    root function
+
+    :param trigger:
+    :return: real valued root function expression
+    """
+    if trigger.is_Relational:
+        root = trigger.args[0] - trigger.args[1]
+
+        # normalize such that we always implement <,
```
FFroehlich, line 1781: remove comment as this doesn't fit the implementation here

python/amici/sbml_import.py
FFroehlich: Sorry, yeah missed the point here, but I think both `and` and `or` should be implementable using `sp.min` and `sp.max`, see MATLAB interface and comment below.

python/amici/sbml_import.py
```diff
@@ -420,6 +418,40 @@ def check_support(self) -> None:
         if any([r.getFast() for r in self.sbml.getListOfReactions()]):
             raise SBMLException('Fast reactions are currently not supported!')
 
+        # check support for events
+        for event in self.sbml.getListOfRules():
```
dweindl, line 422:
```suggestion
        for event in self.sbml.getListOfEvents():
```
?

dweindl:
Now we can import too many models :D. The SBML test suite takes more than the available 6 hours. As short-term solution, I'd suggest to split things into two jobs. Medium-term solution would be improving import performance for tiny models.

paulstapor:
> Now we can import too many models :D. The SBML test suite takes more than the available 6 hours. As short-term solution, I'd suggest to split things into multiple jobs (#1452). Medium-term solution would be improving import performance for tiny models (#1453).

That sounds like a problem I like to have... :D
Thanks for taking care of that!

python/amici/sbml_import.py
```diff
@@ -420,6 +418,52 @@ def check_support(self) -> None:
...
 
+        # Check events for unsupported functionality
+        self.check_event_support()
+
+    def check_event_support(self) -> None:
+        """
+        Check possible events in the model, as AMICI does currently not support
+            - delays in events
+            - priorities of events
+            - events fired at initial time
```
dweindl, line 429:
```suggestion
        Check possible events in the model, as AMICI does currently not support
        
        * delays in events
        * priorities of events
        * events fired at initial time
```

sphinx

dweindl:
> Now we can import too many models :D. The SBML test suite takes more than the available 6 hours. As short-term solution, I'd suggest to split things into multiple jobs (#1452). Medium-term solution would be improving import performance for tiny models (#1453).

There is still something fishy. Not even the first 600 finish within 6h. That doesn't seem to make sense. Some infinite loops somewhere? :grimacing:

python/tests/test_heavisides.py
```diff
@@ -68,120 +61,12 @@ def test_models(model):
...
-    if to_file:
-        libsbml.writeSBMLToFile(
-            document,
-            str(to_file),
-        )
-
-    # Need to return document, else AMICI throws an error.
-    # (possibly due to garbage collection?)
-    return document, model
+    check_trajectories_without_sensitivities(amici_model,
```
FFroehlich, line 65: 👍

FFroehlich (approved):
Looks good from my end. Would keep `event_observable` stuff out for now since its currently defunct, right?

python/amici/sbml_import.py
```diff
@@ -884,6 +937,98 @@ def _process_time(self) -> None:
...
+                    'that affect parameters that are also the target of '
+                    'assignment rules.'
+                )
+            parameter_def = None
+            for symbol_id in {SymbolId.PARAMETER, SymbolId.FIXED_PARAMETER}:
+                if parameter_target in self.symbols[symbol_id]:
+                    # `parameter_target` should only exist in one of the
+                    # `symbol_id` dictionaries.
+                    if parameter_def is not None:
+                        raise KeyError(
```
dweindl, line 965:
```suggestion
                        raise AssertionError(
```
Seems more appropriate.

python/amici/sbml_import.py
```diff
@@ -884,6 +937,98 @@ def _process_time(self) -> None:
...
+                            'Unexpected error. The parameter target of an '
+                            'event assignment was processed twice.'
+                        )
+                    parameter_def = \
+                        self.symbols[symbol_id].pop(parameter_target)
+            if parameter_def is None:
+                raise KeyError(
```
dweindl, line 972:
```suggestion
                raise AssertionError(
```

python/amici/sbml_import.py
```diff
@@ -884,6 +937,98 @@ def _process_time(self) -> None:
...
+                )
+            # Fixed parameters are added as species such that they can be
+            # targets of events.
+            self.symbols[SymbolId.SPECIES][parameter_target] = {
+                'name': parameter_def['name'],
+                'init': sp.Float(parameter_def['value']),
+                #'compartment': None,  # can ignore for amounts
+                'constant': False,
+                'amount': True,
+                #'conversion_factor': 1.0,  # probably can be ignored
```
dweindl, line 984: Probably? :)

python/amici/sbml_import.py
```diff
@@ -884,6 +937,98 @@ def _process_time(self) -> None:
...
+            trigger_sbml = event.getTrigger()
+            trigger_sym = self._sympy_from_sbml_math(trigger_sbml)
+            trigger = _parse_event_trigger(trigger_sym)
+
+            # Currently, all event assignment targets must exist in
+            # self.symbols[SymbolId.SPECIES]
+            state_vector = list(self.symbols[SymbolId.SPECIES].keys())
+
+            # parse the boluses / event assignments
+            bolus = [sp.Symbol('0') for _ in state_vector]
```
dweindl, line 1012: Intended to be a symbol named `0`, instead of a value of `sp.Float(0.0)`?

src/forwardproblem.cpp
```diff
@@ -158,8 +158,9 @@ void ForwardProblem::handleEvent(realtype *tlastroot, const bool seflag) {
          * be true */
         if (t_ == *tlastroot) {
             throw AmiException("AMICI is stuck in an event, as the initial"
```
dweindl, line 160:
```suggestion
            throw AmiException("AMICI is stuck in an event, as the initial "
```

python/amici/sbml_import.py
paulstapor: It at least works in the current version, but agree that the latter is at least more intuitive. Will change.

dweindl: It should work, because the symbol is just printed as `0` to the cpp files where it is interpreted as a number, but any `is_zero` will probably not yield the expected result...

python/amici/ode_export.py
```diff
@@ -1846,19 +1882,72 @@ def _compute_equation(self, name: str) -> None:
...
             ]
 
+        elif name == 'deltasx':
+            event_eqs = []
+            for ie, event in enumerate(self._events):
+                if event._state_update is not None:
+                    # ====== chain rule for the state variables ===============
+                    # get xdot with expressions back-substituted
+                    tmp_eq = smart_multiply(
+                        (self.sym('xdot_old') - self.sym('xdot')),
```
FFroehlich, line 1928: sorry, I think this was an error on my end, this actually has to be `self.eq('xdot')`, not `self.sym('xdot')`, this may also solve the issues with nested events.

python/amici/ode_export.py
```diff
@@ -1846,19 +1882,72 @@ def _compute_equation(self, name: str) -> None:
...
+                        self.eq('stau')[ie])
+                    # construct an enhanced state sensitivity, which accounts
+                    # for the time point sensitivity as well
+                    tmp_dxdp = self.sym('sx') * sp.ones(1, self.num_par())
+                    tmp_dxdp += smart_multiply(self.sym('xdot'),
```
FFroehlich, line 1933:
```suggestion
                    tmp_dxdp += smart_multiply(self.eq('xdot'),
```

python/amici/ode_export.py
FFroehlich, line 1928:
```suggestion
                        (self.sym('xdot_old') - self.eq('xdot')),
```

python/amici/ode_export.py
```diff
@@ -1846,19 +1882,72 @@ def _compute_equation(self, name: str) -> None:
...
+                    tmp_eq += smart_multiply(self.eq('ddeltaxdx')[ie],
+                                             tmp_dxdp)
+                    # ====== chain rule for the time point ====================
+                    tmp_eq += smart_multiply(self.eq('ddeltaxdt')[ie],
+                                             self.eq('stau')[ie])
+                    # ====== partial derivative for the parameters ============
+                    tmp_eq += self.eq('ddeltaxdp')[ie]
+                else:
+                    tmp_eq = smart_multiply(
+                        (self.eq('xdot_old') - self.eq('xdot')),
```
FFroehlich, line 1944:
```suggestion
                        (self.sym('xdot_old') - self.eq('xdot')),
```

python/amici/sbml_import.py
```diff
@@ -468,7 +468,7 @@ def check_event_support(self) -> None:
 
             # check trigger
             trigger_sbml = event.getTrigger()
-            if trigger_sbml.getMath() is None:
+            if trigger_sbml is None or trigger_sbml.getMath() is None:
                 logger.warning(f'Event {event_id} trigger has no trigger '
                                'expression. Ignoring event...')
```
dilpath, line 473:
```suggestion
                               'expression, so a dummy trigger will be set.')
```

I guess this is now the case?

python/amici/ode_export.py
paulstapor: I used to have "eq" here, and this was then wrong for nested events, also on the level of state variables, not only sensitivities... So no, I guess `sym` is definitely more correct than `eq`.

paulstapor: I just see that you suggested multiple changes... Can try, whether all of them together will change things...
Let's give it a try and see...

paulstapor: Arg, nope. Just implemented your suggestions and reran the tests. Both event test models fail now...
So: undo?

python/amici/ode_export.py
paulstapor: This will do nothing, as for "xdot_old" eq == sym

paulstapor:
I just reverted the changes. However, I will need some help with the SMBL test suite... I'm really not sure what's failing there and why things take so much time...

dweindl:
> However, I will need some help with the SMBL test suite... I'm really not sure what's failing there and why things take so much time...

354 to 384 did not complete. 369 fails. Can you fix 369 and see if the problem persists? If so, check 354-384.

python/amici/ode_export.py
FFroehlich: Ah, yes. Looked at the matlab implementation and that also uses the strsyms. I think this has something to do at which point this difference is computed, which makes sense that this becomes relevant for nested events.

python/amici/ode_export.py
```diff
@@ -1846,19 +1882,72 @@ def _compute_equation(self, name: str) -> None:
...
                 for ie in range(self.num_events())
             ]
 
+        elif name == 'deltasx':
+            event_eqs = []
+            for ie, event in enumerate(self._events):
+                if event._state_update is not None:
+                    # ====== chain rule for the state variables ===============
+                    # get xdot with expressions back-substituted
+                    tmp_eq = smart_multiply(
```
FFroehlich, line 1927:
```
tmp_eq = smart_multiply((self.eq('xdot_old') - self.eq('xdot')),
                                           self.eq('stau')[ie])
```
Is added independent of `event._state_update`, so this could be simplified to avoid duplication.

FFroehlich:
Also note that documenation needs to be updated in various places to reflect this is supported now @dweindl could you look into that as soon as tests pass?

python/amici/sbml_import.py
```diff
@@ -467,7 +467,11 @@ def check_event_support(self) -> None:
 
             # check trigger
             trigger_sbml = event.getTrigger()
-            if trigger_sbml is None or trigger_sbml.getMath() is None:
+            if trigger_sbml is None:
+                logger.warning(f'Event {event_id} trigger has no trigger, '
+                               'so a dummy trigger will be set.')
+                return
```
dilpath, line 473:
```suggestion
            if trigger_sbml is None:
                logger.warning(f'Event {event_id} trigger has no trigger, '
                               'so will be skipped.')
                continue
```

python/amici/ode_export.py
paulstapor: Will do asap. But, currently, my main concern is to get the test suite passing... I prefer that over code optimization for the moment...

FFroehlich:
> Also note that documenation needs to be updated in various places to reflect this is supported now @dweindl could you look into that as soon as tests pass?

Also don't forget number of passed tests

tests/testSBMLSuite.py
```diff
@@ -80,6 +80,8 @@ def test_sbml_testsuite_case(test_number, result_path):
 
         # simulate model
         rdata = amici.runAmiciSimulation(model, solver)
+        if rdata['status'] != amici.AMICI_SUCCESS:
+            raise amici.sbml_import.SBMLException('Simulation Failed')
```
dweindl, line 84: @FFroehlich : This may easily cover up future unexpected failures. What about introducing/handling more specific status codes for these event-related issues?

FFroehlich: That just moves the problem further down the road. Add explicit check for test id so this is only allowed for prespecified tests.

dweindl: Thanks. Works for me.

PR #1447, opened 2021-03-08
Fix Matlab interface (diagnostics dimensions)

dweindl:
Got broken in #1438, because matlab and python have different expectation on field lengths.
Matlab expects values only for finite timepoints, python for all timepoints.


FFroehlich (approved):
💩

PR #1454, opened 2021-03-09
Implement support for compilation without sensitivities

FFroehlich:
Unfortunately doesn't seem to yield much of an improvement in import time.


python/amici/sbml_import.py
```diff
@@ -272,21 +273,27 @@ def sbml2amici(self,
...
 
         :param simplify:
             see :attr:`ODEModel._simplify`
 
         :param log_as_log10:
-            If True, log in the SBML model will be parsed as `log10` (default),
-            if False, log will be parsed as natural logarithm `ln`
+            If ``True``, log in the SBML model will be parsed as `log10`
+            (default), if ``False``, log will be parsed as natural logarithm
+            `ln`
```
dweindl, line 292:
```suggestion
            If ``True``, log in the SBML model will be parsed as ``log10``
            (default), if ``False``, log will be parsed as natural logarithm
            ``ln`
```

dweindl (approved):
Great!

dweindl:
> Unfortunately doesn't seem to yield much of an improvement in import time.

Hm, for which model? For large ones, I'd expect a significant improvement.

python/tests/test_sbml_import.py
```diff
@@ -46,6 +50,35 @@ def test_sbml2amici_no_observables(simple_sbml_model):
...
+        solver.setSensitivityOrder(amici.SensitivityOrder.first)
+        solver.setSensitivityMethod(amici.SensitivityMethod.forward)
+        with pytest.raises(RuntimeError):
+            amici.runAmiciSimulation(model, solver)
+
+        # windows acting up
+        del model_module
+        del model
+        del solver
+
```
dweindl, line 80:
```suggestion
        import modules
        del sys.modules['_test']
```
?

PR #1459, opened 2021-03-14
Fix Crauste ref val, fixes #1458

### LINKED ISSUE #1458: LLH reference value for Crauste_CellSystems2017
benchmark_models.yaml reports `190.96398465243203` but AMICI returns `-190.96521897435176` 🤔

yannikschaelte:


FFroehlich (approved):
Did you confirm there was a sign error in the "analytical" solution?

PR #1461, opened 2021-03-15
Fix 1422

### LINKED ISSUE #1422: Steadystate solver fails if preequilibration starts in steadystate
E.g. preequilibration for the conversion reaction example with initial states [0, 0] fails:

```
    solver.setSensitivityMethod(amici.SensitivityMethod_forward)
    solver.setSensitivityOrder(amici.SensitivityOrder_first)
    model.setSteadyStateSensitivityMode(
        SteadyStateSensitivityMode_simulationFSA)
```

> Steady state sensitivity computation failed due to unsuccessful factorization of RHS Jacobian

preeq_status: [[1 0 0]]

preeq_numsteps [[0 0 0]]

x_ss and sx_ss seem to be set correctly, though.

paulstapor:
fixes #1422

### DIFF
--- a/src/steadystateproblem.cpp
+++ b/src/steadystateproblem.cpp
@@ -66,24 +66,19 @@ void SteadystateProblem::workSteadyStateProblem(Solver *solver, Model *model,
     cpu_time_ = (double)((clock() - starttime) * 1000) / CLOCKS_PER_SEC;
 
     /* Check whether state sensis still need to be computed */
-    if (getSensitivityFlag(model, solver, it, SteadyStateContext::newtonSensi))
+    if (getSensitivityFlag(model, solver, it, SteadyStateContext::newtonSensi) && numsteps_[0] > 0)
     {
-        if (numsteps_[0] > 0) {
-            try {
-                /* this might still fail, if the Jacobian is singular and
-                 simulation did not find a steady state */
-                newtonSolver->computeNewtonSensis(sx_);
-            } catch (NewtonFailure const &) {
-                /* No steady state could be inferred. Store simulation state */
-                storeSimulationState(model, solver->getSensitivityOrder() >=
-                                     SensitivityOrder::first);
-                throw AmiException("Steady state sensitivity computation failed due "
-                                   "to unsuccessful factorization of RHS Jacobian");
-            }
-        } else {
-            model->initialize(x_, dx_, sx_, sdx_, true);
+        try {
+            /* this might still fail, if the Jacobian is singular and
+             simulation did not find a steady state */
+            newtonSolver->computeNewtonSensis(sx_);
+        } catch (NewtonFailure const &) {
+            /* No steady state could be inferred. Store simulation state */
+            storeSimulationState(model, solver->getSensitivityOrder() >=
+                                 SensitivityOrder::first);
+            throw AmiException("Steady state sensitivity computation failed due "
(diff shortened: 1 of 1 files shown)


src/steadystateproblem.cpp
```diff
@@ -68,17 +68,22 @@ void SteadystateProblem::workSteadyStateProblem(Solver *solver, Model *model,
...
+                 simulation did not find a steady state */
+                newtonSolver->computeNewtonSensis(sx_);
+            } catch (NewtonFailure const &) {
+                /* No steady state could be inferred. Store simulation state */
+                storeSimulationState(model, solver->getSensitivityOrder() >=
+                                     SensitivityOrder::first);
+                throw AmiException("Steady state sensitivity computation failed due "
+                                   "to unsuccessful factorization of RHS Jacobian");
+            }
+        } else {
```
FFroehlich, line 83: Is this really necessary? To me it looks like the sensi are already initialized above (and that code also accounts for whether we are doing pre/post-equilibration).

FFroehlich (approved):

src/steadystateproblem.cpp
paulstapor: I just tried in my first commit what happens if we just skip the computeNewtonSensis().
There was an error for the Raimundez model, i.e. a model with preequilibration. In that case, no solver object had been created at that point and it ran into a core dump
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
Aborted (core dumped)
```
So I guess the problem was a sx0 wasn't written to sx_. This is, why I added this...
I'm happy about any better idea

FFroehlich: I mean the first code block will be executed, no matter what you do in the changes. 

In your first code attempt, you modified the flag whether sensitivities are necessary or not. What I am proposing here is:

```
    if (getSensitivityFlag(model, solver, it, SteadyStateContext::newtonSensi) & numsteps_[0] > 0)
    {
            try {
                /* this might still fail, if the Jacobian is singular and
                 simulation did not find a steady state */
                newtonSolver->computeNewtonSensis(sx_);
            } catch (NewtonFailure const &) {
                /* No steady state could be inferred. Store simulation state */
                storeSimulationState(model, solver->getSensitivityOrder() >=
                                     SensitivityOrder::first);
                throw AmiException("Steady state sensitivity computation failed due "
                                   "to unsuccessful factorization of RHS Jacobian");
            }
    }
```

FFroehlich: I think the failures in the tests are due to reinitialization with the initial state instead of keeping the simulated state.

paulstapor: You're right, the initialization has been done anyway... I think what you propose is logically equivalent to changing the flag, but let's give it a try...

FFroehlich: > You're right, the initialization has been done anyway... I think what you propose is logically equivalent to changing the flag, but let's give it a try...

```
storeSimulationState(model, getSensitivityFlag(model, solver, it, SteadyStateContext::sensiStorage));
```

paulstapor: Thanks for pointing that one out!
The whole logic when which sensitivity is need in that code is frustrating sometimes...

Issue #1156, opened 2020-07-16
Enable adjoint preequilibration for models with solver state reinitialization


paulstapor:
Related to #1146
At the moment, adjoint preequilibration can cope with most cases which are around, except one important situation: If solver states (i.e., non-constant species) are reinitialized, then the current implementation will fail.
Fixing this requires most likely passing a vector from model creation to the model, or, alternatively, creating a function (f)xB0_fixedParameters with according .cpp and .h files.

PR #1468, opened 2021-03-18
Sx0 in adjoint preequilibration

paulstapor:
If adjoint preequllibraiton was not solved algebraically but via integration, there may be a contribution from sx0 * xB, as xB may be != 0 for singular Jacobians... So we may get a non-trivial contribution here. This PR tries to fix that...

### DIFF
--- a/include/amici/steadystateproblem.h
+++ b/include/amici/steadystateproblem.h
@@ -248,7 +248,6 @@ class SteadystateProblem {
         return x_;
     };
 
-
     /**
      * @brief Returns state sensitivity at steadystate
      * @return sx
@@ -322,6 +321,12 @@ class SteadystateProblem {
     void getAdjointUpdates(Model &model, const ExpData &edata);
 
     /**
+     * @brief Accessor for xB
+     * @return xB
+     */
+    AmiVector const& getAdjointState() const { return xB_; }
+
+    /**
      * @brief Accessor for xQB
      * @return xQB
      */
@@ -372,7 +377,7 @@ class SteadystateProblem {
     AmiVector xQ_;
     /** quadrature state vector */
     AmiVector xQB_;
-    /** quadrature state vector */
+    /** time-derivative of quadrature state vector */
     AmiVector xQBdot_;
 
     /** maximum number of steps for Newton solver for allocating numlinsteps */
--- a/include/amici/steadystateproblem.h
+++ b/include/amici/steadystateproblem.h
@@ -321,8 +321,8 @@ class SteadystateProblem {
     void getAdjointUpdates(Model &model, const ExpData &edata);
 
     /**
-     * @brief Accessor for xB
-     * @return xB
+     * @brief Return the adjoint state
+     * @return xB adjoint state
      */
     AmiVector const& getAdjointState() const { return xB_; }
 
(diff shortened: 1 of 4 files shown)


include/amici/steadystateproblem.h
```diff
@@ -321,6 +320,12 @@ class SteadystateProblem {
      */
     void getAdjointUpdates(Model &model, const ExpData &edata);
 
+    /**
+     * @brief Accessor for xB
+     * @return xB
```
dweindl, line 325:
```suggestion
     * @brief Return the adjoint state
     * @return xB adjoint state
```

src/rdata.cpp
```diff
@@ -444,6 +445,21 @@ void ReturnData::handleSx0Backward(const Model &model,
     auto xQBpreeq = preeq.getAdjointQuadrature();
     for (int ip = 0; ip < model.nplist(); ++ip)
         xQB[ip] += xQBpreeq[ip];
+
+    /* Prefer explicit type declaration here, since this MUST be a reference
+     * (for memory reasons), rather than passing this thing by value */
+    const AmiVectorArray &sx0preeq = preeq.getStateSensitivity();
+    auto xBpreeq = preeq.getAdjointState();
```
dweindl, line 452:
```suggestion
    const auto& xBpreeq = preeq.getAdjointState();
```
(or is there a reason for copying?)

src/rdata.cpp
```diff
@@ -444,6 +445,21 @@ void ReturnData::handleSx0Backward(const Model &model,
     auto xQBpreeq = preeq.getAdjointQuadrature();
     for (int ip = 0; ip < model.nplist(); ++ip)
         xQB[ip] += xQBpreeq[ip];
+
+    /* Prefer explicit type declaration here, since this MUST be a reference
+     * (for memory reasons), rather than passing this thing by value */
```
dweindl, line 450: Not sure if more helpful or confusing. Just taking the reference here would be clear enough for me.

src/rdata.cpp
```diff
@@ -444,6 +445,21 @@ void ReturnData::handleSx0Backward(const Model &model,
...
+    const AmiVectorArray &sx0preeq = preeq.getStateSensitivity();
+    auto xBpreeq = preeq.getAdjointState();
+
+    /* Add the contribution for sx0 from preequilibration. If backward
+     * preequilibration was done by simulation due to a singular Jacobian,
+     * xB does not need to be 0 and we get a non-zero contribution here. */
```
dweindl, line 456:
```suggestion
     * xB is not necessarily 0 and we get a non-zero contribution here. */
```

dweindl (approved):
Good as far as I can tell. Can we easily add a test cases for that using one of the existing models?

src/rdata.cpp
paulstapor: changed it

paulstapor:
> Good as far as I can tell. Can we easily add a test cases for that using one of the existing models?

Not as far as I know... There's no model which would be easily usable for this. Either I take an existing one an modify it on the fly using libsbml and generate the PEtab files for it... Or I rather add another test case with a PEtab model... The test case we want to cover is based on PEtab, so I'd suggest an integration test with a PEtab model here... What do you think? Too complex?

paulstapor:
Maybe one more thing:
I'd prefer to add a test on this when fixing #1156, as this one also needs to be covered then. Moreover, we will need this fix here in CanPathPro asap. As I anyway want to finish up the steadystate solver issues as soon as I'll find time (in latest 1-2 months), I'd suggest to add a more comprehensive test later, as this will be needed anyways then...

dweindl:
> I'd suggest to add a more comprehensive test later, as this will be needed anyways then...

works for me

PR #1475, opened 2021-03-24
fix failing petab import when mesaurement_df is None

paulstapor:
measurement_df is optional, so we should not require it to be in there...

### DIFF
--- a/python/amici/petab_import.py
+++ b/python/amici/petab_import.py
@@ -463,7 +463,9 @@ def import_model_sbml(
         petab.lint.observable_table_has_nontrivial_noise_formula(
             observable_df
         )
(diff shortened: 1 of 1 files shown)


FFroehlich:
already fixed in https://github.com/PEtab-dev/libpetab-python/pull/56

paulstapor:
Also fine for me, as long as parPE CI works again...

dweindl (approved):

dweindl:
> already fixed in [PEtab-dev/libpetab-python#56](https://github.com/PEtab-dev/libpetab-python/pull/56)

For me still fine / cleaner to handle it this way.

PR #1487, opened 2021-04-09
construct condition specific plist in parameter mapping

FFroehlich:


python/amici/parameter_mapping.py
```diff
@@ -236,12 +236,22 @@ def _get_par(model_par, value):
     scales = [petab_to_amici_scale(scale_map_sim_var[par_id])
               for par_id in amici_model.getParameterIds()]
 
+    # plist
+    plist = [
+        amici_model.getParameterIds().index(name)
```
yannikschaelte, line 241: looks like sth one would want to hash if npar >> 100, but probably fine.

yannikschaelte (approved):

PR #1489, opened 2021-04-12
Add residuals for quadratic nllhs with parameter dependent sigma

FFroehlich:
default `min_sigma_ = 50` corresponds to d2d default. This enables optimization of respective objective functions using least squares optimizers, but also allows changing the FIM accordingly, which is then used in trust region optimizers.

### DIFF
--- a/include/amici/model.h
+++ b/include/amici/model.h
@@ -648,7 +648,7 @@ class Model : public AbstractModel, public ModelDimensions {
      * activated, this lower boundary must ensure that log(sigma) + min_sigma > 0.
      * @param min_sigma lower boundary
      */
-    void setMinimumSigmaResiduals(float min_sigma) {
+    void setMinimumSigmaResiduals(double min_sigma) {
         min_sigma_ = min_sigma;
     }
     
@@ -662,7 +662,8 @@ class Model : public AbstractModel, public ModelDimensions {
     
     /**
      * @brief Specifies whether residuals should be added to account for parameter dependent sigma.
-     * If set to true, additional residuals of the form `sqrt{log(\sigma) + C}` will be added. This enables
+     *
+     * If set to true, additional residuals of the form \f[ \sqrt{\log(\sigma) + C} \f] will be added. This enables
      * least-squares optimization for variables with Gaussian noise assumption and parameter dependent
      * standard deviation sigma. The constant C can be set via :meth:`setMinimumSigmaResiduals`.
      * @param sigma_res if true, additional residuals are added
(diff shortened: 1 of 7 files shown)


src/rdata.cpp
```diff
@@ -833,6 +854,14 @@ void ReturnData::fsres(const int it, Model &model, const ExpData &edata) {
             sres.at(idx) = amici::fsres(y_it.at(iy), sy_it.at(iy + ny * ip),
                                         observedData[iy], sigmay_it.at(iy),
                                         ssigmay_it.at(iy + ny * ip));
+            if (sigma_res) {
+                int idx_res =
+                    (iy + it * edata.nytrue() + edata.nytrue() * edata.nt()) *
```
dweindl, line 859: cache result of `edata.n*()` outside nested loop? Could also avoid some repeated `.at()`s, but not sure how relevant.

dweindl (approved):

yannikschaelte (approved):
the formulas are correct (one might want to document how it works, i.e. introduction of additional residuals), aside nothing except what daniel pointed out

include/amici/model.h
```diff
@@ -1704,6 +1740,12 @@ class Model : public AbstractModel, public ModelDimensions {
      * checked for finiteness
      */
     bool always_check_finite_ {false};
+    
+    /** indicates whether sigma residuals are to be added for every  */
+    bool sigma_res_ {false};
+    
+    /** offset to ensure positivity of sigma residuals, only has an effect when `sigma_res_` is `true`  */
+    realtype min_sigma_ {50.0};
```
paulstapor, line 1748: Just as a remark:
The higher the offset, the weaker the contribution of this term to the gradient will be, right? The square root function flattens out for higher values... So, I my gut feeling tells me that this offset term influences the parameter estimation process somehow. Therefore, I would suggest a lower value than 50, which is however still high enough to not cause problems for typical applications. I think already 10 should do the job here...

yannikschaelte: I agree that the offset should have an influence. No idea about practically typical values (ok, log(2*pi*1e-8^2)=-35, one could look at typical bounds in the benchmark collection).

src/rdata.cpp
FFroehlich: as those are not called during simulation, computational cost is pretty negligible here.

include/amici/model.h
FFroehlich: Note that we do not account for the `2pi` factor here (we could). The gradient itself is not influenced by C, but `sres` is and the FIM is. Probably the smartest way to go about this is set some value based on bounds for sigmas in pypesto.

PR #1490, opened 2021-04-13
Fix serialization

dweindl:
Previously the archive was truncated under certain circumstances. The archive must be destroyed *before* returning.

### DIFF
--- a/include/amici/serialization.h
+++ b/include/amici/serialization.h
@@ -299,12 +299,13 @@ T deserializeFromChar(const char *buffer, int size) {
     T data;
 
     try {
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):

FFroehlich:
> Previously the archive was truncated under certain circumstances. The archive must be destroyed _before_ returning.

Sounds like this took a while to figure out what's going on.

dweindl:
> Sounds like this took a while to figure out what's going on.

Very enjoyable experience, yes :hankey:

PR #1493, opened 2021-04-13
fix serialization and == operator

FFroehlich:


dweindl (approved):
Any idea how we can avoid forgetting about updating serialization and HDF5 code? (would you like to include that latter here as well?)

yannikschaelte:
> Any idea how we can avoid forgetting about updating serialization and HDF5 code? (would you like to include that latter here as well?)

Forgot how the tests work, but one could maintain a negative list of attributes _not_ to compare, and check everything by default.

dweindl:
> > Any idea how we can avoid forgetting about updating serialization and HDF5 code? (would you like to include that latter here as well?)
> 
> Forgot how the tests work, but one could maintain a negative list of attributes _not_ to compare, and check everything by default.

The problem is that the serialization part is C++-only where we don't have Python-like introspection. So this would require exposing serialization to Python, which will require boost. We don't want to have boost as general requirement, but it would be possible to use it for testing only.

PR #1494, opened 2021-04-13
Add basic check for serialization functions being up-to-date

dweindl:
Will yield false positives if types are changed. Will not catch size-neutral changes. Will require manually updating the sizes.
Not aware of any not significantly more complicated alternatives...


FFroehlich (approved):

FFroehlich (approved):
:+1: how about directly printing actual size such that it's easier to update these values when necessary?

dweindl:
> +1 how about directly printing actual size such that it's easier to update these values when necessary?

Would be more convenient, yes. Only that it's somewhat ugly [doing that at compile time ](https://stackoverflow.com/questions/20979565/how-can-i-print-the-result-of-sizeof-at-compile-time-in-c).

Also need to check how to best deal with different sizes on different platforms. Maybe enable only for a specific setting used on GHA. Otherwise we need something like `3 * sizeof(int) + 4 * sizeof(double) + ....`.

dweindl:
I think I will rather not merge this. Seems very inconvenient. Would rather (optionally) expose serialization and go for testing via Python, similarly to the current `test_copy_constructors`.

PR #1496, opened 2021-04-15
Doc: Update OpenBLAS Windows installation instructions

willov:
The current version of compileBLAS.cmd / installOpenBLAS.ps1 does not use a "v" in the top-level OpenBLAS-0.3.12 folder. 

Backward slashes got removed during installation of amici. Switching to forward slashes worked.


dweindl (approved):
Thanks for correcting that, @willov

PR #1503, opened 2021-04-26
avoid w in root and stau headers (refactor)

### LINKED ISSUE #1501: Error during sensitivity computation
I am getting errors when computing model sensitivities using `runAmiciSimulation` with a model specified in PETab. Specifically, the error is

`[Warning] AMICI:Inf: AMICI encountered an Inf value at index 0/8 in sxdot!`

A working example of the problem including the PETab files is here: [reproduce_issue.zip](https://github.com/AMICI-dev/AMICI/files/6372359/reproduce_issue.zip). `reproduce_issue.py` fails after printing the above error 37 times. I am using the Python interface and Amici version 0.11.14 on Ubuntu 20.04. 

There are Dirac delta functions appearing in `Oliveira_Nat2021_interior_dwdp.cpp`, but I am not certain they are the problem. Any support regarding the cause of this error and how to resolve it is greatly appreciated. Thanks in advance!

dilpath:
closes #1501 

`from_sbml` is passed through as there might be an issue for PySB symbols (haven't used PySB, so not really sure -- was just looking at `_generate_symbol`).

Could alternatively make `from_sbml` a class attribute, and/or add e.g. a `get_symbol_id` class method.

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -1626,7 +1626,7 @@ class ODEModel:
         """
         # We need to process events and Heaviside functions in the ODE Model,
         # before adding it to ODEExporter
-        self.parse_events(from_sbml=from_sbml)
+        self.parse_events()
 
         for var in self._variable_prototype:
             if var not in self._syms:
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):
> closes #1501
> 
> `from_sbml` is passed through as there might be an issue for PySB symbols (haven't used PySB, so not really sure -- was just looking at `_generate_symbol`).
> 
> Could alternatively make `from_sbml` a class attribute, and/or add e.g. a `get_symbol_id` class method.

PySB doesn't support events, so I don't think any extra handling is necessary here.

One thing I am wondering about is whether we may end up having Heaviside functions in the root function? Does it make sense to add a

```
for root in roots:
    root.set_val(self._process_heavisides(root.get_val(), roots))
```
?

dilpath:
> > closes #1501
> > `from_sbml` is passed through as there might be an issue for PySB symbols (haven't used PySB, so not really sure -- was just looking at `_generate_symbol`).
> > Could alternatively make `from_sbml` a class attribute, and/or add e.g. a `get_symbol_id` class method.
> 
> PySB doesn't support events, so I don't think any extra handling is necessary here.
> 
> One thing I am wondering about is whether we may end up having Heaviside functions in the root function? Does it make sense to add a
> 
> ```
> for root in roots:
>     root.set_val(self._process_heavisides(root.get_val(), roots))
> ```
> 
> ?

I guess this should be equivalent to doing a toposort on `w` to find the order by which expressions should have their roots parsed (and perhaps also parsing expressions before states in `parse_events` [1]?)? Can simply add the suggestion, probably safest at the cost of some computation, any preference?

[1] https://github.com/AMICI-dev/AMICI/blob/2b4158c671d20a4a3646f3754378c9b5e6ec50d3/python/amici/ode_export.py#L1637-L1658

PR #1515, opened 2021-06-17
Python splines

paulstapor:
This one is to replace #1245. However, it seems like in the long time of divergence, there were changes to functions from sbml_import, that may be not captured, as these functions have been moved out of sbml_import.py to sbml_utils.py on the python_splines branch... It looks like automatic merge of these files did not do completely the right thing... So, we have to be careful about functions that have been moved to newly created files, it seems...


src/symbolic_functions.cpp
```diff
@@ -25,82 +25,33 @@
 
 namespace amici {
 
-/**
```
dweindl, line 28: What's the reason for removing all those comments?

src/splinefunctions.cpp
```diff
@@ -0,0 +1,779 @@
...
+                break;
+
+            default:
+                /* We don't need specific coefficients in the cases of:
+                 * noExtrapolation, polynomial, periodic*/
+                break;
+        }
+        /* Write them to the vector */
+        coefficients_extrapolate_sensi[4 * ip + 2] = sp_end -
+            sm_end * nodes_[n_nodes() - 1];
```
dweindl, line 445: @lcontento : If the `default` branch is taken, `sm_end` is undefined here. What is supposed to happen?

src/splinefunctions.cpp
```diff
@@ -0,0 +1,779 @@
...
+                break;
+
+            default:
+                /* We don't need specific coefficients in the cases of:
+                 * noExtrapolation, polynomial, periodic*/
+                break;
+        }
+        /* Write them to the vector */
+        coefficients_extrapolate_sensi[4 * ip] = sp0 - sm0 * nodes_[0];
+        coefficients_extrapolate_sensi[4 * ip + 1] = sm0;
```
dweindl, line 391: @lcontento : If the `default` branch is taken, `sm0` is undefined here. What is supposed to happen?

src/symbolic_functions.cpp
lcontento: They seem to have been moved to the corresponding header file. Which would make more sense to me, but I am not sure what the conventions are for C++ in-source documentation.

dweindl: Ah, sorry, missed that. Makes sense.

src/splinefunctions.cpp
lcontento: Not 100% sure. I have some uncommitted changes to fix the compilation mistakes and there I just "fixed" it by raising an exception in the default case...

src/splinefunctions.cpp
lcontento: Not 100% sure. I have some uncommitted changes to fix the compilation mistakes and there I just "fixed" it by raising an exception in the default case...

src/splinefunctions.cpp
dweindl: That solves the "undefined" issues. However, from that comment, I wouldn't think that this is what's supposed to happen.

lcontento: Yes, it's just a placeholder at the moment to remind the reader that the implementation is missing. I will need to check if that bit is necessary for the basic functionality we are aiming for in this first PR.

dweindl:
ping @lcontento

python/amici/sbml_import.py
```diff
@@ -929,13 +965,33 @@ def _process_rules(self) -> None:
         """
         Process Rules defined in the SBML model.
         """
+        assert not hasattr(self, 'splines')
+        self.splines = []
```
dweindl, line 969: Should be created in `__init__`

python/amici/sbml_utils.py
```diff
@@ -0,0 +1,722 @@
...
+
+class SbmlException(Exception):
+    "Exception class for SBML-related errors."
+    pass
+
+
+################################################################################
+
+
+def createSbmlModel(modelId: str, level: int = 2, version: int = 5):
```
dweindl, line 43: Please go for snake_case

python/amici/sbml_utils.py
```diff
@@ -0,0 +1,722 @@
+"""
```
dweindl, line 1: there are a number of unresolved references in this file

python/amici/splines.py
```diff
@@ -0,0 +1,1584 @@
...
+
+    def check_if_valid(self, importer: SbmlImporter):
+        """
+        Check if the spline described by this object can be correctly
+        be implemented by AMICI. E.g., check whether the formulas
+        for spline grid points, values, ... contain species symbols.
+        """
+        # TODO this is very much a draft
+
+        fixed_parameters = importer.symbols['fixed_parameter']['identifier']
```
dweindl, line 505: Should not be indexed with literals, but with `SymbolId`. See also other cases.

src/model.cpp
```diff
@@ -1928,10 +1967,38 @@ void Model::fdJrzdsigma(const int ie, const int nroots, const realtype t,
...
+    realtype *sspl_data = derived_state_.sspl_.data();
+    for (int ip = 0; ip < nplist(); ip++) {
+        for (int ispl = 0; ispl < nspl; ispl++) {
+            if (splines_[ispl].get_logarithmic_parametrization()) {
+                sspl_data[ispl + nspl * ip] =
+                    state_.spl_[ispl] * splines_[ispl].getSensitivity(t, ip);
+            } else {
+                sspl_data[ispl + nspl * ip] =
+                    splines_[ispl].getSensitivity(t, ip);
+            }
```
dweindl, line 1991: Can/should that `if` not go into `getSensitivity`?

src/model.cpp
```diff
@@ -1928,10 +1967,38 @@ void Model::fdJrzdsigma(const int ie, const int nroots, const realtype t,
     }
 }
 
+void Model::fspl(const realtype t) {
+    for (int ispl = 0; ispl < nspl; ispl++) {
+        if (splines_[ispl].get_logarithmic_parametrization()) {
+            state_.spl_[ispl] = std::exp(splines_[ispl].getValue(t));
+        } else {
+            state_.spl_[ispl] = splines_[ispl].getValue(t);
+        }
```
dweindl, line 1976: Can/should that `if` not go into `getValue`?

src/splinefunctions.cpp
```diff
@@ -0,0 +1,883 @@
...
+        i_node = 0;
+        while (nodes_[i_node + 1] < t) {
+            i_node++;
+        }
+        len = nodes_[i_node + 1] - nodes_[i_node];
+    }
+
+    /* Evaluate the interpolation polynomial */
+    return evaluatePolynomial((t - nodes_[i_node]) / len,
+                              &(coefficients[i_node * 4]));
```
dweindl, line 784: @lcontento It seems there is some out-of-bounds read in `coefficients` here. Please check

lcontento: I think there was an edge case (t = final point) which was not taken care of. But even correcting for that I still get segfaults.

src/model.cpp
dweindl: @lcontento ?

src/model.cpp
dweindl: @lcontento ?

lcontento: I think it can and I think it should, not sure if there is any reason why Paul put it outside.
I can try putting inside `getValue` and see if anything breaks.

src/model.cpp
lcontento: If we put the `if` inside then we would have to recompute the spline value for each parameter, while in this way we can recycle it.

lcontento: As mentioned in the other comment, I honestly think the `if` belongs logically inside the spline methods, but with the current method signature I would not move it. If we had a method that computes both spline value and spline sensitivities at the same time it would be better though in my opinion.

lcontento: see comment regarding same problem in sensitivity computation

python/amici/sbml_utils.py
lcontento: I used camel case to make it stylistically similar to the `libsbml` API, but would not mind changing

src/model.cpp
dweindl: If I understand it correctly, this could be addressed by having `get_{value,sensitivity}` and `get_unscaled_{value,sensitivity}`? Here, we'd use the former, and where required inside `splinesfunctions.cpp`, we'd use the latter?

python/amici/ode_export.py
```diff
@@ -1062,6 +1091,28 @@ def import_from_sbml_importer(
             whether to compute conservation laws
         """
 
+        # add splines as expressions to the model
+        # saved for later substituting into the fluxes
+        spline_subs = {}
+
+        # HACK remove real=True assumptions from spline symbols
+        #   must be done as the first thing because it modifies the SbmlImporter
```
dweindl, line 1099: Why is the `real` assumption a problem?

python/amici/sbml_utils.py
```diff
@@ -0,0 +1,600 @@
...
+
+    return expr
+
+
+def _check_unsupported_functions(
+        sym: sp.Basic,
+        expression_type: str,
+        full_sym: sp.Basic = None
+        ):
+    """
```
dweindl, line 528: This looks like a copy of `import_utils._check_unsupported_functions`. Please check which one is newer.

src/model.cpp
```diff
@@ -240,6 +243,41 @@ void Model::initializeStates(AmiVector &x) {
...
+
+    std::vector<realtype> dspline_valuesdp (allnodes * nplist(), 0.0);
+    std::vector<realtype> dspline_slopesdp (allnodes * nplist(), 0.0);
+    fdspline_valuesdp(dspline_valuesdp.data(), state_.unscaledParameters.data(),
+                      state_.fixedParameters.data());
+    fdspline_slopesdp(dspline_slopesdp.data(), state_.unscaledParameters.data(),
+                      state_.fixedParameters.data());
+
+    // QUESTION is it possible for plist != range(nplist) ?
+    //          would that break computeCoefficientsSensi or not?
```
dweindl, line 271: Yes, `plist` will often not be equal to `range(nplist)`. Seems like a good test case...

python/amici/ode_export.py
lcontento: I don't know.

python/amici/sbml_import.py
dweindl: done

python/amici/sbml_utils.py
dweindl: i see your point, but let's stick to pep8

python/amici/sbml_utils.py
dweindl: resolved

src/splinefunctions.cpp
dweindl: Seems resolved.

python/amici/splines.py
```diff
@@ -0,0 +1,1636 @@
...
+            )
+            if retcode != libsbml.LIBSBML_OPERATION_SUCCESS:
+                raise SbmlAnnotationError('Could not set SBML annotation!')
+            add_assignment_rule(model, parameter_id, formula)
+
+    # def _replace_sbml_time_with_amici_time(self) -> None:
+    #     self._replace_in_all_expressions(
+    #         sbml_time_symbol, amici_time_symbol
+    #     )
+
```
dweindl, line 1036:
```suggestion
```

@lcontento : can be removed?

tests/cpp/unittests/testSplines.cpp
```diff
@@ -0,0 +1,498 @@
...
+         0.0,  0.0,  0.0,
+        -6.0,  1.0,  3.0
+    };
+    std::vector<double> dslopesdp = {
+         0.0,  0.0,      18.75,
+         0.0,  1.0,       3.0,
+         4.0, -0.540302,  0.0,
+         0.0,  0.0,       0.0,
+    };
+    std::vector<double>(spline.n_nodes() * n_params);
```
dweindl, line 445: This is just constructing a vector and discarding it immediately. Supposed to be used somewhere?

lcontento: Just a leftover from copying and editing the previous test. Thanks for catching it!

python/sdist/amici/sbml_import.py
```diff
@@ -851,6 +863,25 @@ def add_d_dt(
                 'dt': d_dt,
             }
 
+    @log_execution_time('processing SBML annotations', logger)
+    def _process_annotations(self) -> None:
+        # Remove all parameters (and corresponding rules)
+        # for which amici:discard is set
```
dweindl, line 869:
```suggestion
        """
        Remove all parameters (and corresponding rules)
        for which amici:discard is set
        """
```

python/tests/test_splines.py
```diff
@@ -0,0 +1,83 @@
...
+import numpy as np
+from splines_utils import (
+    example_spline_1,
+    example_spline_2,
+    example_spline_3,
+    check_splines_full,
+)
+
+
+def test_splines(**kwargs):
```
dweindl, line 11: Rather generic name. Please add docstring with rationale for this test.

python/tests/test_splines.py
```diff
@@ -0,0 +1,83 @@
...
+    tols[2]['x_rtol']    = max(5e-10, tols[2].get('x_rtol', -np.inf))
+    tols[2]['x_atol']    = max(1e-8,  tols[2].get('x_atol', -np.inf))
+    tols[2]['llh_rtol']  = max(5e-14, tols[2].get('llh_rtol', -np.inf))
+    tols[2]['sllh_atol'] = max(5e-5,  tols[2].get('sllh_atol', -np.inf))
+
+    if os.name == 'nt':
+        tols[2]['sllh_atol'] = max(5e-4,  tols[2]['sllh_atol'])
+    
+    # Load precomputed results
+    precomputed_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "test_splines_precomputed.npz")
```
dweindl, line 80: Please document how this file was generated / how it can be recreated, when does it have to be recreated.

dweindl (commented):
to be continued

python/sdist/amici/de_export.py
```diff
@@ -941,6 +970,26 @@ def import_from_sbml_importer(
...
+        # add splines as expressions to the model
+        # saved for later substituting into the fluxes
+        spline_subs = {}
+
+        for ispl, spl in enumerate(si.splines):
+            # HACK remove real=True assumptions from spline symbols
+            #   must be done as the first thing because it modifies the SbmlImporter
+            old_sbml_id = spl.sbml_id
+            new_sbml_id = sp.Symbol(spl.sbml_id.name)
+            si._replace_in_all_expressions(old_sbml_id, new_sbml_id)
```
dweindl, line 982: Why is that an issue? Wouldn't that be handled better by adding `real=True` elsewhere?

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+
+
+class SbmlAnnotationError(SBMLException):
+    pass
+
+
+###############################################################################
+
+
+def create_sbml_model(model_id: str, level: int = 2, version: int = 5) \
```
dweindl, line 54: Move to tests/spline_utils.py, as only used for tests.

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+    if name is True:
+        name = species_id
+
+    # Check whether a species with the same ID already exists
+    # TODO the resulting SBML may still be invalid
+    #      if other types of objects (e.g., parameter) have the same ID
+    if model.getSpecies(species_id):
+        raise SbmlDuplicateComponentIdError(
+            f'A species with ID {species_id} has already been defined'
+        )
```
dweindl, line 157:
```suggestion
    # Check whether an element with the same ID already exists
    if model.getElementBySId(species_id):
        raise SbmlDuplicateComponentIdError(
            f'An element with ID {species_id} has already been defined.'
        )
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+        compartments = model.getListOfCompartments()
+        if len(compartments) != 1:
+            raise ValueError(
+                'Compartment auto-selection is possible '
+                'only if there is one and only one compartment.'
+            )
+        compartment_id = compartments[0].getId()
+    elif not model.getCompartment(compartment_id):
+        raise SbmlMissingComponentIdError(
+            f'No compartment with ID {compartment_id}'
```
dweindl, line 169:
```suggestion
            f'No compartment with ID {compartment_id}.'
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+        )
+
+    sp = model.createSpecies()
+    if sp.setIdAttribute(species_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
+        raise SbmlInvalidIdSyntax(f'{species_id} is not a valid SBML ID')
```
dweindl, line 174:
```suggestion
        raise SbmlInvalidIdSyntax(f'{species_id} is not a valid SBML ID.')
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+    if name is True:
+        name = parameter_id
+
+    # Check whether a parameter with the same ID already exists
+    # TODO the resulting SBML may still be invalid
+    #      if other types of objects (e.g., species) have the same ID
+    if model.getParameter(parameter_id):
+        raise SbmlDuplicateComponentIdError(
+            f'A parameter with ID {parameter_id} has already been defined'
+        )
```
dweindl, line 227:
```suggestion
    # Check whether an element with the same ID already exists
    if model.getElementBySId(parameter_id):
        raise SbmlDuplicateComponentIdError(
            f'An element with ID {parameter_id} has already been defined.'
        )
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+
+    par = model.createParameter()
+    if par.setIdAttribute(parameter_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
+        raise SbmlInvalidIdSyntax(f'{parameter_id} is not a valid SBML ID')
```
dweindl, line 231:
```suggestion
        raise SbmlInvalidIdSyntax(f'{parameter_id} is not a valid SBML ID.')
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+    # TODO the resulting SBML may still be invalid
+    #      if other types of objects (e.g., species) have the same ID
+    if model.getRuleByVariable(variable_id):
+        raise SbmlDuplicateComponentIdError(
+            f'A rule for parameter {variable_id} has already been defined.'
+        )
+    if model.getRule(rule_id):
+        raise SbmlDuplicateComponentIdError(
+            f'A rule with SBML ID {rule_id} has already been defined.'
+        )
```
dweindl, line 282:
```suggestion
    if model.getRuleByVariable(variable_id):
        raise SbmlDuplicateComponentIdError(
            f'A rule for parameter {variable_id} has already been defined.'
        )
    if model.getElementBySId(rule_id):
        raise SbmlDuplicateComponentIdError(
            f'An element with SBML ID {rule_id} has already been defined.'
        )
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+
+    rule = model.createAssignmentRule()
+    if rule.setVariable(variable_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
+        raise SbmlInvalidIdSyntax(f'{variable_id} is not a valid SBML ID')
+    if rule.setIdAttribute(rule_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
+        raise SbmlInvalidIdSyntax(f'{rule_id} is not a valid SBML ID')
```
dweindl, line 288:
```suggestion
        raise SbmlInvalidIdSyntax(f'{variable_id} is not a valid SBML ID.')
    if rule.setIdAttribute(rule_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
        raise SbmlInvalidIdSyntax(f'{rule_id} is not a valid SBML ID.')
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+    # TODO the resulting SBML may still be invalid
+    #      if other types of objects (e.g., species) have the same ID
+    if model.getRuleByVariable(variable_id):
+        raise SbmlDuplicateComponentIdError(
+            f'A rule for parameter {variable_id} has already been defined.'
+        )
+    if model.getRule(rule_id):
+        raise SbmlDuplicateComponentIdError(
+            f'A rule with SBML ID {rule_id} has already been defined.'
+        )
```
dweindl, line 333:
```suggestion
    if model.getRuleByVariable(variable_id):
        raise SbmlDuplicateComponentIdError(
            f'A rule for parameter {variable_id} has already been defined.'
        )
    if model.getElementBySId(rule_id):
        raise SbmlDuplicateComponentIdError(
            f'An element with SBML ID {rule_id} has already been defined.'
        )
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+
+    rule = model.createRateRule()
+    if rule.setVariable(variable_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
+        raise SbmlInvalidIdSyntax(f'{variable_id} is not a valid SBML ID')
+    if rule.setIdAttribute(rule_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
+        raise SbmlInvalidIdSyntax(f'{rule_id} is not a valid SBML ID')
```
dweindl, line 339:
```suggestion
        raise SbmlInvalidIdSyntax(f'{variable_id} is not a valid SBML ID.')
    if rule.setIdAttribute(rule_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
        raise SbmlInvalidIdSyntax(f'{rule_id} is not a valid SBML ID.')
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+    reversible: bool = False,
+) -> libsbml.Reaction:
+    species_id = str(species_id)
+    if reaction_id is None:
+        reaction_id = f'inflow_of_{species_id}'
+
+    if model.getReaction(reaction_id):
+        raise SbmlDuplicateComponentIdError(
+            f'A reaction with SBML ID {reaction_id} has already been defined.'
+        )
```
dweindl, line 360:
```suggestion
    if model.getElementBySId(reaction_id):
        raise SbmlDuplicateComponentIdError(
            f'An element with SBML ID {reaction_id} has already been defined.'
        )
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+
+    reaction = model.createReaction()
+    if reaction.setId(reaction_id) != libsbml.LIBSBML_OPERATION_SUCCESS:
+        raise SbmlInvalidIdSyntax(f'{reaction_id} is not a valid SBML ID')
```
dweindl, line 364:
```suggestion
        raise SbmlInvalidIdSyntax(f'{reaction_id} is not a valid SBML ID.')
```

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+    :param x:
+        Expression to get the units of.
+    :return:
+        A string if the units could be determined, otherwise `None`.
+    """
+    # TODO can the SBML unit inference machinery be used?
+    x = sp.sympify(x)
+    if not x.is_Symbol:
+        return None
+    if x.name == 'time':
```
dweindl, line 395: amici_time_symbol / sbml_time_symbol

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+    if units == '':
+        return None
+    return units
+
+
+###############################################################################
+# SymPy to SBML MathML/AST conversion
+
+
+def pretty_xml(ugly_xml: str) -> str:
```
dweindl, line 412: missing docstring

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,558 @@
...
+        str with mathematical expression
+    :param math_str:
+        parsed math_str
+    """
+    if not isinstance(math_str, str):
+        return math_str
+
+    if ' xor(' in math_str or ' Xor(' in math_str:
+        raise SBMLException('Xor is currently not supported as logical '
+                            'operation.')
```
dweindl, line 556: There is `sympy.Xor`, but I think this function was just moved here. So okay to leave.

dweindl (commented):
tbc

python/sdist/amici/sbml_utils.py
lcontento: Changed to `x.name == sbml_time_symbol.name`. Not sure if equality of symbols is advisable, I am afraid of sympy assumptions complicating things

python/sdist/amici/de_export.py
lcontento: I have no idea, this was something that was put in by Paul

python/tests/splines_utils.py
```diff
@@ -0,0 +1,851 @@
...
+        # Create PEtab problem
+        path, initial_values, T = create_petab_problem(
+            splines, params_true, initial_values,
+            sigma=0.0, folder=folder, **kwargs
+        )
+        petab_problem = petab.Problem.from_yaml(path)
+
+    if amici_model is None:
+        # Create and compile AMICI model
+        model_id = uuid.uuid1().hex[:5] # to prevent folder/module collisions
```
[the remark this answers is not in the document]
dweindl: `uuid4` should mostly do the job:

```python
In [36]: print(Counter([uuid4().hex[-5:] for _ in range(100)]))
Counter({'bc9a1': 1, 'fa18d': 1, 'a290b': 1, '054c5': 1, '90c8a': 1, 'b2468': 1, 'd49a8': 1, 'f0123': 1, '18edf': 1, 'af9b5': 1, '665c2': 1, '0af0f': 1, '8e478': 1, 'a431e': 1, '19923': 1, '28f45': 1, '2cb20': 1, '3520f': 1, 'cb156': 1, 'ad6ef': 1, '4db76': 1, 'a8ca1': 1, '19618': 1, 'b5646': 1, '20ca5': 1, 'b43ca': 1, 'a5bbf': 1, 'a30e0': 1, 'f8c0d': 1, '09f22': 1, '174dc': 1, '4a3f0': 1, 'b0d6a': 1, 'fa997': 1, '6199a': 1, 'c366c': 1, '0b33a': 1, 'd87ce': 1, 'e7510': 1, '5d230': 1, 'bdec8': 1, 'f0be3': 1, 'a18b7': 1, '65124': 1, 'd6abb': 1, 'ae11c': 1, 'c1830': 1, '1288c': 1, 'c5170': 1, '19b52': 1, 'a83c4': 1, '0030d': 1, '18605': 1, '648f0': 1, '03ef8': 1, '3486a': 1, '2ba1d': 1, '78f06': 1, '3b476': 1, 'bc83a': 1, '92fde': 1, '07988': 1, '190ef': 1, '13f8f': 1, '058aa': 1, '0c629': 1, '8f3de': 1, '76dfa': 1, '149e5': 1, 'c9e7d': 1, 'cacb4': 1, '61486': 1, '14327': 1, '6df0e': 1, '99dba': 1, 'cc7df': 1, '017e7': 1, '012fe': 1, 'b2c63': 1, 'ba1be': 1, 'dbac7': 1, 'a35e2': 1, '21e7e': 1, '079a6': 1, 'fb859': 1, '69e78': 1, 'e7857': 1, '7457b': 1, 'cb93e': 1, '38451': 1, '93d07': 1, '896f9': 1, '01391': 1, '41344': 1, '2a1c1': 1, '6f04d': 1, 'cb375': 1, '666a5': 1, '88540': 1, '8b80f': 1})
```

documentation/python_modules.rst
```diff
@@ -26,3 +26,5 @@ AMICI Python API
    amici.conserved_quantities_demartino
    amici.conserved_quantities_rref
    amici.numpy
+   amici.sbml_utils
```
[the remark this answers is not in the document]
lcontento: There are some of those that are used in the example notebooks though. But maybe it does not matter.

python/sdist/amici/sbml_import.py
lcontento: In theory that is only one particular thing this function could do. It just happens it is the only thing it does at the moment. I made a more general doc-string.

python/sdist/amici/sbml_utils.py
dweindl: I take that back. Can stay, as the other related functions are also required elsewhere.

python/tests/splines_utils.py
```diff
@@ -0,0 +1,851 @@
...
+    :param amici_model:
+        AMICI model (if already created)
+
+    :param kwargs:
+        passed to `create_petab_problem`
+    """
+    # If no working directory is given, create a temporary one
+    if folder is None:
+        if keep_temporary:
+            folder = TemporaryDirectory().name
```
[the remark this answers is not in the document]
lcontento: Yes, it is not guaranteed to work. But it's only if you have to debug the thing and look at the generated code. And in practice it works. It could be definitely be made to be guaranteed to work, but it would be overcomplicated for what it is actually used for. Unless you have in mind a concise alternative.

python/examples/example_splines/ExampleSplines.ipynb
```diff
@@ -0,0 +1,1220 @@
...
+   },
+   "source": [
+    "### Specifying derivatives, boundary conditions and extrapolation methods"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "When derivatives are not specified in the `CubicHermiteSpline` constructor, they are computed automatically using finite differences and according to the boundary conditions. If their form is known a prior (e.g., they are known constants or functions of parameters), they can be passed explicitly to the spline constructor."
```
dweindl, line 579:
```suggestion
    "When derivatives are not specified in the `CubicHermiteSpline` constructor, they are computed automatically using finite differences and according to the boundary conditions. If their form is known a priori (e.g., they are known constants or functions of parameters), they can be passed explicitly to the spline constructor."
```

python/examples/example_splines_swameye/ExampleSplinesSwameye2003.ipynb
```diff
@@ -0,0 +1,1945 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "c10af447-e4f1-46e1-bf60-910dc67f5d77",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "# Spline implementation of JAK2-STAT5 signaling pathway\n",
+    "In this notebook a practical example of the usage of AMICI spline functionalities is shown.\n",
+    "The model under consideration is the JAK2-STAT5 signaling pathway (Swameye et al., 2003),\n",
```
dweindl:
```suggestion
    "The model under consideration is the JAK2-STAT5 signaling pathway ([Swameye et al., 2003](https://doi.org/10.1073/pnas.0237333100)),\n",
```

python/examples/example_splines_swameye/ExampleSplinesSwameye2003.ipynb
```diff
@@ -0,0 +1,1945 @@
...
+    "in which the dynamics of the system depend on a measured input function (the quantity `pEpoR` in the model).\n",
+    "\n",
+    "Following the approach of (Schelker et al., 2012), a continuous approximation of this input function is estimated together with the other parameters.\n",
```
dweindl:
```suggestion
    "Following the approach of ([Schelker et al., 2012](https://doi.org/10.1093/bioinformatics/bts393)), a continuous approximation of this input function is estimated together with the other parameters.\n",
```

python/examples/example_splines_swameye/ExampleSplinesSwameye2003.ipynb
```diff
@@ -0,0 +1,1945 @@
...
+    "More precisely, the logarithm of the input function will be estimated in order to enforce the positivity constraint.\n",
+    "We remark that due to technical limitations `log(pEpoR)` cannot be directly represented as a spline in AMICI since it is an observed quantity.\n",
+    "Thus the derivative of `log(pEpoR)` will instead be approximated as a spline and `log(pEpoR)` computed by solving the resulting ODE.\n",
+    "\n",
+    "The model of the signaling pathway will be implemented in SBML using AMICI's spline annotations, experimental data integrated using the PEtab format and parameter estimation will be carried out using the pyPESTO library."
```
dweindl:
```suggestion
    "The model of the signaling pathway will be implemented in SBML using AMICI's spline annotations, experimental data integrated using the PEtab format and parameter estimation will be carried out using the [pyPESTO](https://pypesto.readthedocs.io/) library."
```

python/examples/example_splines_swameye/ExampleSplinesSwameye2003.ipynb
```diff
@@ -0,0 +1,1945 @@
...
+   "execution_count": 146,
+   "id": "bf1e3a52-8f72-466e-b63c-2f15fbe59137",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "# Load existing results if available\n",
+    "if os.path.exists(f'{name}.hdf5'):\n",
+    "    pypesto_result = pypesto.store.read_result(f'{name}.hdf5', problem=pypesto_problem)\n",
```
dweindl:
```suggestion
    "if os.path.exists(f'{name}.h5'):\n",
    "    pypesto_result = pypesto.store.read_result(f'{name}.h5', problem=pypesto_problem)\n",
```

python/examples/example_splines_swameye/ExampleSplinesSwameye2003.ipynb
```diff
@@ -0,0 +1,1945 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "0592af7c-227a-4e52-b383-d43061a4c52f",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "# Save results to disk\n",
+    "# pypesto.store.write_result(pypesto_result, f'{name}.hdf5', overwrite=True)"
```
dweindl:
```suggestion
    "# pypesto.store.write_result(pypesto_result, f'{name}.h5', overwrite=True)"
```

python/examples/example_splines_swameye/ExampleSplinesSwameye2003.ipynb
```diff
@@ -0,0 +1,1945 @@
...
+       "<Figure size 650x350 with 1 Axes>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "# Visualize the results of the multistarts\n",
+    "pypesto.visualize.waterfall(pypesto_result, size=[6.5, 3.5])"
```
dweindl:
```suggestion
    "pypesto.visualize.waterfall(pypesto_result, size=[6.5, 3.5]);"
```

python/examples/example_splines_swameye/ExampleSplinesSwameye2003.ipynb
```diff
@@ -0,0 +1,1945 @@
...
...qvX6fG6XZnAdfU1JCQkMChQ4cidhbw73//e+6//34cDgd+v59f//rX/PKXv+yU85VYJdZIxRopXSleiTUyulKsXYnkNTI6Mq+tt7utrq4mPj7+jH27XQFYVlYW0XsBCyGEEEJE06FDh8jOzj5jn25XAFqWxZEjR/B6vWhaZO5t2Vrtt+qOW38iGWurrhBrd85rpHSlrRRdKbeSVyF5jYyOzKtSCp/PR2ZmZpubaHxRZ3Eete7f//Wvf91m3J2O/4pkrJLXrpHXSOlKxyl1pdxKXoXkNTI6c16lADyPTv3QPPVAzHP9MP2i15/LfCMxz46IVfLa+fMaKZHKbSR0pdxKXoXkNTI6e16jfi/gC4lpmtx7773Mnz+f2tracPv8+fPD0891vqc6l/lGYp4dEavktfPnNVIildtI6Eq5lbwKyWtkdPa8drtjADtKc3MzhYWFzJs3D6fTGe1wLhiS18iQvEaO5DYyJK+RIXmNjM6YVykAhRBCCCG6GbkTiBBCCCFENyMFoBBCCCFENyMFoBBCCCFENyMFYIQ8+eST5Obm4nK5GDt2LO+//360Q+pS3nzzTb797W+TmZmJpmmsXLmyzXSlFHfddRcZGRnExMQwZcoUSkpKohNsF1JYWMjo0aPxer2kpqYyY8YM9uzZ06ZPU1MTc+fOJTk5mdjYWL773e9SUVERpYi7hqeeeoohQ4YQFxdHXFwcBQUFvPbaa+HpktPz48EHH0TTNH7xi1+E2yS3X98999yDpmlthv79+4enS07P3uHDh/nBD35AcnIyMTExDB48mC1btoSnd6bvLikAI2DJkiXceuut3H333Wzbto2hQ4cydepUKisrox1al1FfX8/QoUN58skn253++9//nscee4yFCxeyadMmPB4PU6dOpampqYMj7VqKioqYO3cu7733HmvXriUQCPCtb32L+vr6cJ//+I//4JVXXmHp0qUUFRVx5MgRZs6cGcWoO7/s7GwefPBBtm7dypYtW7jsssuYPn06u3btAiSn58PmzZv54x//yJAhQ9q0S27PzsCBAzl69Gh4ePvtt8PTJKdnp6qqigkTJmC323nttdfYvXs3Dz/8MImJieE+neq7K5oXIbxQjRkzRs2dOzf83DRNlZmZqQoLC6MYVdcFqBUrVoSfW5al0tPT1UMPPRRuq66uVk6nU7344otRiLDrqqysVI
```
dweindl: s/node/nodes/

dweindl: prefix with `\n`, for more readable output

dweindl (commented):
Thanks, looks good to me, once comments are addressed.

Nice notebooks.

python/examples/example_splines/ExampleSplines.ipynb
```diff
@@ -0,0 +1,1220 @@
...
+    "    rdata = amici.runAmiciSimulation(model, solver)\n",
+    "    # Plot results\n",
+    "    if plot:\n",
+    "        fig, ax = plt.subplots()\n",
+    "        ax.plot(rdata['t'], rdata['x'])\n",
+    "        ax.set_xlabel(\"time\")\n",
+    "        ax.set_ylabel(\"concentration\")\n",
+    "    return model, rdata\n",
+    "\n",
+    "def plot_spline(spline, parameters=None, xlim=None):\n",
```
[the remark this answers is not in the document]
lcontento: I have moved the plot method to the spline class. I will regenerate the notebooks just before merging.

src/splinefunctions.cpp
```diff
@@ -0,0 +1,1083 @@
...
+     * ]
+     */
+    coefficients_extrapolate_sensi.resize(4 * nplist, 0.0);
+
+    realtype sm0;
+    for (int ip = 0; ip < nplist; ip++) {
+        realtype sp0 = dvaluesdp[spline_offset + ip];
+        switch (first_node_ep_) {
+            /* This whole switch-case-if-else-if-thing could be moved
+             * outside the loop, I know. Yet, it's at most some thousand
```
FFroehlich: thousand what?

FFroehlich: Also it looks like there is a fair bit of duplication between the code for first/last nodes. Also I remember that @paulstapor had some issues with spline performance, so it might be worthwhile going the extra mile to have a proper implementation. should be possible to have a rather concise implementation using function pointers, right?

FFroehlich: above only applies if this function is evaluated more than once per simulation run, otherwise it would be nevertheless be great if duplication between first/last could be reduced.

FFroehlich:
thanks for pushing this through, code looks like it's in pretty good shape to be approved quickly!

src/splinefunctions.cpp
lcontento: It should be only called once per simulation, the coefficients that are computed there depend only on the parameters and stay constant during the simulation.

lcontento: About reducing duplication between fist/last node, it could be done for that function, but I don't think it would be more readable. For the more complex bc function, where reducing duplication would I think make the code definitely less readable.

python/sdist/amici/splines.py
```diff
@@ -0,0 +1,1687 @@
...
+        # Create assignment rule for spline
+        rule = add_assignment_rule(model, self.sbml_id, self.mathml_formula)
+
+        # Add annotation specifying spline method
+        retcode = rule.setAnnotation(self.amici_annotation)
+        if retcode != libsbml.LIBSBML_OPERATION_SUCCESS:
+            raise SbmlAnnotationError('Could not set SBML annotation!')
+
+        # Create additional assignment rule for periodic extrapolation
+        # NB mod is not in the subset of MathML supported by SBML
```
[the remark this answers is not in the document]
lcontento: NB is https://en.wikipedia.org/wiki/Nota_bene
I thought it was common, but I can change it to "Note:"

include/amici/splinefunctions.h
```diff
@@ -0,0 +1,407 @@
...
+
+    /**
+     * @brief Common constructor for `AbstractSpline` instances.
+     * @param nodes the nodes defining the position at which the value of
+     * the spline is known
+     * (if `equidistant_spacing` is true, it must contain only the first and
+     * the last node; the other nodes will be automatically inserted,
+     * assuming they are uniformly spaced)
+     * @param node_values the values assumed by the spline at the nodes
+     * @param equidistant_spacing whether equidistant nodes are to be computed
```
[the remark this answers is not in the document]
lcontento: If `equidistant_spacing` is `false`, the `nodes` must be a vector of nodes. If `equidistant_spacing` is `true`, then `nodes` must only contain two elements, the first and the last node.

python/sdist/amici/de_export.py
```diff
@@ -941,6 +970,26 @@ def import_from_sbml_importer(
             whether to compute conservation laws
         """
 
+        # add splines as expressions to the model
+        # saved for later substituting into the fluxes
+        spline_subs = {}
+
+        for ispl, spl in enumerate(si.splines):
+            # HACK remove real=True assumptions from spline symbols
```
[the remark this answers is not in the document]
dweindl: Fixed that.

python/amici/ode_export.py
dweindl: Fixed that.

include/amici/splinefunctions.h
```diff
@@ -0,0 +1,407 @@
...
+     * @param spline_offset offset of this spline inside `dvaluesdp`
+     * and `dslopesdp`
+     * @param dvaluesdp derivatives of the spline values with respect to the
+     * parameters (for all splines in the model, not just this one)
+     * @param dslopesdp derivatives of the spline derivatives with respect
+     * to the parameters (for all splines in the model, not just this one)
+     * @remark The contents of `dvaluesdp` and `dslopesdp` may be modified
+     * by this function.
+     */
+    virtual void compute_coefficients_sensi(int nplist,
```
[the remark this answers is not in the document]
lcontento: Does resizing and modifying the content of a `std::vector` count for whether the function should be declared const?

include/amici/splinefunctions.h
```diff
@@ -0,0 +1,407 @@
...
+     * at `t`
+     */
+    realtype get_sensitivity(const realtype t, const int ip) const;
+
+    /**
+     * @brief Get the derivative of this spline with respect to a given
+     * parameter at a given point
+     * @param t point at which the sensitivity is to be evaluated
+     * @param ip index of the parameter
+     * @param value value of the spline at the given time point.
```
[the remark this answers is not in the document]
lcontento: You mean the three argument `get_sensitivity`? If the spline interpolation is computed in log-space, then the final function is actually `f(x; p) = exp(spline(x; p))`. Thus the derivative of `f` with respect to the parameter `p` is `exp(spline(x; p)) * dspline/dp(x; p)`. If we already have the value of `spline(x; p)` (which we have in the AMICI simulation loop) we can pass it to the 3-argument version so that it does not get recomputed.

python/sdist/amici/de_export.py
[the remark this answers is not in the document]
lcontento: @paulstapor may know, I did not touch `de_export.py` much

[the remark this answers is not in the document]
dweindl: Removed in https://github.com/AMICI-dev/AMICI/pull/1515/commits/bd5bf0f1dec03a99f246c68ac2bd4e8dbfeb49a2

python/sdist/amici/splines.py
```diff
@@ -0,0 +1,1687 @@
...
+        :param stop:
+            Last point in the grid (some caveats apply, see ``include_stop``)
+        :param step:
+            Desired step size of the grid. Mutually exclusive with ``length``.
+        :param length:
+            Desired length of the grid. Mutually exclusive with ``step``.
+        :param include_stop:
+            Controls the behaviour when ``step`` is not ``None`` and
+            ``stop - start`` is not an integer multiple of ``step``.
+            If ``True`` (default), the actual endpoint of the grid will be
```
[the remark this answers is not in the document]
lcontento: Changed to
```
        :param always_include_stop:
            Controls the behaviour when ``step`` is not ``None``.
            If ``True`` (default), the endpoint is the smallest
            ``start + k * step``, with ``k`` integer, which is
            greater than or equal to ``stop``.
            Otherwise, the endpoint is the largest
            ``start + k * step``, with ``k`` integer, which is
            smaller than or equal to ``stop``.
```

python/sdist/amici/splines.py
```diff
@@ -0,0 +1,1687 @@
...
+
+    @property
+    def stop(self) -> sp.Basic:
+        """Last point."""
+        return self._xx[-1]
+
+    @property
+    def step(self) -> sp.Basic:
+        """Distance between consecutive points."""
+        if len(self._xx) > 1:
```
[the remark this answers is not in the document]
lcontento: Honestly I do not remember. I moved the check at the end of the init method as you suggested and will see if any test breaks (on top of my head I don't see any reason why it should)

python/sdist/amici/splines.py
```diff
@@ -0,0 +1,1687 @@
...
+        """Spline method."""
+        raise NotImplementedError()
+
+    def check_if_valid(self, importer: sbml_import.SbmlImporter) -> None:
+        """
+        Check if the spline described by this object can be correctly
+        be implemented by AMICI. E.g., check whether the formulas
+        for spline grid points, values, ... contain species symbols.
+        """
+        # TODO this is very much a draft
```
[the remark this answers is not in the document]
lcontento: Changed to
```
        # At the moment only basic checks are done.
        # There may still be some edge cases that break
        # the AMICI spline implementation.
        # If found, they should be checked for here
        # until (if at all) they are accounted for.
```

python/sdist/amici/de_export.py
```diff
@@ -1062,6 +1111,8 @@ def transform_dxdt_to_concentration(species_id, dxdt):
         # add fluxes as expressions, this needs to happen after base
         # expressions from symbols have been parsed
         for flux_id, flux in zip(fluxes, si.flux_vector):
+            # replace splines inside fluxes
+            flux = flux.subs(spline_subs)
```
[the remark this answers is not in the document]
lcontento: @paulstapor may know

python/sdist/amici/splines.py
```diff
@@ -0,0 +1,1687 @@
...
+            logarithmic_parametrization: bool = False
+    ):
+        """Base constructor for ``AbstractSpline`` objects.
+
+        :param sbml_id:
+            The SBML ID of the parameter associated to the spline
+            as a string or a SymPy symbol.
+
+        :param x:
+            The point at which the spline is evaluated.
```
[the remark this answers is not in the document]
lcontento: It was never tested for, but you are right that it should have been done. But I had a look at the C++ code and it seems that Paul hardcoded time as the point where the spline is evaluated, so the functionality from the Python interface (`splines.py`) is not carried over to `de_export.py` and C++. I guess Paul decided to concentrate on the low hanging fruit to make KoCo19 work before he left the group. I now raise errors when such splines are read (but still allow them to be created since the piecewise formulas should still be correct, with a warning that AMICI will not work on the resulting SBML files).

python/sdist/amici/splines.py
```diff
@@ -0,0 +1,1687 @@
...
+        valid_extrapolate = (
+            'no_extrapolation',
+            'constant',
+            'linear',
+            'polynomial',
+            'periodic',
+            None
+        )
+
+        for i in (0, 1):
```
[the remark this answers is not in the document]
lcontento: I cannot because I may have to assign `bc[i] = ...`

python/sdist/amici/splines.py
```diff
@@ -0,0 +1,1687 @@
...
+        # TODO this is very much a draft
+        from .de_export import SymbolId
+        fixed_parameters: List[sp.Symbol] = list(
+            importer.symbols[SymbolId.FIXED_PARAMETER].keys())
+        species: List[sp.Symbol] = list(
+            importer.symbols[SymbolId.SPECIES].keys())
+
+        for x in self.xx:
+            if not x.free_symbols.issubset(fixed_parameters):
+                raise ValueError(
```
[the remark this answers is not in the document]
lcontento: If `x` is numeric, `x.free_symbols` is empty and the test passes.

python/sdist/amici/splines.py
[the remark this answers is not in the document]
lcontento: (there are also some pre-existing uses of NB in the AMICI code base)

python/sdist/amici/sbml_utils.py
```diff
@@ -0,0 +1,551 @@
+"""
+SBML Utilities
```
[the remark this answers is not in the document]
lcontento: except the `add_parameter` function, I think the other functions are only used in tests. How about moving them to the test files and just leave `add_parameter` in the AMICI package?

python/sdist/amici/splines.py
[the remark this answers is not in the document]
FFroehlich: :+1:

python/sdist/amici/sbml_utils.py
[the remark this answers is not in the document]
FFroehlich: doesn't matter much where in the codebase the code is, it's stuff that can and should be handled by 3rd party libraries.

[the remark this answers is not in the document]
dweindl: I generally agree on using simplesbml, but in this case, I am not sure:

* These functions have quite to some extra checks, like checking whether a spline target is already a rule target, or for checking for duplicate IDs, which probably make sense in the context of adding splines to an existing model.
* For really use simplesbml, all the sympy math would have to be converted to strings, to then be parsed again by libsbml. For that, we'd have to for example juggle around the piecewise arguments, which probably doesn't make things nicer in the end.

dweindl (approved):
TODO: #2081

PR #1524, opened 2021-07-08
Py: Add num_threads argument to simulate_petab

dweindl:
for parallel simulation of different conditions


python/amici/petab_objective.py
```diff
@@ -78,6 +79,9 @@ def simulate_petab(
         are assumed to be in linear scale.
     :param log_level:
         Log level, see :mod:`amici.logging` module.
+    :param num_threads:
+        Number of threads to use for simulating multiple conditions
+        (only used if compiled with OpenMP)
```
dilpath, line 84:
```suggestion
        (only used if compiled with OpenMP).
```

dilpath (approved):
:+1:

PR #1527, opened 2021-07-08
Update documentation stack

dweindl:


dweindl:
Requires more recent doxygen version (works with 1.9.2 (60732c55c4bd92cb7ac11f173183b93ff63509f7)). Will probably take a while until that's available on RTD.

PR #1528, opened 2021-07-09
Py: Add failfast argument to simulate_petab

dweindl:


python/amici/petab_objective.py
```diff
@@ -51,6 +51,7 @@ def simulate_petab(
         scaled_parameters: Optional[bool] = False,
         log_level: int = logging.WARNING,
         num_threads: int = 1,
+        failfast: bool = True,
```
dilpath, line 54:
```suggestion
        fail_fast: bool = True,
```

python/amici/petab_objective.py
```diff
@@ -82,6 +83,9 @@ def simulate_petab(
     :param num_threads:
         Number of threads to use for simulating multiple conditions
         (only used if compiled with OpenMP).
+    :param failfast:
```
dilpath, line 86:
```suggestion
    :param fail_fast:
```

python/amici/petab_objective.py
```diff
@@ -140,8 +144,9 @@ def simulate_petab(
         amici_model=amici_model)
 
     # Simulate
-    rdatas = amici.runAmiciSimulations(amici_model, solver, edata_list=edatas,
-                                       num_threads=num_threads)
+    rdatas = amici.runAmiciSimulations(
+        amici_model, solver, edata_list=edatas,
+        num_threads=num_threads, failfast=failfast)
```
dilpath, line 149:
```suggestion
        num_threads=num_threads, fail_fast=fail_fast)
```

dilpath (approved):
Also fine as is

dilpath:
Ah, would need to change elsewhere too, nevermind.

PR #1530, opened 2021-07-11
Allow specifying maximum integration time

dweindl:


src/solver_cvodes.cpp
```diff
@@ -996,7 +1024,16 @@ static int froot(realtype t, N_Vector x, realtype *root,
...
-    auto model = static_cast<Model_ODE *>(user_data);
+    auto typed_udata = static_cast<CVodeSolver::user_data_type *>(user_data);
+    Expects(typed_udata);
+    auto model = dynamic_cast<Model_ODE *>(typed_udata->first);
+    Expects(model);
+    auto solver = dynamic_cast<CVodeSolver *>(typed_udata->second);
+    Expects(model);
+
+    if(solver->timeExceeded()) {
+        throw IntegrationFailure(AMICI_MAX_TIME_EXCEEDED, t);
```
FFroehlich, line 1035: `return AMICI_MAX_TIME_EXCEEDED`?

FFroehlich (approved):

src/solver_cvodes.cpp
dweindl: Good point, thanks.

Issue #672, opened 2019-03-26
Check condition number in NewtonSolver::getSens


FFroehlich:
NewtonSolver::getSens() willingly solved the sensitivity linear systems for a model that had a jacobian with frobenius norm ~`1e32`. Might be reasonable to warn the user that the computed sensitivities are going to be complete garbage.

paulstapor:
I'm wondering whether or not it might be reasonable to directly compute the rank of the Jacobian in the python part already. When done, it would be clear in advance whether or not the Newton solver can actually be used or not. This would allow to avoid the Newton solver directly, if it won't be applicable anyway...

paulstapor:
As condition numbers are not necessarily easy to compute, I was thinking that it might be reasonable to include some LA library for this. I think it's rather not the best idea to add routines for SVD or QR decomposition by hand here which would allow for condition number computation...

I was thinking of adding (and then compiling and linking) [Armadillo](http://arma.sourceforge.net/docs.html) for this task, but maybe someone of you has better ideas for this...

FFroehlich:
What about just using `KLU_condest` https://github.com/AMICI-dev/AMICI/blob/d16078e8b6ac8b4ae31d4e7ff01738bc0291ccdd/ThirdParty/SuiteSparse/KLU/Source/klu_diagnostics.c#L164 or `KLU_rcond` https://github.com/AMICI-dev/AMICI/blob/d16078e8b6ac8b4ae31d4e7ff01738bc0291ccdd/ThirdParty/SuiteSparse/KLU/Source/klu_diagnostics.c#L489 ?

paulstapor:
Nice!
Didn't know those existed... Yes, sure, that's preferred then!

PR #1537, opened 2021-07-27
Switch the default steady-state sensitivity method to simulationFSA.

stephanmg:
Changing the default steady-state sensitivity method to `simulationFSA` might increase the computation time but on the other hand be more robust across benchmark models.

### DIFF
--- a/documentation/python_installation.rst
+++ b/documentation/python_installation.rst
@@ -404,7 +404,7 @@ Note that this might require to satify missing dependencies manually as
 specified by PEP 518 (Dependencies must already be installed if using the 
 no build isolation option) via 
 
-.. code-blocks:: bash
+.. code-block:: bash
 
    pip install --user dependency
 
--- a/documentation/python_installation.rst
+++ b/documentation/python_installation.rst
@@ -387,6 +387,27 @@ installation is done completely anew.
 
 Now, you are ready to use AMICI in the virtual environment.
 
+In case you desire not to use AMICI in a virtual environment, you can install
+via
+
+.. code-block:: bash
+
+    pip install --user --no-build-isolation -e python/sdist
+
+after running
+
+.. code-block:: bash
+
+   scripts/buildAll.sh
+
+Note that this might require to satify missing dependencies manually as
+specified by PEP 518 (Dependencies must already be installed if using the 
+no build isolation option) via 
+
+.. code-blocks:: bash
+
+   pip install --user dependency
+
 .. note::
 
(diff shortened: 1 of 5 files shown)


paulstapor (approved):

dweindl:
That may break existing applications (using adjoint sensitivities), right?

dweindl:
> That may break existing applications (using adjoint sensitivities), right?

Ah, no, I think it shouldn't. But it may lead to surprisingly long simulation times for people relying on the current defaults?

Please include a few words on the rationale for the change and potential additional (undesired) consequences in the PR description.

paulstapor:
> Please include a few words on the rationale for the change and potential additional (undesired) consequences in the PR description.

hm... I think in those cases, where it's currently used, it will rather lead to AMICI throwing an error, since the new default can't be combined with adjoints. Hence, one will have to set it to newtonOnly by hand. However, There would be a good way around, which is however slightly more complex than a one-liner...
(Can outline it, if helpful...)

dweindl:
> I think in those cases, where it's currently used, it will rather lead to AMICI throwing an error, since the new default can't be combined with adjoints

Hm. Does it then really solve more problems than it creates?

FFroehlich:
> > I think in those cases, where it's currently used, it will rather lead to AMICI throwing an error, since the new default can't be combined with adjoints
> 
> Hm. Does it then really solve more problems than it creates?

Agreed, I think fixing https://github.com/AMICI-dev/AMICI/issues/672 would be the better approach and shouldn't require too much effort.

paulstapor:
> Hm. Does it then really solve more problems than it creates?

Afaik, it only affects the CanPathPro related things... Because models, which have a looong runtime and where you can't use the simulationFSA mode but really *must* use adjoints: That's atm exclusively CanPathPro... However, chaging that default to something more stable would help quite some people I think...

dweindl:
> Afaik, it only affects the CanPathPro related things... Because models, which have a looong runtime and where you can't use the simulationFSA mode but really _must_ use adjoints: That's atm exclusively CanPathPro...

If it will break stuff for everybody using preequilibration with adjoints and unset steadystate sensitivity mode, it doesn't really matter whether or not adjoints are "required" for the given application.

I vote for leaving it as it is, as I don't see a big enough advantage, but I don't feel too strong about it.

stephanmg:
So shall we close this issue as `invalid` in favor of issue #672?

stephanmg:
Closed for now in favor of another issue #672.

PR #1539, opened 2021-08-05
Adjoint sensitivities for Python-import models with events

paulstapor:


paulstapor:
Okay, this seems to be indeed the last status of my work on adjoints with events. Sorry it took me so long to actually double-check it.
In principle, the implemented algorithm seems to work fine for multiple test models, but it is - for some reason that I failed to understand - inexact, i.e., gradients are off by a little. This problem only occurred for the Leonhardt test model (purely parameter dependent event with state update). However, this little (in the range of a few percent) being off markedly affected the performance of parameter estimation, as I checked in some (by now) deleted file.

FFroehlich:
> Okay, this seems to be indeed the last status of my work on adjoints with events. Sorry it took me so long to actually double-check it. In principle, the implemented algorithm seems to work fine for multiple test models, but it is - for some reason that I failed to understand - inexact, i.e., gradients are off by a little. This problem only occurred for the Leonhardt test model (purely parameter dependent event with state update). However, this little (in the range of a few percent) being off markedly affected the performance of parameter estimation, as I checked in some (by now) deleted file.
> 
> I don't really know what to do with this. The work done here enables adjoint computations for a couple of models with events and heaviside. However, it seems to have a bug for certain types of models with events. Possible reasons might be:
> 
> * an error in the implementation
> * a problem with interpolation of the forward solution (which is discontinuous), or, as a result, with the right hand side, or dfdp

when using adjoint sensitivities with events, one should _always_ use polynomial instead of hermite interpolation.

stephanmg:
> 

Super late to the party... but could this be the issue of the imprecision?

It seems like Hermite interpolation is used for the forward problem solution, which is also used for the backwards problem.

dweindl:
> > 
> 
> Super late to the party... but could this be the issue of the imprecision?
> 
> It seems like Hermite interpolation is used for the forward problem solution, which is also used for the backwards problem.

Does changing that fix the issue? :)

stephanmg:
> > > 
> > 
> > 
> > Super late to the party... but could this be the issue of the imprecision?
> > It seems like Hermite interpolation is used for the forward problem solution, which is also used for the backwards problem.
> 
> Does changing that fix the issue? :)



> > > 
> > 
> > 
> > Super late to the party... but could this be the issue of the imprecision?
> > It seems like Hermite interpolation is used for the forward problem solution, which is also used for the backwards problem.
> 
> Does changing that fix the issue? :)

Will give it a try...

dweindl:
Resolved some conflicts and updated this branch. 

Current status is that most tests pass and sensitivities look good, but for two cases, there are still issues, rather related to implementation than to choice of tolerances:

* `python/tests/test_events.py::test_models[events_plus_heavisides]`
  
  ```
               fsa        asa         fd      abs_diff      rel_diff
  k1      5.740767   5.740767   5.740766  4.469314e-12  7.785221e-13
  k2      4.146212   4.146212   4.146213 -7.933210e-12 -1.913363e-12
  k3     -2.278242  -2.278242  -2.278242  2.761347e-12 -1.212051e-12
  alpha  -0.593429  -0.593429  -0.593429 -4.544578e-09  7.658161e-09
  beta   -1.596328  -1.596336  -1.596324  8.518073e-06 -5.336042e-06 <------ ?!
  gamma   7.464110   7.464110   7.464111 -5.656077e-09 -7.577698e-10
  delta   6.290323   6.290323   6.290322 -7.910117e-12 -1.257506e-12
  eta    15.224854  15.221247  15.224853  3.607176e-03  2.369268e-04 <------ !!
  zeta   -9.385052  -9.385052  -9.385052 -3.037570e-13  3.236605e-14
  ```

* `python/tests/test_events.py::test_models[nested_events]`
  
  ```
                 fsa       asa            fd      abs_diff      rel_diff
  k1        3.358876  3.358876  3.358875e+00 -1.590994e-11 -4.736686e-12
  k2        0.952255  0.952255  9.522543e-01 -2.982503e-12 -3.132044e-12
  inflow_1  1.888090  1.888090  1.888090e+00 -1.052491e-12 -5.574370e-13
  decay_1  -2.850859 -2.850859 -2.850860e+00  2.958522e-12 -1.037765e-12
  decay_2   1.150632  1.150112  1.150632e+00  5.200652e-04  4.519821e-04 <------ !!
  bolus     4.712958  4.712958  4.426220e+06 -1.454481e-11 -3.086132e-12 <------ fd?!
  ```
  
  Any contributions would be highly appreciated.

python/sdist/amici/de_model.py
```diff
@@ -1767,6 +1767,65 @@ def _compute_equation(self, name: str) -> None:
...
+                    self.sym("xdot") - self.sym("xdot_old"),
+                    self.eq("dtaudx")[ie],
+                )
+                if event.updates_state:
+                    # ==== 2nd group of terms: Derivatives of Dirac deltas ===
+                    # Part 2a: explicit time dependence of bolus function
+                    tmp_eq -= smart_multiply(
+                        self.eq("ddeltaxdt")[ie], self.eq("dtaudx")[ie]
+                    )
+                    # Part 2b: implicit time dependence of bolus function
```
FFroehlich, line 1787: This addition of dx+dx_old looks fishy to me, is this to avoid explicit dependence on `useValueFromTriggerTime`?

dweindl: Yes, this is due to `useValueFromTriggerTime`. Previously, `deltax` only contained `x`, but now it may depend on both `x` and `x_old`. This is also done for forward sensitivities and works there:

https://github.com/AMICI-dev/AMICI/blob/ce28017159c9172a93b3b5ad6c602ebd3c0cd9d3/python/sdist/amici/de_model.py#L1760-L1764

dweindl:
This PR extends support for adjoint sensitivity analysis for models with events.
This does not yet cover all cases. Checking gradients for a given model is still highly recommended.

Changes:
* Generate functions for computing the adjoint state and quadrature updates (`fdeltaxB`, `fdeltaqB`) during Python model import
* Add Python-specific functions with updated signature for `fdeltaxB`, `fdeltaqB`
* Store additional information at discontinuities
* Test adjoint sensitivities in existing test cases. Some failure are currently ignored and require further investigation.

FFroehlich (approved):

PR #1543, opened 2021-09-01
Fix: Argument "outdir" in ODEExporter.__init__ only stored but never …

### LINKED ISSUE #1542: Argument "outdir" in ODEExporter.__init__ only stored but never used
As a consequence,  `output_dir` in  `SbmlImporter.sbml2amici` is ignored.

There is also some redundancy in path management in `ODEExporter.set_paths` and `.__init__` that should be cleaned up.

dweindl:
…used

* Honor user-provided output directory (Fixes #1542)
* Remove unused instance variable ODEExporter.output_dir
* Remove redundant logic for setting model_path and model_swig path
* Handle default directory in ODEExporter.set_paths

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -2702,8 +2702,11 @@ class ODEExporter:
 
     def _prepare_model_folder(self) -> None:
         """
-        Remove all files from the model folder.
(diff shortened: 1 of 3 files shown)


FFroehlich (commented):
👍

PR #1546, opened 2021-09-09
Improve checking support for SBML extensions

### LINKED ISSUE #1545: SBML import: check for SBML extensions does not work for SBML-multi
The check for SBML extensions during model import does not work for SBML-multi:

https://github.com/AMICI-dev/AMICI/blob/bd3b89d3839e6518532d547e54e7fe66ef3cfec3/python/amici/sbml_import.py#L409-L411

Multi-elements don't show up in that list. Not sure about other SBML extensions.

Might be preferable to check for enable packages and their required attribute.

dweindl:
Ignore enabled extensions that do not affect model equations or that are not actually used.

Closes #1545

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -419,7 +419,10 @@ class SbmlImporter:
             # the "required" attribute is only available in SBML Level 3
             for i_plugin in range(self.sbml.getNumPlugins()):
                 plugin = self.sbml.getPlugin(i_plugin)
-                if not plugin.getRequired():
+                if plugin.getPackageName() in ('layout',):
+                    # 'layout' plugin does not have the 'required' attribute
+                    continue
+                if hasattr(plugin, 'getRequired') and not plugin.getRequired():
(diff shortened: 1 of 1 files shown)


python/amici/sbml_import.py
```diff
@@ -406,9 +406,20 @@ def check_support(self) -> None:
...
         rule, or assignment rule, to produce change in the system over time.
         """
-        if hasattr(self.sbml, 'all_elements_from_plugins') \
-                and self.sbml.all_elements_from_plugins.getSize():
-            raise SBMLException('SBML extensions are currently not supported!')
+
+        # Check for required but unsupported SBML extensions
+        for i_plugin in range(self.sbml.getNumPlugins()):
+            plugin = self.sbml.getPlugin(i_plugin)
+            if not plugin.getRequired():
```
dweindl, line 413: Meh, that's only a SBML Level 3 thing...

FFroehlich (approved):

PR #1548, opened 2021-09-13
Add sllh computation back to `petab_objective.simulate_petab`

dilpath:
Was removed in #1498 due to erroneous implementation.

- parameters are now scaled within `simulate_petab`
- added FD checks, essentially copied from `pypesto.objective.ObjectiveBase.check_grad{_multi_eps}` [1]
- added a test case: a PEtab problem based on the Lotka-Volterra example in `yaml2sbml` [2], extended to test more of the `simulate_petab` SLLH computation with
    - multiple experimental conditions
    - preequilibration

[1] https://github.com/ICB-DCM/pyPESTO/blob/32a5daf888dd6c49fa3ca3b8b39e055b6f15ca9f/pypesto/objective/base.py#L397-L620
[2] https://github.com/yaml2sbml-dev/yaml2sbml/tree/773f4c563d90689cd974da8c114b6c7eaf316802/doc/examples/Lotka_Volterra/Lotka_Volterra_python


tests/benchmark-models/test_petab_model.py
```diff
@@ -61,6 +74,98 @@ def parse_cli_args():
...
+                continue
+
+            backward_llh = simulate(backward_parameters)[LLH]
+            forward_llh = simulate(forward_parameters)[LLH]
+
+            fd = (forward_llh - backward_llh) / (2 * epsilon)
+            central_differences[x_id].append(fd)
+
+    min_fd = {
+        x_id: np.nanmin(differences)
```
dweindl, line 149:
```suggestion
        x_id: np.nanmin(np.abs(differences))
```
?

python/amici/petab_objective.py
```diff
@@ -55,6 +60,10 @@ def simulate_petab(
 ) -> Dict[str, Any]:
     """Simulate PEtab model.
 
+    NB: supplied parameters are scaled internally, independently of
```
dweindl, line 63: something like...?

```suggestion
.. note::
    Supplied parameters are scaled internally according to `parameterScale` in the parameter table, independently of
```

dweindl (approved):
LGTM, but please wait for others' feedback.

Why was the extra test model required? Would keep things simpler if we could use benchmark collection model that's already used elsewhere.

python/amici/petab_objective.py
yannikschaelte: Clarify: What you mean is: Regardless of `scaled_parameters`, sensitivities will always be on the scale as specified in `parameterScale` in the parameter table?

python/amici/petab_objective.py
```diff
@@ -150,6 +160,15 @@ def simulate_petab(
 
     # Compute total llh
     llh = sum(rdata['llh'] for rdata in rdatas)
+    # Compute total sllh
```
yannikschaelte, line 163: I guess this was not the case before either, but maybe one could specify whether or not to calculate sensis? In the latter case, the simulation in runAmiciSimulations would be cheaper, and this mapping here not be performed.

python/amici/petab_objective.py
```diff
@@ -159,10 +178,116 @@ def simulate_petab(
...
+        AMICI model from which ``rdatas`` were obtained.
+    :param rdatas:
+        Simulation results.
+    :param parameter_mapping:
+        PEtab parameter mapping to condition-specific
+            simulation parameters
+    :param edatas:
+        Experimental data used for simulation.
+    :param petab_scale:
+        Return sensitivities on the PEtab parameter scales
```
yannikschaelte, line 208: alternative?

python/tests/test_petab_objective.py
```diff
@@ -0,0 +1,43 @@
...
+    amici_solver.setSensitivityOrder(amici.SensitivityOrder_first)
+
+    result = amici.petab_objective.check_grad_multi_eps(
+        petab_problem=petab_problem,
+        amici_model=amici_model,
+        amici_solver=amici_solver,
+        detailed=True,
+    )
+
+    assert (result.abs_err < ATOL).all()
```
yannikschaelte, line 42: we discussed this, right? seems plausible to me this way. if problematic, maybe use not the optimal parameter vector.

python/amici/petab_objective.py
```diff
@@ -159,10 +178,116 @@ def simulate_petab(
...
+    :return:
+        Aggregated likelihood sensitivities.
+    """
+    accumulated_sllh = {}
+    model_parameter_ids = amici_model.getParameterIds()
+
+    if petab_scale and petab_problem is None:
+        raise ValueError(
+            'Please provide a PEtab problem if scaled SLLH is desired, or set '
+            '`petab_scale==False`.'
```
yannikschaelte, line 221:
```suggestion
            '`petab_scale=False`.'
```

python/amici/petab_objective.py
```diff
@@ -159,10 +178,116 @@ def simulate_petab(
...
+        )
+
+    # Check for issues in all condition simulation results.
+    for rdata in rdatas:
+        # Condition failed during simulation.
+        if rdata['status'] != amici.AMICI_SUCCESS:
+            return None
+        # Condition simulation result does not provide SLLH.
+        if rdata.get('sllh', None) is None:
+            return None
```
yannikschaelte, line 231: one could argue about whether to silently return None or cry out if sth is inconsistent, but anything fine for me.

python/amici/petab_objective.py
```diff
@@ -159,10 +178,116 @@ def simulate_petab(
...
+
+            # Skip fixed parameters (?)
+            if not isinstance(petab_parameter_id, str):
+                continue
+
+            # Initialize
+            if petab_parameter_id not in accumulated_sllh:
+                accumulated_sllh[petab_parameter_id] = 0
+
+            partial_parameter_sllh = condition_parameter_sllh
```
yannikschaelte, line 258:
```suggestion
            # rename for readability
            partial_parameter_sllh = condition_parameter_sllh
```
?

python/amici/petab_objective.py
```diff
@@ -159,10 +178,116 @@ def simulate_petab(
...
+                )
+                petab_parameter_scale = (
+                    petab_problem
+                    .parameter_df
+                    .loc[petab_parameter_id, PARAMETER_SCALE]
+                )
+                # SLLH is (un)scaled here with methods that were written for
+                # parameters.
+                # First unscale w.r.t. model scale.
+                partial_parameter_sllh = unscale_parameter(
```
yannikschaelte, line 275: I thought we had somewhere also a function rescale(val, scale_in, scale_out), but nvm

yannikschaelte (approved):
if llh and sllh are correct (in particular the gradient check is happy), LGTM

python/tests/test_petab_objective.py
dilpath: We discussed something similar in pyPESTO [1], where an optimized vector is used, and the value of `RTOL` is high (`1`) (especially given an optimized vector).

Here, `RTOL` is smaller (`1e-3`), and an optimized vector is not used (also, the measurements here are nonsense).

[1] https://github.com/ICB-DCM/pyPESTO/blob/main/test/petab/test_amici_objective.py#L116

dilpath:
> LGTM, but please wait for others' feedback.
> 
> Why was the extra test model required? Would keep things simpler if we could use benchmark collection model that's already used elsewhere.

The extra test model is because a suitable benchmark problem may not exist. I tested a few with pyPESTO and some errors were always high. I could look for more (I may have also been using an optimized vector).

The extra test model might still be a good idea, since many/all PEtab features could be added to it that do not all exist in a single benchmark problem simultaneously.

python/tests/test_petab_objective.py
yannikschaelte: sounds good then

python/amici/petab_objective.py
dilpath: The optional `solver` argument is expected to be provided and already setup to compute sensitivities, if sensitivities are desired. Could default to first order forward? Either makes sense to me (no default at least ensures the user defines how sensitivities should be computed). Before review, it defaulted to `None`, and was handled in `aggregate_sllh`. Now defaults to `None` here, and `aggregate_sllh` is only called if sensitivities were computed (so, an error is now raised there, as per your other suggestion).

python/amici/petab_objective.py
```diff
@@ -159,10 +180,119 @@ def simulate_petab(
...
+            )
+
+    for condition_parameter_mapping, edata, rdata in \
+            zip(parameter_mapping, edatas, rdatas):
+        for sllh_parameter_index, condition_parameter_sllh in \
+                enumerate(rdata['sllh']):
+            # Get PEtab parameter ID
+            # Use ExpData if it provides a parameter list, else default to
+            # Model.
+            if edata.plist is not None and edata.plist:
```
FFroehlich, line 246:
```suggestion
            if edata.plist:
```

python/amici/petab_objective.py
```diff
@@ -159,10 +180,119 @@ def simulate_petab(
...
+            '`petab_scale=False`.'
+        )
+
+    # Check for issues in all condition simulation results.
+    for rdata in rdatas:
+        # Condition failed during simulation.
+        if rdata['status'] != amici.AMICI_SUCCESS:
+            return None
+        # Condition simulation result does not provide SLLH.
+        if rdata.get('sllh', None) is None:
```
FFroehlich, line 233:
```suggestion
        if rdata.sllh is None:
```

python/amici/petab_objective.py
```diff
@@ -159,10 +180,119 @@ def simulate_petab(
...
+    if petab_scale and petab_problem is None:
+        raise ValueError(
+            'Please provide a PEtab problem if scaled SLLH is desired, or set '
+            '`petab_scale=False`.'
+        )
+
+    # Check for issues in all condition simulation results.
+    for rdata in rdatas:
+        # Condition failed during simulation.
+        if rdata['status'] != amici.AMICI_SUCCESS:
```
FFroehlich, line 230:
```suggestion
        if rdata.status != amici.AMICI_SUCCESS:
```

python/amici/petab_objective.py
```diff
@@ -159,10 +180,119 @@ def simulate_petab(
...
+        if rdata.get('sllh', None) is None:
+            raise ValueError(
+                'The sensitivities of the likelihood for a condition were '
+                'not computed.'
+            )
+
+    for condition_parameter_mapping, edata, rdata in \
+            zip(parameter_mapping, edatas, rdatas):
+        for sllh_parameter_index, condition_parameter_sllh in \
+                enumerate(rdata['sllh']):
```
FFroehlich, line 242:
```suggestion
                enumerate(rdata.sllh):
```

python/amici/petab_objective.py
```diff
@@ -159,10 +180,119 @@ def simulate_petab(
...
+                model_parameter_index = edata.plist[sllh_parameter_index]
+            else:
+                model_parameter_index = amici_model.plist(sllh_parameter_index)
+            model_parameter_id = model_parameter_ids[model_parameter_index]
+            petab_parameter_id = (
+                condition_parameter_mapping
+                .map_sim_var[model_parameter_id]
+            )
+
+            # Skip fixed parameters (?)
```
FFroehlich, line 256: indices for parameters with numerical values are excluded from plist so this check can be removed.

python/amici/petab_objective.py
```diff
@@ -159,10 +180,119 @@ def simulate_petab(
...
+                )
+                petab_parameter_scale = (
+                    petab_problem
+                    .parameter_df
+                    .loc[petab_parameter_id, PARAMETER_SCALE]
+                )
+                # SLLH is (un)scaled here with methods that were written for
+                # parameters.
+                # First unscale w.r.t. model scale.
+                condition_parameter_sllh = unscale_parameter(
```
FFroehlich, line 280: I am confused what this is supposed to do? The factors from the chain rule would be a bit different, no?

FFroehlich (commented):
Not sure whether the implementation here is correct. Might also explain why errors are high for petab test models. I agree with Daniel that we should ideally get this working on all of the examples from the benchmark collection as well.

python/amici/petab_objective.py
dilpath: Ah, leftover from a while ago I think. I think the scales have always matched so will remove this `if` block and replace with an error if they don't match

python/amici/petab_objective.py
dweindl: > The optional solver argument is expected to be provided and already setup to compute sensitivities, if sensitivities are desired.

Yes!

dilpath:
> Not sure whether the implementation here is correct. Might also explain why errors are high for petab test models. I agree with Daniel that we should ideally get this working on all of the examples from the benchmark collection as well.

Sure, then I'll compare the gradients for all benchmark problems between AMICI and pyPESTO. Might take a week or so due to other commitments.

FFroehlich:
@dilpath just had a look at this as it would be helpful in https://github.com/AMICI-dev/AMICI/tree/jax_notebook. In the current implementation, gradients are wrong because `simulate_petab` now applies scaling, but this transformation is not accounted for in gradient computation. This means that this scaling is either removed and handled via amici (preferable) or transformation needs to be accounted for during aggregation (okay if not possible to handle otherwise).

dilpath:
> @dilpath just had a look at this as it would be helpful in #1996. In the current implementation, gradients are wrong because `simulate_petab` now applies scaling, but this transformation is not accounted for in gradient computation. This means that this scaling either needs to be removed and handled via amici (preferable) or transformation needs to be accounted for during aggregation (okay if not possible to handle otherwise).

Thanks for the feedback! Gradients were computed w.r.t. scaled parameters before, for optimization. They are now w.r.t. unscaled parameters by default, with a flag to switch to scaled parameters.

FFroehlich:
requires https://github.com/ICB-DCM/fiddy/pull/18 and https://github.com/ICB-DCM/fiddy/pull/17

python/sdist/amici/petab_objective.py
```diff
@@ -153,6 +164,29 @@ def simulate_petab(
...
+    if solver.getSensitivityOrder() != amici.SensitivityOrder.none:
+        sllh = aggregate_sllh(
+            amici_model=amici_model,
+            rdatas=rdatas,
+            parameter_mapping=parameter_mapping,
+            petab_scale=scaled_parameters,
+            petab_problem=petab_problem,
+            edatas=edatas,
+        )
+        if not scaled_gradients:
```
dweindl, line 178:
```suggestion
        if not scaled_gradients and sllh is not None:
```

may be None in case of integration errors. see notebook failures on RTD.

dilpath:
The current status is that the test for each benchmark model needs to be made to pass, e.g. by finding a vector for each model that reliably passes the gradient check, and storing these vectors in the test.

FFroehlich:
> The current status is that the test for each benchmark model needs to be made to pass, e.g. by finding a vector for each model that reliably passes the gradient check, and storing these vectors in the test.

thanks! will try to get this wrapped up this week

tests/benchmark-models/test_petab_benchmark.py
```diff
@@ -0,0 +1,173 @@
...
+        'Raimundez_PCB2020',
+        'Lucarelli_CellSystems2018',
+        'Isensee_JCB2018',
+        'Beer_MolBioSystems2014',
+        'Alkan_SciSignal2018',
+        # excluded due to excessive numerical failures
+        'Crauste_CellSystems2017',
+        'Fujita_SciSignal2010',
+    )
+    and str(petab_path.stem) == 'Okuonghae_ChaosSolitonsFractals2020'
```
dweindl: I am confused. Does this only select `Okuonghae_ChaosSolitonsFractals2020`?

tests/benchmark-models/test_petab_benchmark.py
```diff
@@ -0,0 +1,173 @@
...
+from fiddy.extensions.amici import (
+    simulate_petab_to_cached_functions,
+)
+
+
+# Absolute and relative tolerances for finite difference gradient checks.
+ATOL: float = 1e-3
+RTOL: float = 1e-2
+
+benchmark_path = Path(__file__).parent.parent.parent / "Benchmark-Models-PEtab" / "Benchmark-Models"
```
dweindl, line 25: Could use the benchmark models package to simplify things, but fine as is.

tests/benchmark-models/test_petab_benchmark.py
```diff
@@ -0,0 +1,173 @@
...
+        # excluded due to excessive numerical failures
+        'Crauste_CellSystems2017',
+        'Fujita_SciSignal2010',
+    )
+    and str(petab_path.stem) == 'Okuonghae_ChaosSolitonsFractals2020'
+]
+
+debug = False
+if debug:
+    debug_path = Path('../../python/tests/debug')
```
dweindl, line 48: please change to something `__file__`-based.

python/tests/petab_test_problems/lotka_volterra/model/writer.py
```diff
@@ -0,0 +1,19 @@
+from pathlib import Path
+
+import petab
+import yaml2sbml
+
+
+yaml2sbml_yaml = 'lotka_volterra.yaml'
+petab_path = Path('..') / 'petab'
```
dweindl, line 8: please use `__file__`-based path

python/tests/test_petab_objective.py
```diff
@@ -0,0 +1,80 @@
...
+                    ),
+                    scaled_parameters=scaled_parameters,
+                    scaled_gradients=scaled_gradients,
+                )
+            )
+
+    # Each combination of `scaled_parameters` and `scaled_gradients` passes the
+    # gradient check.
+    for result_id, result in results.items():
+        assert (result.rel_err < ATOL).all(), result_id
```
dweindl, line 62:
```suggestion
        assert (result.abs_err < ATOL).all(), result_id
```
?

python/tests/petab_test_problems/lotka_volterra/model/convert_to_petab.sh
dweindl, line 1: Please add `#!` and make paths relative to file location

python/tests/petab_test_problems/lotka_volterra/model/convert_to_petab.sh
```diff
@@ -0,0 +1,21 @@
+petab_path=../petab
```
dweindl, line 1:
```suggestion
#!/bin/bash
set -eou pipefail
script_dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
petab_path=$script_dir/../petab
```

+ add script_dir to other filenames used below

documentation/example_errors.ipynb
dweindl, line 1: Was this supposed to be changed here? Not sure. If not, please revert to version from `develop` (i.e. symlink)

.github/workflows/test_benchmark_collection_models.yml
```diff
@@ -60,6 +60,12 @@ jobs:
         git clone --depth 1 https://github.com/benchmarking-initiative/Benchmark-Models-PEtab.git \
           && export BENCHMARK_COLLECTION="$(pwd)/Benchmark-Models-PEtab/Benchmark-Models/" \
           && AMICI_PARALLEL_COMPILE=2 tests/benchmark-models/test_benchmark_collection.sh
+          
+    # run gradient checks
+    - name: Run Gradient Checks
+      run: |
+        pip install git+https://github.com/ICB-DCM/fiddy.git \
```
dweindl, line 67: Does that reuse the imported models from the test above, or does it re-import everything?

dweindl (commented):
tbc

python/sdist/amici/petab_objective.py
```diff
@@ -811,3 +993,322 @@ def rdatas_to_simulation_df(
...
+    amici_solver: amici.Solver,
+    problem_parameters: Optional[Dict[str, float]] = None,
+    scaled_parameters: bool = False,
+    parameter_ids: Sequence[str] = None,
+    eps: float = 1e-5,
+    verbosity: int = 1,
+    detailed: bool = False,
+    ignore_missing_parameters: bool = False,
+    scaled_gradients: bool = True,
+) -> pd.DataFrame:
```
dweindl, line 1069: Why not fiddy?

dweindl (approved):
:+1:

python/sdist/amici/petab_objective.py
FFroehlich: good point, @dilpath ?

python/sdist/amici/petab_objective.py
```diff
@@ -53,9 +58,14 @@ def simulate_petab(
         log_level: int = logging.WARNING,
         num_threads: int = 1,
         failfast: bool = True,
+        scaled_gradients: bool = False,
 ) -> Dict[str, Any]:
     """Simulate PEtab model.
 
+    .. note::
+        Regardless of `scaled_parameters`, sensitivities are returned on the
+        scales defined in the PEtab parameters table.
```
dilpath, line 67:
```suggestion
        Regardless of `scaled_parameters`, unscaled sensitivities are returned,
        unless `scaled_gradients=True`.
```

python/sdist/amici/petab_objective.py
dilpath: Agreed, will do

PR #1551, opened 2021-09-21
SBML import: Use more descriptive IDs for flux expressions

dweindl:
Previously, those were flux_r0, flux_r1, ...; now they include the reaction ID if it was set in the imported SBML model.


FFroehlich (commented):
👍

PR #1594, opened 2021-11-18
add parsing support for sympy Piecewise, fixes #1591

### LINKED ISSUE #1591: Piecewise functions are not parsed as Heaviside for PySB
pysb allows `Piecewise` in expressions, needs parsing for correct simulation/sensitivities.

FFroehlich:


python/amici/pysb_import.py
```diff
@@ -10,6 +10,10 @@
     Expression, LogLikelihood, generate_measurement_symbol
 )
 
+from .sbml_import import (
+    _parse_special_functions
+)
```
dweindl, line 15: Let's move that function to `.import_utils` instead of pulling SBML stuff into pysb import.

dweindl (approved):

PR #1595, opened 2021-11-19
Fix recursion depth computation

FFroehlich:


python/amici/ode_export.py
```diff
@@ -2027,9 +2027,10 @@ def _derivative(self, eq: str, var: str, name: str = None) -> None:
             nonzeros = np.asarray(
                 derivative.applyfunc(lambda x: int(not x.is_zero))
             ).astype(np.int64)
-            if max(nonzeros.shape):
-                while nonzeros.max():
-                    nonzeros = nonzeros.dot(nonzeros)
+            recursion = nonzeros.copy()
```
dilpath, line 2030: Could be

```python
if not nonzeros.is_nilpotent():
    raise ...
nilpotent_matrix = nonzeros.copy()
nilpotent_index = 1
while not nilpotent_matrix.is_zero_matrix():
    nilpotent_index += 1
    nilpotent_matrix = nilpotent_matrix * nonzeros

**Review by dilpath — APPROVED** [review]
:+1:

File: python/amici/ode_export.py [hunk 2]
```diff
@@ -2027,9 +2027,10 @@ def _derivative(self, eq: str, var: str, name: str = None) -> None:
...
+            if max(recursion.shape):
+                while recursion.max():
```
dweindl, line 2032:
```suggestion
                while recursion.max():
```
Does this need `abs`?

dweindl (approved):

python/amici/ode_export.py
FFroehlich: ah we intentionally convert to a numpy matrix here though since sympy often has performance issues with the simplest operations (including `is_zero_matrix` as far as I remember) because of type checks and simplifications etc. I would assume that checking `is_nilpotent` either does exactly the same operations as implemented in the remaining code (but without vectorization, which can only be applied for numerical matrices).

python/amici/ode_export.py
FFroehlich: no. `recursion` in effectively counts the number of paths of length `self._w_recursion_depth` from i to j, so the number is always positive.

PR #1599, opened 2021-11-30
Require sympy>=1.9,pysb>=1.13.1 / update Heaviside

dweindl:
Require sympy>=1.9, and pysb>=1.13.1 to be compatible with the newer sympy.

Adapt to changes in sympy's Heaviside signature


FFroehlich:
👍

PR #1606, opened 2021-12-04
Remove sympy patch for min/max code printing

dweindl:
Obsolete after requiring sympy>=1.9.


FFroehlich (approved):
👍  do we also want to remove the _monkeypatched function or is this still employed somewhere else?

dweindl:
> +1 do we also want to remove the _monkeypatched function or is this still employed somewhere else?

Thought about it. Would keep it. Although it should probably be moved to a more suitable module.
If you prefer to remove it, I'd do it in a separate PR.

PR #1609, opened 2021-12-06
Remove assert_fun argument from gradient checking, improve output

dweindl:
Breaking change:
* previously required argument `assert_fun` is removed from amici.gradient_check.*
* `check_result` and `check_close` are now private (but were anyways not very useful by themselves)

`assert_fun` was probably always used to pass `assert`, and stood in the way of providing proper error messages and meaningful stack traces.

### DIFF
--- a/python/amici/gradient_check.py
+++ b/python/amici/gradient_check.py
@@ -14,15 +14,17 @@ import copy
 from typing import Callable, Optional, List, Sequence
 
 
-def check_finite_difference(x0: Sequence[float],
-                            model: Model,
-                            solver: Solver,
-                            edata: ExpData,
-                            ip: int,
-                            fields: List[str],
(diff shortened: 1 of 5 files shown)


python/amici/gradient_check.py
```diff
@@ -14,20 +14,17 @@
...
+        x0: Sequence[float],
+        model: Model,
+        solver: Solver,
+        edata: ExpData,
+        ip: int,
+        fields: List[str],
+        atol: Optional[float] = 1e-4,
+        rtol: Optional[float] = 1e-4,
+        epsilon: Optional[float] = 1e-3
+        ) -> None:
```
FFroehlich, line 27:
```suggestion
) -> None:
```

FFroehlich (approved):
👍

PR #1610, opened 2021-12-06
Refactor amici.ode_export, add AmiciCxxCodePrinter

dweindl:
Move custom codeprinting to a separate code printer class and reuse the same instance for all printing.
Avoids plenty of reinstantiations of the code printer and keeps all settings in one place.

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -31,13 +31,11 @@ from typing import (
     Set, Any
 )
 from string import Template
-from sympy.printing import cxxcode
-from sympy.printing.cxx import _CXXCodePrinterBase
 from sympy.matrices.immutable import ImmutableDenseMatrix
 from sympy.matrices.dense import MutableDenseMatrix
 from sympy.logic.boolalg import BooleanAtom
 from itertools import chain
-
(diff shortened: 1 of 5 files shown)


python/amici/ode_export.py
```diff
@@ -3114,7 +3027,7 @@ def _get_function_body(self,
                             f'reinitialization_state_idxs.cend(), {index}) != '
                             'reinitialization_state_idxs.cend())',
                             f'    {function}[{index}] = '
-                            f'{_print_with_exception(formula)};'
+                            f'{self.model._code_printer.doprint(formula)};'
```
FFroehlich, line 3030: wouldn't it make more sense to have `_code_printer` as member of the exporter?

FFroehlich (approved):
:+1: nice cleanup!

python/amici/ode_export.py
dweindl: Fully agreed. The problem is that ODEModel also generates code (which I think should be changed, but maybe rather separately). Having it in ODEModel for now allows using it also in ODEExporter, but not vice versa.

PR #1615, opened 2021-12-07
Refactor ODE export

dweindl:
* Use class instead of dict for function info, to simplify default argument handling and to find more easily where the respective values are used.

* Allow specifying non-void return type.

* Remove redundant virtual with override

* Cleanup.

Will simplify some changes in #1515


FFroehlich (approved):
👍

PR #1618, opened 2021-12-10
Sphinx doc: Ensure amici import with TYPE_CHECKING==True

dweindl:
Ensure all relevant entities protected by [`if typing.TYPE_CHECKING:`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) are imported. Otherwise sphinx annotation checks will fail.


FFroehlich (approved):
👍

PR #1622, opened 2021-12-15
Add `id` field to ExpData and ReturnData

dweindl:
A ReturnData will have the same id as the ExpData used for simulation. Allows for more informative output and makes it easier to link ReturnDatas to specific simulation conditions.


FFroehlich (approved):
👍  great!

PR #1624, opened 2021-12-16
Fix wrong array size in warnings

dweindl:


FFroehlich:
But indices can't be bigger than `size()-1`.

dweindl:
> But indices can't be bigger than `size()-1`.

"NaN value at index 0/0" looks pretty unintuitive to me... Can we then just drop the size?

FFroehlich:
> > But indices can't be bigger than `size()-1`.
> 
> "NaN value at index 0/0" looks pretty unintuitive to me... Can we then just drop the size?

sounds reasonable, ` {NaN,Inf} value at index %s[%i]`?

PR #1634, opened 2022-01-11
Allowing all integer parameter vector.

stephanmg:
Evaluating an objective function with an all integer parameter vector throws the following error:

`TypeError: in method 'SimulationParameters_parameters_set', argument 2 of type 'std::vector< amici::realtype,std::allocator< amici::realtype > > const &'`

This PR mitigates the requirement for manual creation of a float vector of parameters.


yannikschaelte (approved):
looks reasonable to me

FFroehlich (approved):

PR #1638, opened 2022-01-13
Add functions to provide state IDs/names for x_solver

### LINKED ISSUE #1549: Provide state IDs for x_solver
In case conservation laws have been identified during model import, there is, to my knowledge, currently no way to obtain the state IDs for `x_solver`. This would be nice to have to be able to interpret e.g. `ReturnData::J`. `Model::getStateIds()` provides only the full list of state IDs for `x_rdata`.

dweindl:
Closes #1549


FFroehlich (approved):
👍

PR #1640, opened 2022-01-14
More informative NaN/Inf warnings

### LINKED ISSUE #1375: More informative output when encountering non-finite values
Currently on the index of any values being NaN / Inf is printed. Would be more helpful if the names / IDs were printed in addiiton where available.

dweindl:
Show more informative warnings when encountering `NaN` or `inf` in model quantities (include IDs where available and provide 2D indices instead of flat indices).

Closes #1375


FFroehlich (approved):
👍

PR #1647, opened 2022-02-01
SBML import: Fix special function parsing inside pow()

### LINKED ISSUE #1646: SBML import: special functions inside exp() are not parsed correctly
SBML import: special functions inside pow() are not parsed correctly.

dweindl:
Fixes #1646


FFroehlich (approved):
👍

PR #1648, opened 2022-02-01
Change Model::setStateIsNonNegative logic

dweindl:
Previously `model.setStateIsNonNegative(model.getStateIsNonNegative())`
failed if conservation laws were enabled. That feels unintuitive.
Now it works.

### DIFF
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -609,8 +609,8 @@ std::vector<bool> const &Model::getStateIsNonNegative() const {
 }
 
 void Model::setStateIsNonNegative(std::vector<bool> const &nonNegative) {
-    auto any_state_non_negative = std::any_of(state_is_non_negative_.begin(),
-                                              state_is_non_negative_.end(),
+    auto any_state_non_negative = std::any_of(nonNegative.begin(),
+                                              nonNegative.end(),
                                               [](bool x) { return x; });
 
     if (any_state_non_negative && nx_solver != nx_rdata) {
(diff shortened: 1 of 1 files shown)


src/model.cpp
```diff
@@ -609,7 +609,11 @@ std::vector<bool> const &Model::getStateIsNonNegative() const {
 }
 
 void Model::setStateIsNonNegative(std::vector<bool> const &nonNegative) {
-    if (nx_solver != nx_rdata) {
+    auto any_state_non_negative = std::any_of(state_is_non_negative_.begin(),
```
FFroehlich, line 612: I am not sure what's implemented here, shouldn't this be applied to `nonNegative`? Not implementing this was primarily a result of lazyness. If we specify what kind of indices are passed here (solver or rdata x), we could also support setting non_negative states, just need to map between the two.

dweindl: True. Thanks.

I'll open a separate issue for that.

FFroehlich (approved):

src/model.cpp
dweindl: > If we specify what kind of indices are passed here (solver or rdata x), we could also support setting non_negative states, just need to map between the two.

With x_solver indices we wouldn't have to check whether it actually is a solver state. However, I'd guess that most users wouldn't want to care about x_solver at all, and therefore, x_rdata would be more appropriate. Or we just do it ID-based. That's probably the most convenient. What would be your preference?

FFroehlich: > > If we specify what kind of indices are passed here (solver or rdata x), we could also support setting non_negative states, just need to map between the two.
> 
> With x_solver indices we wouldn't have to check whether it actually is a solver state. However, I'd guess that most users wouldn't want to care about x_solver at all, and therefore, x_rdata would be more appropriate. Or we just do it ID-based. That's probably the most convenient. What would be your preference?

I agree, I would expect that users want to pass x_rdata. For id based specification we should probably add another function as changing this function would be API breaking.

PR #1650, opened 2022-02-02
fix using x_pos in all model functions

FFroehlich:


src/model.cpp
```diff
@@ -1072,29 +1074,31 @@ void Model::getEventTimeSensitivity(std::vector<realtype> &stau,
...
+            state_.unscaledParameters.data(), state_.fixedParameters.data(),
+            state_.h.data(), ie, xdot.data(), xdot_old.data());
 
     if (always_check_finite_) {
         app->checkFinite(derived_state_.deltax_, "deltax");
     }
 
     // update
-    amici_daxpy(nx_solver, 1.0, derived_state_.deltax_.data(), 1, x.data(), 1);
+    amici_daxpy(nx_solver, 1.0, derived_state_.deltax_.data(), 1, x.data(),
```
dweindl, line 1100: For the update, we don't want to use `computeX_pos`?

dweindl (approved):
Great, thanks.

src/model.cpp
FFroehlich: Yes, it think that makes sense!

PR #1654, opened 2022-02-03
Swig docstrings

dweindl:
Simplify sphinx conf.py

### DIFF
--- a/documentation/conf.py
+++ b/documentation/conf.py
@@ -514,23 +514,6 @@ def process_docstring(app, what, name, obj, options, lines):
         )
         return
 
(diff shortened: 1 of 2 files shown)


FFroehlich (approved):
👍

FFroehlich:
Do we still need the `fix_typehints` function?

dweindl:
> Do we still need the `fix_typehints` function?

For this moment, yes. At some point, we should let swig take care of all that. Will still be a bit of work, and there is still some swig issue in the way that should addressed in the next release (currently [scheduled](https://github.com/swig/swig/milestone/3) for end of march).

PR #1655, opened 2022-02-04
Fix special function parsing inside min() and max()

dweindl:
Fixes a bug where special functions occurring inside min() and max() would not be correctly parsed and import would fail.


FFroehlich (approved):
👍

PR #1663, opened 2022-02-07
Fix convergence check in Newton method

plakrisenko:
It seems that the value of `x_newton_` in `SteadystateProblem::applyNewtonsMethod`([applyNewtonsMethod](https://github.com/AMICI-dev/AMICI/blob/82d2cbb9d00ac9c3746b4e85c0fb1d6ebbff2863/src/steadystateproblem.cpp#L490)) doesn't change, but it's used to compute weighted root mean square of `xdot` and, therefore, is used to check convergence. As a result, convergence of Newton's method is achieved only due to `xdot_` becoming sufficiently small. 
For two models that require postequilibration I've observed simulations that finished with `posteq_status = [-4 1 0]`and `posteq_numstep = [0 0 0]`, i.e. Newton's method didn't converge, but the check here https://github.com/AMICI-dev/AMICI/blob/82d2cbb9d00ac9c3746b4e85c0fb1d6ebbff2863/src/steadystateproblem.cpp#L614 returned True and the simulation was considered successful with zero steps. 
Changing `x_newton_` to `x_` fixes the issue.


dweindl:
Looks like this was introduced here: https://github.com/AMICI-dev/AMICI/pull/446

FFroehlich (approved):
Omg, yes! Thanks for figuring that out and fixing it. Solution looks good, but I think we could delete `x_newton_` altogether and replace all occurences with `x_` to avoid any confusion in the future.

dweindl (approved):

PR #1664, opened 2022-02-07
cmake: Add AMICI_CXX_OPTIONS

dweindl:
to set libamici-specific compiler options.


FFroehlich (approved):
what happened to `set(ENABLE_SWIG ON)`?

PR #1666, opened 2022-02-08
add support for pysb local functions

FFroehlich:


dweindl (approved):
Looks good, as far as I can tell.

PR #1672, opened 2022-02-16
Cached simplify in ODE export

dilpath:
Decreases import time of a specific model with events by approximately 2.6x (n=1 ...).

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -959,7 +959,8 @@ class ODEModel:
     """
 
     def __init__(self, verbose: Optional[Union[bool, int]] = False,
-                 simplify: Optional[Callable] = sp.powsimp):
+                 simplify: Optional[Callable] = sp.powsimp,
+                 cache_simplify: bool = True):
         """
         Create a new ODEModel instance.
 
@@ -969,6 +970,10 @@ class ODEModel:
 
         :param simplify:
             see :meth:`ODEModel._simplify`
+
+        :param cache_simplify:
+            Whether to cache calls to the simplify method. Can e.g. decrease
+            import times for models with events.
         """
         self._states: List[State] = []
         self._observables: List[Observable] = []
@@ -1036,8 +1041,8 @@ class ODEModel:
             }
 
         self._lock_total_derivative: List[str] = list()
-        self._simplify: Callable = None
-        if simplify is not None:
+        self._simplify: Callable = simplify
+        if cache_simplify and simplify is not None:
             def cached_simplify(
                 expr: sp.Expr,
                 _simplified: Dict[str, sp.Expr] = {},
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -1051,7 +1051,7 @@ class ODEModel:
                 expr_str = repr(expr)
                 if expr_str not in _simplified:
                     _simplified[expr_str] = _simplify(expr)
-                return copy.deepcopy(_simplified[expr_str])
+                return _simplified[expr_str]
             self._simplify = cached_simplify
         self._x0_fixedParameters_idx: Union[None, Sequence[int]]
         self._w_recursion_depth: int = 0
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -960,7 +960,7 @@ class ODEModel:
 
     def __init__(self, verbose: Optional[Union[bool, int]] = False,
                  simplify: Optional[Callable] = sp.powsimp,
-                 cache_simplify: bool = True):
+                 cache_simplify: bool = False):
         """
         Create a new ODEModel instance.
 
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -1036,7 +1036,18 @@ class ODEModel:
             }
 
         self._lock_total_derivative: List[str] = list()
-        self._simplify: Callable = simplify
+        self._simplify: Callable = None
+        if simplify is not None:
+            def cached_simplify(
+                expr: sp.Expr,
+                _simplified: Dict[str, sp.Expr] = {},
+                _simplify: Callable = simplify,
+            ) -> sp.Expr:
+                expr_str = str(expr)
+                if expr_str not in _simplified:
+                    _simplified[expr_str] = _simplify(expr)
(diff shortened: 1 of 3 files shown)


python/amici/ode_export.py
```diff
@@ -1036,7 +1036,18 @@ def __init__(self, verbose: Optional[Union[bool, int]] = False,
...
+        self._simplify: Callable = None
+        if simplify is not None:
+            def cached_simplify(
+                expr: sp.Expr,
+                _simplified: Dict[str, sp.Expr] = {},
+                _simplify: Callable = simplify,
+            ) -> sp.Expr:
+                expr_str = str(expr)
+                if expr_str not in _simplified:
+                    _simplified[expr_str] = _simplify(expr)
```
dweindl, line 1048:
```suggestion
                    _simplified[id(expr)] = _simplify(expr)
```

Will that work? Would save the time for stringification and some memory...

dweindl: Maybe not. Will fail if a memory address happens to be reused for a different expression. Not sure how often that will happen, but it's potentially unsafe.

python/amici/ode_export.py
```diff
@@ -1036,7 +1036,18 @@ def __init__(self, verbose: Optional[Union[bool, int]] = False,
...
         self._lock_total_derivative: List[str] = list()
-        self._simplify: Callable = simplify
+        self._simplify: Callable = None
+        if simplify is not None:
+            def cached_simplify(
+                expr: sp.Expr,
+                _simplified: Dict[str, sp.Expr] = {},
+                _simplify: Callable = simplify,
+            ) -> sp.Expr:
+                expr_str = str(expr)
```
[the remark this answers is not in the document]
dilpath: Agreed, though in this case it's not so useful since `repr` is `str`.
```python
>>> sp.Expr.__repr__
<function Printable.__str__ at 0x7fe4c85d3af0>
```

python/amici/pysb_import.py
```diff
@@ -188,14 +194,21 @@ def ode_model_from_pysb_importer(
...
 
     :return:
         New ODEModel instance according to pysbModel
     """
 
-    ode = ODEModel(verbose=verbose, simplify=simplify)
+    ode = ODEModel(
+        verbose=verbose,
+        simplify=simplify,
+        cache_simplify=cache_simplify,
```
dweindl, line 210: If it's known to fail for pysb models, I think we shouldn't offer it as an option, but rather set it to `False` and add a comment that it isn't compatible.

dilpath: Hm, now not sure. Just tried to debug the PySB issue by settings the default value in `pysb2amici` and `ode_model_from_pysb_importer` to `True`, but all tests in `pytest test_pysb.py` passed.

dweindl: Also the PEtab pysb test suite? I am worried about the failure in e.g. https://github.com/AMICI-dev/AMICI/runs/5219969341?check_suite_focus=true. If caching silently leads to wrong model equations, this is a severe issue. Unless we have a good explanation of why that happened there and know that it won't happen again, I'd keep it out of there.

dilpath:
After debugging with @dweindl , it seems to work for PySB (the tests in `AMICI/python/tests/test_pysb.py` and `AMICI/tests/petab_test_suite/test_petab_suite.py` pass).

However, it fails for PySB if `deepcopy` is in the cache method, with unknown causes. One reason for failure is due to incorrect gradients; with `deepcopy`, some AMICI C++ files related to gradients are not created. e.g. after running `pytest --only-pysb --petab-cases 1` inside `AMICI/tests/petab_test_suite`, the diff for model files, with and without `deepcopy`, is:
```
Only in nocopy/build/temp.linux-x86_64-3.8: _model_dwdw_colptrs.o
Only in nocopy/build/temp.linux-x86_64-3.8: _model_dwdw.o
Only in nocopy/build/temp.linux-x86_64-3.8: _model_dwdw_rowvals.o
Only in nocopy/build/temp.linux-x86_64-3.8: _model_dxdotdp_explicit_colptrs.o
Only in nocopy/build/temp.linux-x86_64-3.8: _model_dxdotdp_explicit.o
Only in nocopy/build/temp.linux-x86_64-3.8: _model_dxdotdp_explicit_rowvals.o
Only in nocopy/build/temp.linux-x86_64-3.8: _model_dydx.o
Only in nocopy/build/temp.linux-x86_64-3.8: _model_sx0.o
Files copy/build/temp.linux-x86_64-3.8/wrapfunctions.o and nocopy/build/temp.linux-x86_64-3.8/wrapfunctions.o differ
Files copy/CMakeLists.txt and nocopy/CMakeLists.txt differ
Files copy/_model/__model.cpython-38-x86_64-linux-gnu.so and nocopy/_model/__model.cpython-38-x86_64-linux-gnu.so differ
Files copy/_model/__pycache__/__init__.cpython-38.pyc and nocopy/_model/__pycache__/__init__.cpython-38.pyc differ
Only in nocopy: _model_dwdw_colptrs.cpp
Only in nocopy: _model_dwdw.cpp
Only in nocopy: _model_dwdw.h
Only in nocopy: _model_dwdw_rowvals.cpp
Only in nocopy: _model_dxdotdp_explicit_colptrs.cpp
Only in nocopy: _model_dxdotdp_explicit.cpp
Only in nocopy: _model_dxdotdp_explicit.h
Only in nocopy: _model_dxdotdp_explicit_rowvals.cpp
Only in nocopy: _model_dydx.cpp
Files copy/_model.h and nocopy/_model.h differ
Only in nocopy: _model_sx0.cpp
```

One possible cause of failure might be related to the PySB objects that exist inside `ODEModel` and are sent to `cached_simplify`, but this method expects SymPy objects.

Hence, this feature is turned off by default, for PySB models.

For SBML models, this caching can increase import times. For example, for a large model [1], this increased import times by approx. 10 mins, from 36 mins to 46 mins (some variability). However, no significant difference in RAM usage was seen. As the benefit of the caching appears to be model-specific, it is turned off by default for SBML models as well.

[1] https://github.com/ICB-DCM/CS_Signalling_ERBB_RAS_AKT/tree/master/FroehlichKes2018/PEtab

PR #1678, opened 2022-02-17
Support for parameter-dependent conservation laws

### LINKED ISSUE #1673: Support for parameter-dependent conservation laws
So far, conservation laws can only be state-dependent. It would be great if they could also be parameter-dependent to cope, for example, with parameterized compartment sizes.

dweindl:
Adds support for parameter-dependent conservation laws. This allows applying conservation laws to models comprising species  in different compartments where model states correspond to concentrations. 

Closes #1673

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -3120,7 +3120,8 @@ class ODEExporter:
                 outer_cases[ie] = copy.copy(inner_lines)
             lines.extend(get_switch_statement('ie', outer_cases, 1))
 
-        elif function in sensi_functions:
+        elif function in sensi_functions \
+                and equations.shape[1] == self.model.num_par():
             cases = {
                 ipar: self.model._code_printer._get_sym_lines_array(
                     equations[:, ipar], function, 0)
@@ -3128,7 +3129,6 @@ class ODEExporter:
                 if not smart_is_zero_matrix(equations[:, ipar])
             }
             lines.extend(get_switch_statement('ip', cases, 1))
-
         elif function in multiobs_functions:
             if function == 'dJydy':
                 cases = {
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -166,7 +166,8 @@ functions = {
         _FunctionInfo(
             'realtype *dydp, const realtype t, const realtype *x, '
             'const realtype *p, const realtype *k, const realtype *h, '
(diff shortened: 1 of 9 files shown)


python/amici/ode_export.py
```diff
@@ -1854,8 +1870,39 @@ def _compute_equation(self, name: str) -> None:
...
 
         elif name == 'dtcldp':
-            # force symbols
-            self._eqs[name] = self.sym(name)
+            self._derivative('total_cl', 'p', name=name)
+
+        elif name == 'stotal_cl':
+            # stotal_cl = dtotal_cl/dp +  dtotal_cl/dx_rdata * sx_rdata
+            # shape: ncl x np
+            dtotal_cldp = smart_jacobian(self.eq('total_cl'), self.sym('p'))
```
FFroehlich, line 1878:
```suggestion
            self._eqs[name] = self.eq('tcldp')
```

python/amici/ode_export.py
```diff
@@ -1854,8 +1870,39 @@ def _compute_equation(self, name: str) -> None:
...
+            dtotal_cldx_rdata = smart_jacobian(self.eq('total_cl'),
+                                               self.sym('x_rdata'))
+            self._eqs[name] = dtotal_cldp
```
FFroehlich, line 1881:
```suggestion
```

python/amici/ode_export.py
```diff
@@ -1854,8 +1870,39 @@ def _compute_equation(self, name: str) -> None:
...
+            tmp = smart_multiply(dtotal_cldx_rdata, self.sym('sx_rdata'))
+            for ip in range(self._eqs[name].shape[1]):
+                self._eqs[name][:, ip] += tmp
+        elif name == 'sx_rdata':
+            # sx_rdata =  dx_rdata/dx_solver * sx_solver
+            #              + dx_rdata/d_tcl * stcl + dxrdata/dp
+            # shape: nx_rdata, 1
+            dx_rdata_dx_solver = smart_jacobian(self.eq('x_rdata'),
```
FFroehlich, line 1889: if we don't have any cls, we can simply `self._eqs[name] = self.sym('sx_solver')`, will probably speed things up quite a bit for larger models.

FFroehlich (approved):
👍

python/amici/ode_export.py
dweindl: Thanks, makes sense

PR #1683, opened 2022-02-24
Sparsify conserved quantities to avoid circular dependencies

### LINKED ISSUE #1682: Improve handling of circular dependencies in conservation laws
Currently, in case a circular dependency is detected, no CLs for non-constant species are applied. This should be changed to applying all but those giving rise to circular dependency.

dweindl:
For SBML models: Sparsify conserved quantities to avoid circular dependencies among state expressions.

Closes #1682

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -1449,7 +1449,9 @@ class SbmlImporter:
                 float(entry) for entry in self.stoichiometric_matrix.T.flat()
             ]
         except TypeError:
-            # https://github.com/AMICI-dev/AMICI/issues/1673
+            # Due to the numerical algorithm currently used to identify
+            #  conserved quantities, we can't have symbols in the
+            #  stoichiometric matrix
             warnings.warn("Conservation laws for non-constant species in "
                           "combination with parameterized stoichiometric "
                           "coefficients are not currently supported "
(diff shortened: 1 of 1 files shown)


python/amici/sbml_import.py
```diff
@@ -1468,48 +1470,46 @@ def _add_conservation_for_non_constant_species(
...
             rng_seed=32)
 
+        # sparsify conserved quantities
+        # construct and reduce A * x0 = total_cl
+        A = sp.zeros(len(cls_coefficients), len(ode_model._states))
+        for i_cl, (cl, coefficients) in enumerate(zip(cls_state_idxs,
+                                                      cls_coefficients)):
+            for i, c in zip(cl, coefficients):
+                A[i_cl, i] = sp.Rational(c)
+        rref, pivots = A.rref()
```
FFroehlich, line 1480: Brilliant!

Maybe add a comment that this is reduced row echelon form and why this makes sense?

I am also wondering whether we still need the whole monte carlo mumbo jumbo?

python/amici/sbml_import.py
```diff
@@ -1539,7 +1539,7 @@ def _add_conservation_for_non_constant_species(
         try:
             sorted_state_exprs = toposort_symbols(state_exprs)
         except CircularDependencyError as e:
-            # see SBML semantic test suite, case 18 for an example
+            # This should never happen
```
FFroehlich, line 1542: if this should never happen maybe raise an error instead of logging a warning?

FFroehlich (approved):

python/amici/sbml_import.py
dweindl: Thanks. Will do. My feeling is also that we could skip at least certain parts of the algorithm, although I am not sure what exactly. This should be rewritten cleanly at some point. Would leave it for now, though.

python/amici/sbml_import.py
dweindl: Makes sense. Will re-raise.

PR #1685, opened 2022-02-24
Fix IndexError conserved_moieties._reduce

dweindl:
Fixes `IndexError` occurring under rare circumstances:

```
  File "/venv/src/amici/python/sdist/amici/conserved_moieties.py", line 698, in _relax
    if not len(matrix[j]) or matrix[j][0] != matrix[k][i]:
IndexError: list index out of range
```

### DIFF
--- a/python/amici/conserved_moieties.py
+++ b/python/amici/conserved_moieties.py
@@ -695,10 +695,11 @@ def _relax(
         while i < len(matrix[k]):
             for i_species in range(k1 + 1, K):
                 j = order[i_species]
-                if not len(matrix[j]) or len(matrix[k]) <= i \
-                        or matrix[j][0] != matrix[k][i]:
+                if not len(matrix[j]) or matrix[j][0] != matrix[k][i]:
                     continue
 
+                # subtract rows
+                # matrix2[k] = matrix2[k] - matrix2[j] * matrix2[k][i]
                 row_k: List[float] = [0] * num_reactions
                 for a in range(len(matrix[k])):
                     row_k[matrix[k][a]] = matrix2[k][a]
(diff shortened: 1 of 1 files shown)


python/amici/conserved_moieties.py
```diff
@@ -695,7 +695,8 @@ def _relax(
         while i < len(matrix[k]):
             for i_species in range(k1 + 1, K):
                 j = order[i_species]
-                if not len(matrix[j]) or matrix[j][0] != matrix[k][i]:
+                if not len(matrix[j]) or len(matrix[k]) <= i \
```
FFroehlich, line 698: Why is this not covered by `i < len(matrix[k])`?

I would guess that 

```
[row_idx for row_idx, row_val in enumerate(row_k)
                             if row_val != 0]
```

generates a too short matrix, but why? This smells like a more serious bug in the algorithm.

dweindl: I agree. I think the issue is that the size may change in the loop after that initial check. The same issue exists in the original implementation. I am not sure how to address that better.

stephanmg: Unluckily there is extremely scarce commenting in the original implementation, mostly in the form of re-stating the method name, etc., so I would also tend to say this is a bug in the original implementation until further notice.

dweindl: My feeling is that it's okay, that `len(matrix[k])` is decreasing across iterations, i.e., that `matrix2[k] -= matrix2[j] * matrix2[k][i]` introduces zeros further to the right. The problem rarely occurs, and I guess that it just never happened for the test cases of the original implementation, and therefore, not adequately handled.

FFroehlich: ah, that makes sense!

FFroehlich (approved):

PR #1687, opened 2022-02-28
Speed up model import

dweindl:


dweindl:
Partially continued in #1722 . Simultaneous substitutions turned out to be slower. Closing.

PR #1690, opened 2022-02-28
SBML import: Raise in case of nested observables

### LINKED ISSUE #1689: Observable-dependent observables
Observable-dependent observables are not handled correctly and don't raise an error either. Needs to be addressed one way or another.

dweindl:
Closes #1689


FFroehlich (approved):
👍

PR #1692, opened 2022-03-01
Add support for observable-dependent sigmas

### LINKED ISSUE #609: Support of observable-dependent sigmas
Currently it is not possibly to have state-dependent sigmas. Is it only about changing respective function signatures, or will that require further adaptations?

### LINKED ISSUE #1689: Observable-dependent observables

dweindl:
Standard deviations may now depend on observables.

Side-effects:
* `ssigmay` will not be available anymore with adjoint sensitivity analysis.
* Observable IDs shadowing other model entities are no longer allowed

Closes #609


src/model.cpp
```diff
@@ -1524,6 +1584,28 @@ void Model::fdJydy(const int it, const AmiVector &x, const ExpData &edata) {
...
+                        &derived_state_.dJydsigma_.at(iyt * nJ * ny), nJ,
+                        derived_state_.dsigmaydy_.data(), ny, 1.0,
+                        tmp_dense.data(), nJ);
+
+            auto tmp_sparse = SUNSparseFromDenseMatrix(
+                tmp_dense.get(), 0.0, CSC_MAT);
+            auto ret = SUNMatScaleAdd(1.0, derived_state_.dJydy_.at(iyt).get(),
+                                      tmp_sparse);
+            if(ret != SUNMAT_SUCCESS) {
+                SUNMatDestroy(tmp_sparse);
```
FFroehlich, line 1602: move `SUNMatDestroy(tmp_sparse);` before if clause?

src/rdata.cpp
```diff
@@ -142,7 +142,6 @@ void ReturnData::initializeFullReporting(bool quadratic_llh) {
             srz.resize(nmaxevent * nz * nplist, 0.0);
         }
 
-        ssigmay.resize(nt * ny * nplist, 0.0);
```
FFroehlich, line 145: don't delete but only initialize if `sensi_meth == SensitivityMethod::forward`?

FFroehlich (approved):
Nice! Should be possible to add Raia model (https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/tree/model_Raia_CancerResearch2011) to bmc tests now, right?

src/rdata.cpp
dweindl: That's already taken care of by `ReturnData::initializeResidualReporting` which is called a few lines above.
https://github.com/AMICI-dev/AMICI/blob/353e7f9106c6592921f191a0bc816280afdff3c0/src/rdata.cpp#L75-L86

src/model.cpp
dweindl: Absolutely.

dweindl:
> Nice! Should be possible to add Raia model (https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/tree/model_Raia_CancerResearch2011) to bmc tests now, right?

Thanks for your help.

`Raia_CancerResearch2011` can now be imported. However, the computed likelihood is very different from https://github.com/AMICI-dev/AMICI/blob/067826d40be0704c311c345ab5cd199f8548bc1b/tests/benchmark-models/benchmark_models.yaml#L58-L59.

PR #1694, opened 2022-03-07
Sympy 1.10 compatibility

dweindl:
sympify now automatically parses min/max, thereby escaping our argument conversion for float.
Fixed here by moving the conversion from import to code generation.

### DIFF
--- a/python/amici/cxxcodeprinter.py
+++ b/python/amici/cxxcodeprinter.py
@@ -20,9 +20,28 @@ class AmiciCxxCodePrinter(CXX11CodePrinter):
             return code
         except TypeError as e:
             raise ValueError(
-                f'Encountered unsupported function in expression "{expr}": '
-                f'{e}!'
-            )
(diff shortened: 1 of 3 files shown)


python/amici/cxxcodeprinter.py
```diff
@@ -20,9 +20,28 @@ def doprint(self, expr: sp.Expr, assign_to: Optional[str] = None) -> str:
...
         except TypeError as e:
             raise ValueError(
-                f'Encountered unsupported function in expression "{expr}": '
-                f'{e}!'
-            )
+                f'Encountered unsupported function in expression "{expr}"'
+            ) from e
+
+    def _print_Max(self, expr):
+        from sympy.functions.elementary.miscellaneous import Max
```
FFroehlich, line 27: implement this as a generic function `_print_MinMax` and implement min/max as specialization `_print_MinMax(self, expr, {'min', 'max'})?

FFroehlich (approved):

Issue #1119, opened 2020-06-05
Make models independent of AMICI base package


dweindl:
Currently, running model simulations requires the model package and the matching AMICI base package. It would be convenient, if the model package was independent.

This would simplify comparing models generated from different AMICI versions and would allow simple sharing of model packages independently.

Disadvantage would be, that generated modules and source distributions would be larger.

dweindl:
Closing, is this is not possible without a complete API change. The costs would significantly outweigh the benefits.

PR #1696, opened 2022-03-09
Refactor: Move swig wrappers out of __init__.py

### LINKED ISSUE #1215: Move convenience swig wrappers from __init__.py to separate module
Still import them to `amici`, but only if the extension was built. Avoids undefined symbols in case the extension wasn't built.

dweindl:
Closes #1215.

Will also make it easier to reuse those wrappers for #1119


FFroehlich (approved):
👍

PR #1710, opened 2022-03-11
Don't compute sx_rdata symbolically

dweindl:


src/abstract_model.cpp
```diff
@@ -654,4 +654,32 @@ void AbstractModel::fdwdw_rowvals(SUNMatrixWrapper &/*dwdw*/) {
                        __func__);
 }
 
+void AbstractModel::fdx_rdatadx_solver(realtype */*dx_rdatadx_solver*/,
+                               const realtype */*x*/, const realtype */*tcl*/,
+                               const realtype */*p*/, const realtype */*k*/)
+{
+    throw AmiException("Requested functionality is not supported as %s is "
```
FFroehlich, line 661: only throw if nx_solver != nx_rdata ?

dweindl: At the moment this is checked in `fsx_rdata`. If `nx_solver == nx_rdata`, this will never be called. I am not sure where else this might be required, so I thought it might be better to just throw.

src/model.cpp
```diff
@@ -2183,15 +2183,42 @@ void Model::fx_rdata(realtype *x_rdata, const realtype *x_solver,
...
+    if (nx_solver == nx_rdata) {
+        std::copy_n(sx_solver, nx_solver, sx_rdata);
+        return;
+    }
+
+    // sx_rdata =  dx_rdata/dx_solver * sx_solver
+    //             + dx_rdata/d_tcl * stcl + dxrdata/dp
+
+    // 1) sx_rdata(nx_rdata, 1) = dx_rdatadp
+    std::fill(sx_rdata, sx_rdata + nx_rdata, 0.0);
```
FFroehlich, line 2199:
```suggestion
    std::fill_n(sx_rdata, nx_rdata, 0.0);
```

PR #1711, opened 2022-03-11
Fix documentation for sz, srz, ssigmaz dimensions

dweindl:
Incorrect dimensions were stated in the docs as pointed out by @dilpath


FFroehlich (commented):
👍

PR #1716, opened 2022-03-11
CI: Fix C++ tests under valgrind

### LINKED ISSUE #1715: Running C++ tests under valgrind fails silently
e.g. [here](https://github.com/AMICI-dev/AMICI/runs/5127016370?check_suite_focus=true)

dweindl:
Previously, tests were not found and silently skipped. Probably since changing to gtest. Fixes #1715.


FFroehlich (approved):
🙈

PR #1722, opened 2022-03-14
speed: equation-wise check for disjoint free symbols in `smart_jacobian`

dweindl:
Speeds up model import for larger models:

With this change:
        Elapsed (wall clock) time (h:mm:ss or m:ss): 28:27.82
Before:
        Elapsed (wall clock) time (h:mm:ss or m:ss): 32:21.23


FFroehlich (approved):
👍

PR #1726, opened 2022-03-15
Make model packages independent of amici package

### LINKED ISSUE #1119: Make models independent of AMICI base package

dweindl:
Closes #1119


dweindl:
Not going to happen anytime soon. Various issues: This would completely change the API. A smooth transition to the new API would be difficult to implement, as various classes would have to be defined in different places (same name, but different objects). Also, having the models independent of amici-base would complicate/break the current code coverage analysis.

PR #1727, opened 2022-03-16
Steady state sensitivity modes

plakrisenko:


tests/petab_test_suite/test_petab_suite.py
```diff
@@ -134,7 +134,7 @@ def check_derivatives(problem: petab.Problem, model: amici.Model) -> None:
     # Required for case 9 to not fail in
     #  amici::NewtonSolver::computeNewtonSensis
     model.setSteadyStateSensitivityMode(
-        SteadyStateSensitivityMode_simulationFSA)
+        SteadyStateSensitivityMode_integrateIfNewtonFails)
```
FFroehlich, line 137:
```suggestion
        SteadyStateSensitivityMode.integrateIfNewtonFails)
```

src/steadystateproblem.cpp
```diff
@@ -49,6 +49,14 @@ SteadystateProblem::SteadystateProblem(const Solver &solver, Model &model)
         throw AmiException("Preequilibration using adjoint sensitivities "
                            "is not compatible with using forward "
                            "sensitivities during simulation");
+
+    /* Turn off Newton's method if SteadyStateSensitivityMode is integrationOnly 
+    in forward sensitivity case */
+    if (solver.getSensitivityMethod() == SensitivityMethod::forward &&
```
FFroehlich, line 55: I would prefer if this was rather implemented as a check in `findSteadyState`. On the one hand this means we do not apply any changes to the solver that the user may not expect (changing newton steps means switching to a different sensitivity mode afterwards wouldn't necessarily reactivate it). On the other hand it avoids issues in edge cases when convergence is achieved in the first newton solver step, which is carried out even when newton steps are set to 0.

FFroehlich (changes requested):

src/steadystateproblem.cpp
plakrisenko: Ah, I see. Wasn't sure about this part. Thanks for the explanation

FFroehlich (approved):
not sure why the last remaining test fails, don't see anything that could lead to test failure.

src/steadystateproblem.cpp
```diff
@@ -90,16 +90,21 @@ void SteadystateProblem::workSteadyStateBackwardProblem(
 void SteadystateProblem::findSteadyState(const Solver *solver, Model *model,
                                          int it) {
     steady_state_status_.resize(3, SteadyStateStatus::not_run);
+    bool turnOffNewton = solver->getSensitivityMethod() ==
```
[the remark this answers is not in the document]
plakrisenko: Because in adjoint case steady-state computation part is independent of sensitivities computation.
If `model->getSteadyStateSensitivityMode() == SteadyStateSensitivityMode::integrationOnly` in adjoint case, I think model steady-state could still be found by Newton method.

include/amici/defines.h
```diff
@@ -180,7 +180,8 @@ enum class NonlinearSolverIteration {
 /** Sensitivity computation mode in steadyStateProblem */
 enum class SteadyStateSensitivityMode {
```
FFroehlich, line 181: I think we need some documentation of the behavior for the different values, for example in the function `Model::setSteadyStateSensitivityMode()`. https://github.com/AMICI-dev/AMICI/blob/master/python/examples/example_constant_species/ExampleEquilibrationLogic.ipynb also needs to be updated.

src/steadystateproblem.cpp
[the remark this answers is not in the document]
FFroehlich: but that also excludes `solver->getSensitivityMethod()==SensitivityMethod::none` or situations where `solver->getSensitivityOrder() < SensitivityOrder::first` where anyways no sensitivities are computed. It also ignores `solver->getSensitivityMethodPreequilibration()`.

 I think the best way is to use SteadystateProblem::getSensitivityFlag here.

FFroehlich (changes requested):
Sorry, noticed a couple more things that need to be adressed :(

include/amici/defines.h
plakrisenko: Yeah, this notebook definitely  needs to be updated.

plakrisenko:
> Sorry, noticed a couple more things that need to be adressed :(

You are right about the preequilibration. 
I guess the logic should be rather 
`    bool turnOffNewton = model->getSteadyStateSensitivityMode() ==
        SteadyStateSensitivityMode::integrationOnly && 
        (it == -1 && solver->getSensitivityMethodPreequilibration() ==
         SensitivityMethod::forward || solver->getSensitivityMethod() ==
        SensitivityMethod::forward);
`
Right?

src/steadystateproblem.cpp
```diff
@@ -360,8 +371,10 @@ bool SteadystateProblem::getSensitivityFlag(const Model *model,
...
         solver->getSensitivityOrder() >= SensitivityOrder::first &&
         steady_state_status_[1] == SteadyStateStatus::success &&
-        model->getSteadyStateSensitivityMode() ==
-            SteadyStateSensitivityMode::simulationFSA;
+        (model->getSteadyStateSensitivityMode() == 
+         SteadyStateSensitivityMode::integrationOnly || 
+         model->getSteadyStateSensitivityMode() == 
+         SteadyStateSensitivityMode::integrateIfNewtonFails);
 
     bool simulationStartedInSteadystate =
```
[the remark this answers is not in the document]
FFroehlich: For models that start in steady-state (e.g., computed outside of amici or extracted from other simulations), we don't want to be forced to deactivate the newton solver. See #1422.

FFroehlich:
> > Sorry, noticed a couple more things that need to be adressed :(
> 
> You are right about the preequilibration. I guess the logic should be rather ` bool turnOffNewton = model->getSteadyStateSensitivityMode() == SteadyStateSensitivityMode::integrationOnly && (it == -1 && solver->getSensitivityMethodPreequilibration() == SensitivityMethod::forward || solver->getSensitivityMethod() == SensitivityMethod::forward);` Right?



> > Sorry, noticed a couple more things that need to be adressed :(
> 
> You are right about the preequilibration. I guess the logic should be rather ` bool turnOffNewton = model->getSteadyStateSensitivityMode() == SteadyStateSensitivityMode::integrationOnly && (it == -1 && solver->getSensitivityMethodPreequilibration() == SensitivityMethod::forward || solver->getSensitivityMethod() == SensitivityMethod::forward);` Right?

Yes, that looks correct.

FFroehlich (approved):
👍

PR #1729, opened 2022-03-16
remove SPBCG solver

FFroehlich:
remove SPBCG solver due to divergence from remaining codebase and lack of willing individuals to maintain code.


dweindl (approved):
:wave:

PR #1731, opened 2022-03-16
harmonize convergence checks

FFroehlich:
use the same routines for newton and steadystate, newton still doesn't check sensitivity convergence


src/steadystateproblem.cpp
```diff
@@ -16,6 +16,8 @@
 #include <memory>
 #include <cvodes/cvodes.h>
 
+#define CONV_T = RCONST(1.0)
```
dweindl:
```suggestion
constexpr realtype convergence_threshold = 1.0;
```
?

dweindl (approved):
:+1:

src/steadystateproblem.cpp
```diff
@@ -433,41 +441,36 @@ realtype SteadystateProblem::getWrmsNorm(const AmiVector &x,
...
-                solver->getRelativeToleranceSteadyStateSensi(), ewt_);
-            converged = wrms_ < RCONST(1.0);
-        }
+        if (wrms >= conv_thresh)
+            break;
+        model->fsxdot(t_, x_, dx_, ip, sx_[ip], dx_, xdot_);
+        wrms = getWrmsNorm(sx_[ip], xdot_, atol_sensi_, rtol_sensi_, ewt_);
     }
-    return converged;
+    return wrms;
```
plakrisenko, line 473: wrms for the last parameter is returned?

FFroehlich: > wrms for the last parameter is returned?

for the current `x_`, yes.

src/steadystateproblem.cpp
```diff
@@ -433,41 +441,36 @@ realtype SteadystateProblem::getWrmsNorm(const AmiVector &x,
...
     for (int ip = 0; ip < model->nplist(); ++ip) {
-        if (converged) {
-            sx_ = solver->getStateSensitivity(t_);
-            model->fsxdot(t_, x_, dx_, ip, sx_[ip], dx_, xdot_);
-            wrms_ = getWrmsNorm(
-                sx_[ip], xdot_, solver->getAbsoluteToleranceSteadyStateSensi(),
-                solver->getRelativeToleranceSteadyStateSensi(), ewt_);
-            converged = wrms_ < RCONST(1.0);
-        }
+        if (wrms >= conv_thresh)
```
plakrisenko, line 468: wouldn't it break at the first step because INFINITY >= 1?

FFroehlich: yes, good point!

src/steadystateproblem.cpp
plakrisenko: shouldn't it be an aggregation of all nplist `wrms`es?

FFroehlich: ah, now I understand the question. We want the maximum of the `wrms` over all `ip`, but if any of them is larger than the convergence threshold, we can stop as we know we aren't converged yet. Will fix and document.

plakrisenko (approved):

Issue #1649, opened 2022-02-01
Enable Model::setStateIsNonNegative for models with conservation laws


dweindl:
Not supported so far. See https://github.com/AMICI-dev/AMICI/pull/1648#discussion_r797021260.

FFroehlich:
Supported as far as possible, actual fix for a seperate problem in https://github.com/AMICI-dev/AMICI/commit/3d4975cd7fbffd9be588bc8b8f1e925590b3931f

PR #1734, opened 2022-03-17
Skip setting state-is-nonnegative in set_model_settings for models wi…

dweindl:
…th conservation laws

Until #1649 is addressed in some other way.

### DIFF
--- a/python/amici/swig_wrappers.py
+++ b/python/amici/swig_wrappers.py
@@ -222,5 +222,10 @@ def set_model_settings(
         values are provided to the setters.
     """
     for setting, value in settings.items():
+        if setting == 'StateIsNonNegative' and model.ncl():
+            # In case of conservation laws, settings StateIsNonNegative is not
+            #  supported, and would raise an error. We can just skip it here.
+            # tracked here: https://github.com/AMICI-dev/AMICI/issues/1649
+            continue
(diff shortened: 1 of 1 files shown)


dilpath (approved):
Could add a warning, in case a user tries to manually set it with this method? Then there may be one warning per start of a multi-start optimization in pyPESTO.

dweindl:
> Could add a warning, in case a user tries to manually set it with this method? Then there may be one warning per start of a multi-start optimization in pyPESTO.

I think that would rather confuse people. I don't think the method is really convenient to use in any scenario other than `get_model_settings` -> `set_model_settings`.

FFroehlich (commented):
Why is this not handled by
```
    auto any_state_non_negative = std::any_of(nonNegative.begin(),
                                              nonNegative.end(),
                                              [](bool x) { return x; });
```
? If there are conservation laws, shouldn't all the stored values be set to False?

dweindl:
> Why is this not handled by
> 
> ```
>     auto any_state_non_negative = std::any_of(nonNegative.begin(),
>                                               nonNegative.end(),
>                                               [](bool x) { return x; });
> ```
> 
> ? If there are conservation laws, shouldn't all the stored values be set to False?

Good question.

dweindl:
Better addressed by #1738

PR #1736, opened 2022-03-18
Fix Model::setStateIsNonNegative

dweindl:
Fixes a logic error when handling all-false input for models with conservation laws.
And adds at test.

### DIFF
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -622,7 +622,7 @@ void Model::setStateIsNonNegative(std::vector<bool> const &nonNegative) {
     if (nx_solver != nx_rdata) {
         if(any_state_non_negative)
             throw AmiException("Non-negative states are not supported with"
-                               " conservation laws enabled");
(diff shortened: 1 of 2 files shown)


src/model.cpp
```diff
@@ -617,11 +617,15 @@ std::vector<bool> const &Model::getStateIsNonNegative() const {
...
-
-    if (any_state_non_negative && nx_solver != nx_rdata) {
-        throw AmiException("Non-negative states are not supported with"
-                           " conservation laws enabled");
+                                              [](bool x) {
+                                                  return x; });
+    if (nx_solver != nx_rdata) {
+        if(any_state_non_negative)
+            throw AmiException("Non-negative states are not supported with"
+                               " conservation laws enabled");
```
FFroehlich, line 625:
```suggestion
                               " conservation laws enabled.");
```

FFroehlich (approved):

PR #1738, opened 2022-03-23
Fix undefined reference to dladdr

kristianmeyerr:
The explicit linking to libdl can avoid undefined reference to dladdr on some ubuntu systems.


dweindl:
Thanks @kristianmeyerr 

For future reference: This is required if AMICI is to be used without HDF5. When using HDF5, libdl comes in as transitive dependency so that this issue never surfaced here.

dweindl (approved):

PR #1744, opened 2022-03-25
Remove legacy options/members `amioption.newton_preeq` and `Solver::r…

### LINKED ISSUE #1148: Remove `newton_preeq` (Matlab) and `requires_preequilibration` (C++, Python)
The two flags `newton_preeq` (Matlab) and `requires_preequilibration` (in solver class, C++, Python) don't do anything useful right now and are pure legacy stuff. They keep confusing different users (including me), how they should be set.
At least to what I understand, the question whether preequilibration is done or not is answered based on whether an ExpData object was passed to AMICI whether this object has a nonempty `preequilibrationCondition`.
At least, that's what should decide...

Unless someone else says/knows what these flags would be good for at the moment or any later point, I would suggest to remove them.

dweindl:
…equires_preequilibration_`

Removes `amioption.newton_preeq` (Matlab) and `Solver::requires_preequilibration_`, `Solver::{get|set}Preequilibration` (C++, Python), that became obsolete. See discussion in https://github.com/AMICI-dev/AMICI/issues/1148.

Closes #1148.


FFroehlich (commented):
👍

PR #1746, opened 2022-03-25
speedup edata construction from petab problems

FFroehlich:


FFroehlich:
tested with 20k conditions.

Not sure, might be better to rather implement this as part libpetab API.

dweindl (approved):
:+1: 

Agreed, could also be included in the PEtab library, but fine for now.

PR #1747, opened 2022-03-28
Non const cls option for mc

stephanmg:


dweindl:
Hi @stephanmg, this just ignores all conserved quantities that are not already found inside `_kernel`, I don't think this is what we want.

PR #1748, opened 2022-03-28
SBML import: Alternative algorithm for identifying conservation laws

dweindl:
... without Monte Carlo mumbo-jumbo.

Compute left nullspace of S via (numerical) Gaussian elimination. Going via SVD would be faster and numerically safer, but this seems to lead to weird coefficients. 

For the model in `test_compute_moiety_conservation_laws_demartino2014`, this algorithm is slower than the one already implemented (~1min vs ~10s), but at least it's deterministic and not highly dependent on RNG seeds. For another model of similar size, the new algorithm takes ~2.5min instead of previously 1h20.


python/amici/sbml_import.py
```diff
@@ -1441,22 +1440,21 @@ def process_conservation_laws(self, ode_model) -> None:
...
-        self,
-        ode_model: ODEModel,
-        conservation_laws: List[ConservationLaw]
-    ) -> List[int]:
-        """Add non-constant species to conservation laws
+    def _get_conservation_laws_demartino(
+            self,
+            ode_model: ODEModel,
+    ) -> List[Tuple[int, List[int], List[float]]]:
+        """Identify conservation laws based on algorithm by DeMartiono et al.
```
dweindl, line 1447:
```suggestion
        """Identify conservation laws based on algorithm by DeMartino et al.
```

python/amici/sbml_import.py
```diff
@@ -1501,7 +1499,105 @@ def _add_conservation_for_non_constant_species(
...
+            List of species indices which later remain in the ODE solver
+        """
+        # indices of retained species
+        species_solver = list(range(ode_model.num_states_rdata()))
+
+        algorithm = os.environ.get("AMICI_EXPERIMENTAL_SBML_NONCONST_CLS", "")
+        if algorithm.lower() == "demartino":
+            raw_cls = self._get_conservation_laws_demartino(ode_model)
+        else:
+            raw_cls = self._get_conservation_laws_new()
```
dweindl, line 1594: Ideas for better algorithm names? 

I'd still keep the previous algorithm as an option, at least for another while.

python/amici/sbml_import.py
```diff
@@ -1501,7 +1499,105 @@ def _add_conservation_for_non_constant_species(
...
+        for i_cl, target_state_model_idx in enumerate(pivots):
+            # collect values for species engaged in the current CL
+            state_idxs = [i for i, coeff in enumerate(rref[i_cl, :])
+                          if coeff]
+            coefficients = [coeff for coeff in rref[i_cl, :] if coeff]
+            raw_cls.append((target_state_model_idx, state_idxs,
+                            coefficients),)
+        return raw_cls
+
+    def _get_conservation_laws_new(
```
FFroehlich, line 1513:
```suggestion
    def _get_conservation_laws_rref(
```

FFroehlich (approved):

python/amici/sbml_import.py
dweindl: Might be somewhat confusing, as the other algorithm is also doing that as a first step, but I don't have a better idea :sweat_smile:

PR #1751, opened 2022-03-30
Fix get_model_settings for initial states + sensitivities

dweindl:
Fixes a couple of issues in `amici.{get,set}_model_settings`

* Initial states and initial state sensitivities should only be included if custom values have been set. Otherwise, `set_model_settings(model, get_model_settings(model))` will permanently set the given values, preventing evaluation of `Model::fx0` / `Model::fsx0`.

  Fixes https://github.com/ICB-DCM/pyPESTO/issues/837

* If a model had custom initial state sensitivities set, those would not have been set correctly by `set_model_settings(model, get_model_settings(model))`, because `setParameter{List,Scale}` was called after `setInitialStateSensitivities` and cleared initial state sensitivities. (Even if they would have been set, they would have been wrong in case of scaled parameters, due to using `setInitialStateSensitivities` instead of `setUnscaledInitialStateSensitivities`).

### DIFF
--- a/python/amici/swig_wrappers.py
+++ b/python/amici/swig_wrappers.py
@@ -172,6 +172,10 @@ def writeSolverSettingsToHDF5(
 # is a tuple where the first and second elements are the getter and setter
 # methods, respectively.
 model_instance_settings = [
+    # `setParameter{List,Scale}` will clear initial state sensitivities, so
+    #  `setParameter{List,Scale}` has to be called first.
+    'ParameterList',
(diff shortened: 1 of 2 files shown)


dilpath (approved):
:+1:

Thanks a lot!

FFroehlich (approved):
makes sense, do we want to add a test for this?

dweindl:
> makes sense, do we want to add a test for this?

Yes. Helped to discover another 2 old issues and 1 newly introduced issue...

PR #1754, opened 2022-04-03
Speedup conservation law computation

FFroehlich:
So it turns out, using `sum` for symbolic variables is a really bad idea (see https://github.com/sympy/sympy/issues/13945).

Tested on a model with ~8k species.
Previous:
```
2022-04-02 20:52:20.502 - amici.pysb_import - INFO - Finished computing PySB conservation laws    + (3.17E+03s)
```

Now:
```
2022-04-03 09:30:18.220 - amici.pysb_import - INFO - Finished computing PySB conservation laws    + (1.90E+02s)
```


dweindl (approved):
:flushed: :+1:

PR #1761, opened 2022-04-04
Exploit stoichiometric matrix in pysb import

FFroehlich:
Exploits stoichiometric matrix to compute `dxdotdx_explicit`, `dxdotdp_explicit` and  `dxdotdw` as sparse matrix multiplication with derivatives of the flux vector that can be computed efficiently.

### DIFF
--- a/python/amici/pysb_import.py
+++ b/python/amici/pysb_import.py
@@ -240,7 +240,7 @@ def ode_model_from_pysb_importer(
         for noise_distr in noise_distributions.values()
     )
 
-    _process_stoichiometric_matrix
+    _process_stoichiometric_matrix(model, ode)
 
     ode.generate_basic_variables()
 
--- a/python/amici/pysb_import.py
+++ b/python/amici/pysb_import.py
@@ -240,9 +240,90 @@ def ode_model_from_pysb_importer(
         for noise_distr in noise_distributions.values()
     )
 
+    _process_stoichiometric_matrix
+
     ode.generate_basic_variables()
 
     return ode
+
+
+@log_execution_time('processing PySB stoich. matrix', logger)
+def _process_stoichiometric_matrix(pysb_model: pysb.Model,
+                                   ode_model: ODEModel) -> None:
+
+    """
+    Exploits the PySB stoichiometric matrix to generate xdot derivatives
+
+    :param pysb_model:
+        pysb model instance
+
+    :param ode_model:
+        ODEModel instance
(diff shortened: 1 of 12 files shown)


FFroehlich:
Ran two quick tests with smaller models. Looks like its faster even for small models and improvement increases with model size.

`nx` = 50 (*1.8 fold improvement*)

new:
```
2022-04-04 18:52:33.544 - amici.ode_export - INFO - Finished generating cpp code                    (1.59E+01s)
```

old:
```
2022-04-04 18:54:37.827 - amici.ode_export - INFO - Finished generating cpp code                    (2.94E+01s)
```

`nx` = 93 (*2.1 fold improvement*)
new:
```
2022-04-04 19:00:46.940 - amici.ode_export - INFO - Finished generating cpp code                    (5.59E+01s)
```

old:
```
2022-04-04 18:57:21.643 - amici.ode_export - INFO - Finished generating cpp code                    (1.17E+02s)
```

FFroehlich:
For model with ~8k species, processing via stoichiometric matrix completes in ~10min where `dxdotdw` alone needs >30min and `dxdotdx` computation has been running for >18h.

python/amici/pysb_import.py
```diff
@@ -240,11 +240,111 @@ def ode_model_from_pysb_importer(
...
+    p_idx = dict()
+    wx_idx = dict()
+
+    def get_cached_index(symbol, sarray, index_cache):
+        if symbol not in index_cache:
+            idx = sarray.index(symbol)
+            index_cache[symbol] = idx
+        else:
+            idx = index_cache[symbol]
+        return idx
```
dweindl, line 293:
```suggestion
        idx = index_cache.get(symbol, None)
        if idx is not None:
            return idx
        idx = sarray.index(symbol)
        index_cache[symbol] = idx
        return idx
```

Avoids looking up `symbol` in `index_cache` twice

python/amici/pysb_import.py
```diff
@@ -240,11 +240,111 @@ def ode_model_from_pysb_importer(
...
+
+    for ir, rxn in enumerate(pysb_model.reactions):
+        for ix in np.unique(rxn['reactants']):
+
+            if ix in solver_index:
+                # species
+                idx = solver_index[ix]
```
dweindl, line 300: Change the same way as suggested above?

dweindl (approved):
Nice!

PR #1763, opened 2022-04-05
update windows installation

paulflang:
@dweindl : I just installed AMICI on Windows using Visual Studio. It actually still largely works as described. But I made minor updates to the docs that may help future users.


dweindl (approved):
Thanks a lot @paulflang. That's very helpful.

PR #1764, opened 2022-04-05
Use `amici.AmiciVersionError` to indicate version mismatch

dweindl:
Using a more specific exception will allow downstream packages to handle version mismatches better, by e.g. automatically re-importing such models.


FFroehlich (approved):
👍

PR #1765, opened 2022-04-06
Speedup for models with conservation laws

### LINKED ISSUE #1762: Potential speedups for larger models
Some ideas in case we run into trouble with larger models again:

1) `dtotal_cldx_rdata`: if we store the species coefficients we should be able to compute the derivative for free
2) `dx_rdatadx_solver`: same story, store species coefficient and get the derivative for free
3) `dwdx`: the first couple of rows in this matrix are equal to `dtotal_cldx_rdata*dx_rdatadx_solver`, which we should be able to compute efficiently using sparse matrix multiplication.

FFroehlich:
fixes #1762 

Explanation: By storing the coefficients for conservation laws we can compute super fast derivatives w.r.t. `x`, speeding up `dtotal_cldx_rdata`, `dx_rdata_dx_solver` and parts of `dwdx`.

### DIFF
--- a/python/amici/ode_export.py
+++ b/python/amici/ode_export.py
@@ -1617,6 +1617,16 @@ class ODEModel:
             # force symbols
             self._eqs[name] = self.sym(name)
 
+        elif name == 'dwdx':
+            x = self.sym('x')
+            self._eqs[name] = sp.Matrix([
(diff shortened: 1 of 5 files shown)


dweindl (approved):
Nice :+1:

PR #1766, opened 2022-04-07
Sparsify jacobian

FFroehlich:
Sparsifies `smart_jacobian`, ~2x speedup for performance benchmark.


FFroehlich:
`OKAY: 812.31 < 2100` 😎

PR #1774, opened 2022-04-08
Remove long deprecated sbml2amici arguments modelName and constantParameters

dweindl:
I think it's safe to remove them...

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -205,24 +205,25 @@ class SbmlImporter:
         self.symbols = copy.deepcopy(default_symbols)
         self._local_symbols = {}
 
-    def sbml2amici(self,
-                   model_name: str = None,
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):
👍  yes

python/tests/test_misc.py
```diff
@@ -28,32 +28,6 @@ def test_parameter_scaling_from_int_vector():
     assert scale_vector[2] == amici.ParameterScaling.none
 
 
-def test_sbml2amici_no_observables():
```
FFroehlich, line 31: ?

dweindl: Duplicate :man_shrugging: 

https://github.com/AMICI-dev/AMICI/blob/75f7a20be536e839c35d2b6b89da6066f8e206be/python/tests/test_sbml_import.py#L40-L54

PR #1775, opened 2022-04-08
Improve efficiency of newton step conv check

FFroehlich:
Some profiling suggests that one convergence check comes at the cost of about 10 simulation steps, so checking convergence in every step dominates computation time. Might be better to have some adaptive approach based on the value of `wrms_`, but this will hopefully work sufficiently well in most cases.

### DIFF
--- a/src/steadystateproblem.cpp
+++ b/src/steadystateproblem.cpp
@@ -660,7 +660,7 @@ void SteadystateProblem::runSteadystateSimulation(const Solver &solver,
          system is prepared for newton type convergence check */
         if (wrms_ < conv_thresh && check_sensi_conv_ &&
             sensitivityFlag == SensitivityMethod::forward &&
-            convergence_check_frequency % sim_steps == 0) {
+            sim_steps % convergence_check_frequency == 0) {
             updateSensiSimulation(solver);
             wrms_ = getWrmsFSA(model);
         }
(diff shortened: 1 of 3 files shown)


src/steadystateproblem.cpp
```diff
@@ -654,7 +659,8 @@ void SteadystateProblem::runSteadystateSimulation(const Solver &solver,
         /* getWrms needs to be called before getWrmsFSA such that the linear
          system is prepared for newton type convergence check */
         if (wrms_ < conv_thresh && check_sensi_conv_ &&
-            sensitivityFlag == SensitivityMethod::forward) {
+            sensitivityFlag == SensitivityMethod::forward &&
+            convergence_check_frequency % sim_steps == 0) {
```
dweindl, line 663: This will check convergence at iterations 1, 5, and 25, and never thereafter. I don't think that's intentional?! 

```suggestion
            sim_steps % convergence_check_frequency == 0) {
```

FFroehlich:
> Some profiling suggests that one convergence check comes at the cost of about 10 simulation steps, so checking convergence in every step dominates computation time. Might be better to have some adaptive approach based on the value of `wrms_`, but this will hopefully work sufficiently well in most cases.

Note that the computational complexity comes from refactorization of the jacobian. What we could also do is something similar to CVODES, where refactorization is done adaptively. As far as I understand the documentation, the default maximal number of steps between refactorizations (set by CVodeSetLSetupFrequency) is 20, so 25 is probably a good value here.

PR #1776, opened 2022-04-08
Specify initial timepoint with ExpData

dilpath:
Might be useful to add other things too, e.g. `t_presim`, but I have only tried `tstart_` for my use case so far.
https://github.com/AMICI-dev/AMICI/blob/a2e543796d9f478673afd8cbd60fd2286912dc12/include/amici/simulation_parameters.h#L181


dweindl (approved):
LGTM

FFroehlich (approved):
👍  yes makes sense. Wouldn't put too much effort into presimulation as it is likely going to be superseded by allowing sequences of conditions.

PR #1778, opened 2022-04-08
Parallelize matrix simplification

dweindl:
Allow for optional parallelization of sympy matrix simplification using `multiprocessing`. Number of processes is controlled via environment variable `AMICI_IMPORT_NPROCS`.

This only helps for larger models and may slow down import of smaller models slightly.

Benchmarking (only code generation; derivatives computed sequentially in both cases):
* 1 process: 2:26:19 (h:mm:ss)  (100%)
* 4 processes: 1:53:01 (h:mm:ss) (78%)


FFroehlich (approved):
Nice! Would be nice to add some documentation on the respective environment variable.

PR #1780, opened 2022-04-09
limit newton step convergence check take 2

FFroehlich:
Also fixes a potential bug where backwards solve counts would be saved to the wrong step count.


dweindl (approved):
:+1:

PR #1783, opened 2022-04-21
Avoid error if no measurements in PEtab problem; fix type handling in PEtab parameter mapping

dilpath:
No measurements can occur if simulating a timecourse as a series of PEtab problems, where each PEtab problem has a subset of the original measurements table, and some timecourse period happens to have no measurements during it.

`iterrows` fails if `simulation_conditions` is passed as a dictionary.
https://github.com/AMICI-dev/AMICI/blob/da34367cba51e35aaaddbf00707a16a5135b01c7/python/amici/petab_objective.py#L281

### DIFF
--- a/python/amici/petab_objective.py
+++ b/python/amici/petab_objective.py
@@ -153,9 +153,10 @@ def simulate_petab(
 
     # Log results
     sim_cond = petab_problem.get_simulation_conditions_from_measurement_df()
-    for i, rdata in enumerate(rdatas):
-        logger.debug(f"Condition: {sim_cond.iloc[i, :].values}, status: "
-                     f"{rdata['status']}, llh: {rdata['llh']}")
+    if not sim_cond.empty:
+        for i, rdata in enumerate(rdatas):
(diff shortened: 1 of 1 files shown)


python/amici/petab_objective.py
```diff
@@ -153,9 +153,10 @@ def simulate_petab(
 
     # Log results
     sim_cond = petab_problem.get_simulation_conditions_from_measurement_df()
-    for i, rdata in enumerate(rdatas):
-        logger.debug(f"Condition: {sim_cond.iloc[i, :].values}, status: "
-                     f"{rdata['status']}, llh: {rdata['llh']}")
+    if not sim_cond.empty:
+        for i, rdata in enumerate(rdatas):
+            logger.debug(f"Condition: {sim_cond.iloc[i, :].values}, status: "
+                         f"{rdata['status']}, llh: {rdata['llh']}")
```
dweindl, line 159: Might be confusing to look at the log and not seeing any output for the respective condition? Maybe just replace `{sim_cond.iloc[i, :].values}` by `N/A` if unavailable?

dweindl (approved):

Issue #1788, opened 2022-05-02
ValueError: tuple.index(x): x not in tuple in `simulate_petab_amici`


matthiaskoenig:
When running the following petab
[icg_petab_2022-02-05.zip](https://github.com/AMICI-dev/AMICI/files/8603378/icg_petab_2022-02-05.zip)
I get a ValueError. I validated the files with the `petablint` and do not get any errors/warnings.

```python
from pathlib import Path

import petab
from amici.petab_import import import_petab_problem
from amici.petab_objective import simulate_petab


def simulate_petab_amici(yaml_file: Path):
    """Simulate PETab."""

    petab_problem = petab.Problem.from_yaml(yaml_file)
    model = import_petab_problem(petab_problem)
    simulate_petab(petab_problem, model)


if __name__ == "__main__":
    output_dir: Path = Path(__file__).parent / "results" / "icg"
    yaml_file: Path = output_dir / "icg.yaml"
    simulate_petab_amici(yaml_file=yaml_file)
```
Results in 
```bash
Traceback (most recent call last):
  File "/home/mkoenig/git/pkdb_models/pkdb_models/models/icg/fit_petab/icg_amici2.py", line 21, in <module>
    simulate_petab_amici(yaml_file=yaml_file)
  File "/home/mkoenig/git/pkdb_models/pkdb_models/models/icg/fit_petab/icg_amici2.py", line 15, in simulate_petab_amici
    simulate_petab(petab_problem, model)
  File "/home/mkoenig/.virtualenvs/pkdb_models/lib/python3.9/site-packages/amici/logging.py", line 194, in wrapper_timer
    rval = func(*args, **kwargs)
  File "/home/mkoenig/.virtualenvs/pkdb_models/lib/python3.9/site-packages/amici/petab_objective.py", line 133, in simulate_petab
    edatas = create_edatas(
  File "/home/mkoenig/.virtualenvs/pkdb_models/lib/python3.9/site-packages/amici/petab_objective.py", line 521, in create_edatas
    edata = create_edata_for_condition(
  File "/home/mkoenig/.virtualenvs/pkdb_models/lib/python3.9/site-packages/amici/petab_objective.py", line 597, in create_edata_for_condition
    y, sigma_y = _get_measurements_and_sigmas(
  File "/home/mkoenig/.virtualenvs/pkdb_models/lib/python3.9/site-packages/amici/petab_objective.py", line 695, in _get_measurements_and_sigmas
[...]

PR #1791, opened 2022-05-03
Workaround for PEtab problems with state-dependent noise models

dweindl:
PEtab does currently not allow observables in noiseFormula and AMICI cannot handle states in sigma expressions. Therefore, where possible, replace species occurring in the error model definition by observableIds.

See also #1788

### DIFF
--- a/python/amici/petab_import.py
+++ b/python/amici/petab_import.py
@@ -533,7 +533,8 @@ def import_model_sbml(
                            key=lambda symbol: symbol.name)
         for free_sym in free_syms:
             sym = str(free_sym)
-            if sbml_model.getElementBySId(sym) is None and sym != 'time':
+            if sbml_model.getElementBySId(sym) is None and sym != 'time' \
+                    and sym not in observables:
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):
test with` Raia_CancerResearch2011` bechmark model? if gradients don't match for that model just ignore?

FFroehlich:
> test with` Raia_CancerResearch2011` bechmark model? if gradients don't match for that model just ignore failure?

FFroehlich:
> test with` Raia_CancerResearch2011` bechmark model? if llh value doesn't match just ignore for that model?

dweindl:
> test with` Raia_CancerResearch2011` bechmark model? if gradients don't match for that model just ignore?

In its [current](https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/commit/4bc0af9b142c73f9cce57f7899e511f0f53492ab) form, that one uses observables in `noiseFormula`, not states. So not a good test case (yet).

PR #1805, opened 2022-06-01
Re-arrange application of parameters from edata to avoid initial sensis error

dilpath:
When using `ExpData` to specify simulation settings, initial state sensitivities were set before parameters were set [1]. Setting initial state sensitivities involves unscaling scaled-sensitivities using parameters [2].

If the parameter scale is `log*`, but the parameter value in the model is `0`, then there will be a division by zero [2]. This is also incorrect, if sensitivities should be computed (and unscaled) with respect to parameters supplied with `ExpData`.

This PR just re-arranges how the settings are set, so that parameters from `ExpData` are set in the model before model parameters are used to unscale scaled-initial-state-sensitivities.

[1] https://github.com/AMICI-dev/AMICI/blob/477fd1bb037f03f1c22ec0da02309eaf93dae7cf/src/edata.cpp#L371-L409=
[2] https://github.com/AMICI-dev/AMICI/blob/8fe888416d009728b3bb34b833d7f9d392735ea1/src/model.cpp#L763-L784=

### DIFF
--- a/.github/workflows/deploy_protected.yml
+++ b/.github/workflows/deploy_protected.yml
@@ -28,7 +28,7 @@ jobs:
     - uses: actions/checkout@master
     - run: git archive -o docker/amici.tar.gz --format=tar.gz HEAD
     - name: Publish to Registry
-      uses: elgohr/Publish-Docker-Github-Action@2.8
+      uses: elgohr/Publish-Docker-Github-Action@v4
       with:
         name: dweindl/amici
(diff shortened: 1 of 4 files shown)


FFroehlich (approved):
👍 

Would be great to at least have a comment that mentions this (ideally a test) so this doesn't get messed up in the future

dweindl (approved):
:+1:

PR #1806, opened 2022-06-10
move sim_steps increment

plakrisenko:
moved `sim_steps` increment to the beginning of the while loop, as otherwise if the simulation is successful on the first step resulting `sim_steps = 0`. 

moved also the max steps check and the large time point check, as otherwise one step is made even if maxSteps is set to zero


FFroehlich (approved):
👍

PR #1810, opened 2022-06-27
Treat non-estimated parameters as constants during SBML-PEtab import

dweindl:
So far, only parameters from the condition table were treated as constants, but not other non-estimated ones. The latter are added here.

Also 
* changes the signature `petab_import.import_model_sbml` to take a `petab.Problem` instead of individual files.
* uses the recent PEtab model abstraction and fixes some DeprecationWarnings


python/amici/petab_import.py
```diff
@@ -73,25 +73,60 @@ def _add_global_parameter(sbml_model: libsbml.Model,
...
-        condition_df: Optional[pd.DataFrame] = None,
+        petab_problem: petab.Problem
 ) -> List[str]:
     """
     Determine, set and return fixed model parameters.
 
-    Parameters specified in `condition_df` are turned into constants.
+    Non-estimated parameters and parameters specified in the condition table
+    are turned into constants (unless they are overridden).
     Only global SBML parameters are considered. Local parameters are ignored.
```
FFroehlich, line 83: we are going to use the same function for other models aswell right? maybe state this comment only applies to SBML models?

python/amici/petab_import.py
```diff
@@ -73,25 +73,60 @@ def _add_global_parameter(sbml_model: libsbml.Model,
...
+    #  listed in the parameter table, but are not marked as estimated, can be
+    #  turned in to AMICI constants
+    # due to legacy API, we might not always have a parameter table, though
+    fixed_parameters = set()
+    if petab_problem.parameter_df is not None:
+        all_parameters = get_valid_parameters_for_parameter_table(
+            model=petab_problem.model,
+            condition_df=petab_problem.condition_df,
+            observable_df=petab_problem.observable_df
+            if petab_problem.observable_df is not None
```
FFroehlich, line 118: wouldn't it be better to have some default handling in `get_valid_parameters_for_parameter_table`?

FFroehlich (approved):
👍

python/amici/petab_import.py
dweindl: That will require a new `libpetab` release, so I will leave that in amici for now.

PR #1814, opened 2022-06-30
Fix missing CPU time reset

### LINKED ISSUE #1812: ReturnData::cpu_timeB is incorrect when re-using the same Solver object
Seems like it's not reset correctly.

dweindl:
Fixes incorrect CPU times reported if a Solver instance is used for multiple simulations. (So far, the cumulative CPU time would have been reported).

Fixes #1812


FFroehlich (approved):
👍

PR #1822, opened 2022-07-05
Doc: Fix petab intersphinx mapping

dweindl:
This probably broke when separating PEtab format specification and Python library.


dilpath (approved):
:+1:

PR #1824, opened 2022-07-10
Update valgrind suppressions / fix cpu_timeB reset

### LINKED ISSUE #1812: ReturnData::cpu_timeB is incorrect when re-using the same Solver object

dweindl:
* Update valgrind suppressions
* Formatting
* Fix special missing cpu_timeB reset in case `Solver` is used without adjoints after initial use with adjoints (fixes #1812) 
* Raise in case of contract violation (instead of terminating without proper message)
* Unique names for test models


FFroehlich (approved):
👍

PR #1828, opened 2022-07-12
Show model path in version incompatibility warnings

dweindl:
Makes it easier to find and delete the problematic model.


FFroehlich (approved):
:+1:

PR #1830, opened 2022-07-14
Fix ImportErrors during package installation with recent setuptools

### LINKED ISSUE #1829: Python package installation fails with setuptools-63.2.0
Python package installation fails with setuptools-63.2.0 (or maybe already setuptools-63.0.0) due to import errors (e.g. https://github.com/AMICI-dev/AMICI/actions/runs/2668045168).

This is because `amici._imported_from_setup` fails, because `setup()` is no longer in the call stack.

Working on it.

dweindl:
Closes #1829


FFroehlich (approved):
👍

PR #1832, opened 2022-07-18
GHA: More nightly tests

dweindl:


.github/workflows/test_valgrind.yml
```diff
@@ -8,6 +8,8 @@ on:
     branches:
       - master
   workflow_dispatch:
+  schedule:
+    - cron:  '48 4 * * *'
```
FFroehlich, line 12: to track the update that triggered future issues?

FFroehlich (approved):
👍

.github/workflows/test_valgrind.yml
dweindl: :heavy_check_mark:

PR #1833, opened 2022-07-18
fix_cls

FFroehlich:
Fixes an issue where flattening would end up in an infinite loop due to coefficients of a cl state being `0.0`. Harmonizing checks should be future proof.


dweindl (approved):
:+1:

PR #1835, opened 2022-07-21
Checked narrowing casts

dweindl:
Use gsl::narrow instead of plain static_cast. This will let us spot quickly when things get too large for our omnipresent to-int conversions.


FFroehlich (approved):
👍

PR #1836, opened 2022-07-22
Update to SUNDIALS 5.8.0

dweindl:
See https://github.com/LLNL/sundials/releases/tag/v5.8.0 for changes.


dweindl:
FAILED test_compare_conservation_laws_sbml.py::test_adjoint_pre_and_post_equilibration
FAILED test_preequilibration.py::test_equilibration_methods_with_adjoints

Looks like this breaks state reinitialization. I don't see any related change in SUNDIALS :/

src/solver_cvodes.cpp
```diff
@@ -118,6 +118,8 @@ void CVodeSolver::initSteadystate(const realtype /*t0*/, const AmiVector &/*x0*/
        as re-calling init would unset solver settings. */
     auto cv_mem = static_cast<CVodeMem>(solver_memory_.get());
     cv_mem->cv_f = fxBdot_ss;
+
```
FFroehlich, line 121: maybe add a comment that both functions must be set for simulation to work properly and that just setting `cv_mem->cv_f` will result in the previous function being used as alternate Nls function?

FFroehlich (commented):
👍

Issue #1813, opened 2022-06-29
ReturnData::numstepsB is not set correctly when re-using the same Solver instance


dweindl:
First run with a given Solver/Model/ExpData:
`ReturnData::numstepsB = {0, 3, 9, 15, }`
second run with the same inputs:
`ReturnData::numstepsB = {15, 3, 9, 15, }`

Looks as if for the first timepoint, the number of steps from the last timepoint from the previous run are used. 
(Results obtained with adjoint preequilibration enabled, not sure if it matters).

dweindl:
The problem seems to be that `handleDataPointB` is called before solver stats are reset in `setupB`: https://github.com/AMICI-dev/AMICI/blob/e835f94f00c68a93090a7d86a6be97931506ddba/src/backwardproblem.cpp#L79-L80
Just swapping the two lines breaks a number of tests.

FFroehlich:
fixed in #1824

dweindl:
> fixed in #1824

Unfortunately not.

dweindl:
Added a test case in #1841, would be great if you could have a look into that.

PR #1841, opened 2022-08-01
Fix resetting ReturnData::numstepsB when re-using Solver

dweindl:
So far only a test case demonstrating #1813.


FFroehlich (approved):

FFroehlich:
sounds like numsteps is also not properly ordered in some cases?

dweindl:
> sounds like numsteps is also not properly ordered in some cases?

The last one seems to be 0 sometimes :thinking:

dweindl:
> > sounds like numsteps is also not properly ordered in some cases?
> 
> The last one seems to be 0 sometimes thinking

Ah, for t=inf it should just copy the previous value, as it's supposed to be the cumulative sum.

FFroehlich:
> > > sounds like numsteps is also not properly ordered in some cases?
> > 
> > 
> > The last one seems to be 0 sometimes thinking
> 
> Ah, for t=inf it should just copy the previous value, as it's supposed to be the cumulative sum.

would keep the numbers as is, as otherwise we would also have to carry over numbers from pre-equilibration and it's probably less confusing if these values are kept seperate.

dweindl:
> would keep the numbers as is, as otherwise we would also have to carry over numbers from pre-equilibration and it's probably less confusing if these values are kept separate.

Ok. Could consider changing from cumulative sum to deltas, at some point...

PR #1842, opened 2022-08-03
Validate PEtab problems before attempting import

### LINKED ISSUE #1691: Lint petab models
CLI already lints, but programmatic interface doesn't lint by default. Model linting should definitely be performed when import fails to avoid having to reimplement validation checks in amici and provide helpful error messages.

dweindl:
Closes #1691


FFroehlich (approved):
👍

PR #1844, opened 2022-08-04
Fix potential deadlocks during parallel simplification

dweindl:
For larger models, I experienced occassional deadlocks during import. Unclear why.
Most such issues seem to be related to forking, so let's do without.


FFroehlich (approved):
👍  nice!

PR #1846, opened 2022-08-06
Cleanup includes

dweindl:
Removing unused includes might improve compile times slightly.

Moves a few functions/classes to more appropriate places.

Also adds a CMake target for the matlab extension, making it easier to check for compilation errors there.


FFroehlich (approved):
👍

PR #1848, opened 2022-08-12
Pin setuptools for python-C++ code coverage

dweindl:
setuptools>=64 builds the extension in some temporary directory, even with editable installation and no build isolation. This breaks our current code coverage analysis. Not sure how to best address this. For now, pin setuptools.


FFroehlich (approved):
👍

PR #1849, opened 2022-08-20
Release 0.12.0

dweindl:


FFroehlich:
Would argue that with the addition of event observables, we should bump the version number to `0.12.0`.

dweindl:
> Would argue that with the addition of event observables, we should bump the version number to `0.12.0`.

Fine with me. Wasn't done very consistently so far... Should do on any new feature...

FFroehlich:
> > Would argue that with the addition of event observables, we should bump the version number to `0.12.0`.
> 
> Fine with me. Wasn't done very consistently so far... Should do on any new feature...

Yes, good point! Some of those steady-state computation features should definitively have been released as minor release rather than a patch. We should probably discuss what we consider "new functionality", but let's try to be more consistent in the future.

dweindl:
> Yes, good point! Some of those steady-state computation features should definitively have been released as minor release rather than a patch. We should probably discuss what we consider "new functionality", but let's try to be more consistent in the future.

Yes. I'd say new functionality is anything that is not clearly only a bug fix (without API change). We don't need to worry about running out of version numbers...

FFroehlich (approved):

PR #1851, opened 2022-08-21
Remove multiplications with 1.0 in xdot

dweindl:
Use sympy.Integer instead of sympy.Float for integer coefficients. 

Closes #1681 

Reduces file size, makes code more readable.

### DIFF
--- a/python/amici/sbml_import.py
+++ b/python/amici/sbml_import.py
@@ -664,7 +664,7 @@ class SbmlImporter:
                 self.sbml.getConversionFactor()
             )
         else:
-            conversion_factor = 1.0
+            conversion_factor = 1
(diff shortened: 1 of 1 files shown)


python/amici/sbml_import.py
```diff
@@ -927,7 +927,7 @@ def _process_reactions(self):
                     self.stoichiometric_matrix[species['index'],
                                                reaction_index] += \
                         sign * stoichiometry * species['conversion_factor']
-
+            print(self.stoichiometric_matrix)
```
dweindl, line 930:
```suggestion

```

FFroehlich (approved):
👍  neat

PR #1852, opened 2022-08-21
Extract common subexpressions

dweindl:
Add option to extract common subexpressions during code generation (at the model function level, not added to `w`).

Disabled by default, can be enabled via environment variable `AMICI_EXTRACT_CSE=1`.

For large models, this may significantly reduce file size and speed up compilation.
Downside: The resulting files are hard to read / debug.

With my test model (SBML import, 2 processes, sequential compilation with clang14, -O2):

|                 | before         | after          |
|-----------------|----------------|----------------|
| import time     | 160 min (100%) | 164 min (103%) |
| code size       | 89 MB (100%)   | 27 MB (30%)    |
| compile time    | 169 min (100%) | 90 min (53%)   |
| compile RAM     | 7.49 GB (100%) | 1.18 GB (16%)  |

### DIFF
--- a/python/amici/cxxcodeprinter.py
+++ b/python/amici/cxxcodeprinter.py
@@ -1,9 +1,11 @@
 """C++ code generation"""
+import os
 import re
-from typing import List, Optional, Tuple, Dict
+from typing import Dict, List, Optional, Tuple
 
 import sympy as sp
 from sympy.printing.cxx import CXX11CodePrinter
+from sympy.utilities.iterables import numbered_symbols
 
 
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):
:+1: Please document usage of environment variable. Maybe it makes sense to introduce a dedicated doc page that lists and explains all available environment variables?

FFroehlich:
> 👍 Please document usage of environment variable. Maybe it makes sense to introduce a dedicated doc page that lists and explains all available environment variables?

Oh, and maybe add a test that runs cse on a nontrivial testcase?

PR #1853, opened 2022-08-22
Deterministic code generation

dweindl:
Sort toposort results within each group to always generate identical code.
Makes it easier to `diff` models.


FFroehlich (approved):
👍

PR #1860, opened 2022-08-23
Proper type annotations for the swig interface

dweindl:
Replaces a big part of the swig-generated C++ annotations by proper Python annotations, so that they can actually be used for type checking.


FFroehlich (approved):
Neat! 👍

PR #1862, opened 2022-08-30
Add downstream test for pypesto

yannikschaelte:
Add downstream tests of tools that depend on AMICI and for which we want to ensure near-continuous compatibility (e.g. pyPESTO). These tests are not expected to pass, but can inform when AMICI changes can lead to problems downstream.

### DIFF
--- a/.github/workflows/test_downstream.yml
+++ b/.github/workflows/test_downstream.yml
@@ -38,10 +38,9 @@ jobs:
       run: |
         scripts/installAmiciSource.sh
 
-    - name: Install pyPESTO
+    - name: Download pyPESTO
       run: |
         git clone https://github.com/ICB-DCM/pyPESTO.git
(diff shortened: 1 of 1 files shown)


yannikschaelte:
Marking as `allow-failure` appears not possible yet https://github.com/community/community/discussions/15452

FFroehlich:
Not convinced that this is necessary or desired. AMICI API is pretty stable and we are pretty consistent in maintaining downstream compatibility. The issue in https://github.com/ICB-DCM/pyPESTO/issues/924#issuecomment-1234777385 was a result of abusing the API.

dweindl:
I don't think we should add downstream tests here. I'd say, there should rather be a test with amici@develop in pypesto.

yannikschaelte:
Ok!

PR #1863, opened 2022-09-05
Skip additional tests under valgrind

dweindl:
Valgrind runs are taking too long again.

* Let's skip some more tests under valgrind. (`python/tests/test_events.py::test_models[events_plus_heavisides]` alone takes ~1.5h)
* Split Valgrind C++ and Python tests into 2 parallel jobs
* Clean up running-under-valgrind test logic, to also work without GHA or setting extra env vars

Valgrind jobs now finish in about 30min instead of 3h.

### DIFF
--- a/python/tests/test_pregenerated_models.py
+++ b/python/tests/test_pregenerated_models.py
@@ -11,6 +11,8 @@ import h5py
 import numpy as np
 import pytest
 from amici.gradient_check import check_derivatives, _check_results
+from amici.testing import skip_on_valgrind
+
 
 cpp_test_dir = Path(__file__).parents[2] / 'tests' / 'cpp'
 options_file = str(cpp_test_dir / 'testOptions.h5')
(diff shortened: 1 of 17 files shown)


python/tests/test_pregenerated_models.py
```diff
@@ -22,8 +24,7 @@
                for case in list(expected_results[sub_test].keys())]
 
 
-@pytest.mark.skipif(os.environ.get('GITHUB_JOB') == 'valgrind',
-                    reason="Takes too long under valgrind")
+@skip_on_valgrind
```
FFroehlich, line 27: wasn't aware these tests are currently skipped and not sure whether it's a good idea since we lose quite a bit of coverage, especially regarding second order and event observables.

FFroehlich (approved):
👍

python/tests/test_pregenerated_models.py
dweindl: Those are the same features that we are testing directly from C++ (also under valgrind). Unless we are worried about the swig-generated code, I don't think it adds much value to re-run those via Python. Even then, I don't think there are any swig-functions exercised in these tests that aren't already covered elsewhere.

FFroehlich: ah, good point!

PR #1865, opened 2022-09-10
Fix extraction of common subexpressions

dweindl:
Got broken in #1859 (https://github.com/AMICI-dev/AMICI/commit/cc5f7cc681b393f9500eb6284823528ce83b25e6#diff-44c38f8a81d42beabd5ab6dfd5db9a33bda06ded71f7cbfa5a47a7a1b8fd824eL3028), so that `AMICI_EXTRACT_CSE=1` had no effect.

That change also broke the use of human-readable lhs symbols in generated code. Also fixed here.

Added tests for both.


FFroehlich (approved):
👍

PR #1868, opened 2022-09-13
Reduce symlinking

FFroehlich:
Reduces symlinking by moving files from `python/amici` to `python/sdist/amici`.


FFroehlich:
Files that remain in `python/amici`: `__init__.py`, `import_utils.py`, `setuptools.py` and `swig.py`. Not sure whether any of those could also be moved.

dweindl:
> Not sure whether any of those could also be moved.

Should be possible. Nothing special about them, as far as I can tell.

dweindl (approved):

PR #1877, opened 2022-10-26
PEtab import: Add option for treating fixed parameters as constants

dweindl:
Makes the changes from https://github.com/AMICI-dev/AMICI/pull/1810/ optional, as sometimes users prefer more flexible models.
The default remains to treat non-estimated parameters in PEtab as constants in AMICI.

Also adds a basic check for loading existing models during PEtab, to see if they are compatible to what was supposed to be imported. Only checks non-constant parameter IDs so far.

### DIFF
--- a/python/sdist/amici/petab_import.py
+++ b/python/sdist/amici/petab_import.py
@@ -340,14 +340,16 @@ def import_petab_problem(
     return model
 
 
-def check_model(amici_model: amici.Model, petab_problem: petab.Problem) -> None:
+def check_model(
+    amici_model: amici.Model,
+    petab_problem: petab.Problem,
+) -> None:
     """Check that the model is consistent with the PEtab problem."""
     if petab_problem.parameter_df is None:
         return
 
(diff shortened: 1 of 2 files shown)


python/sdist/amici/petab_import.py
```diff
@@ -306,18 +325,46 @@ def import_petab_problem(
...
 
 
+def check_model(amici_model: amici.Model, petab_problem: petab.Problem) -> None:
+    """Check that the model is consistent with the PEtab problem."""
+    if petab_problem.parameter_df is None:
+        return
+
+    amici_ids_free = set(amici_model.getParameterIds())
+    amici_ids = amici_ids_free | set(amici_model.getFixedParameterIds())
+    petab_ids = set(petab_problem.parameter_df.index)
```
dweindl, line 350: currently unused

dilpath (approved):
:+1:

FFroehlich (approved):
👍

PR #1377, opened 2021-01-18
Use SymPy simple floating-point optimizations

lcontento:
SymPy offers some simple floating-point optimizations, mainly `exp(1-x) -> expm1(x)` and `log(1+x) -> log1p(x)`. Using them by default should not break anything as far as I can see. Only downside should be an increase in code generation time.

PR #1878, opened 2022-10-28
Make sympy code optimizations optional

dweindl:
Turns out that applying sympy optimizations as introduced in https://github.com/AMICI-dev/AMICI/pull/1377 is potentially very costly. Therefore, they are disabled by default. They can be enabled by setting `AmiciCxxCodePrinter.optimizations` as shown in the test case.


FFroehlich (approved):
👍

PR #1892, opened 2022-11-21
Code-gen: Fix missing `return` in non-void-returning functions

dweindl:
... and set `-Werror` for compiling models in debug mode.


FFroehlich (commented):
👍

PR #1894, opened 2022-11-21
Add macos hdf5 header/library paths

Doresic:
Added the correct folder to search for hdf5 in macOS M1 systems in `setuptools.py`


dweindl (approved):
Thanks :+1:

PR #1897, opened 2022-11-22
Fix `petab_import.import_model(..., compile=False)` failure

dweindl:
Fixes a bug where amici would fail importing an extension that was never built.


FFroehlich (approved):
how did this ever work before? 🤣

Issue #1899, opened 2022-11-23
tempfile is deprecated


FFroehlich:
`WARNING: tempfile is deprecated; consider using mktemp instead.`

Happens in during performance test

PR #1900, opened 2022-11-23
Fix performance test warnings

dweindl:
* Fix missing upload artifact
* Don't use 'tempfile' - fixes #1899

PR #1901, opened 2022-11-24
Used "mktemp" instead of "tempfile" in performance testing

j0ker70:
[Issue] #1899
[Problem] "tempfile" has become deprecated in shell-script. Reference: https://github.com/koalaman/shellcheck/wiki/SC2186 [Solution] Used "mktemp" in place of tempfile


FFroehlich:
Thank you for your contribution, this was already fixed in https://github.com/AMICI-dev/AMICI/pull/1900.

PR #1903, opened 2022-11-25
MultiArch docker build

### LINKED ISSUE #1902: Multi-Arch Dockerfile
Current Dockerfile is only compiled for amd64, would be nice to also have an arm64 image (better performance, avoids issues with AVX instructions in rosetta).

FFroehlich:
fixes #1902

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,7 +19,7 @@ COPY amici.tar.gz /tmp
 RUN pip3 install --upgrade pip build
 RUN pip3 install -v python-libsbml
 RUN mkdir -p /tmp/amici/ && cd /tmp/amici
-RUN tar -xzf ../amici.tar.gz && cd /tmp/amici/python/sdist
+RUN tar -xzf /amici.tar.gz && cd /tmp/amici/python/sdist
 
 # build sdist & install
 
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,9 @@
 FROM ubuntu:22.04
 
-RUN apt update \
-    && apt-get install -y \
+# install dependencies
+
+RUN apt update
+RUN apt-get install -y \
     g++ \
     libatlas-base-dev \
     python3 \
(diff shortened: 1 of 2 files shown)


FFroehlich:
for docker deploy check see here: https://github.com/AMICI-dev/AMICI/actions/runs/3563041908/jobs/5985384608

docker/Dockerfile
```diff
@@ -1,22 +1,36 @@
...
     g++ \
     libatlas-base-dev \
     python3 \
     python3-dev \
     python3-pip \
     python3-venv \
-    swig
+    swig \
+    # required for wheel-less libsbml installation 
+    cmake
```
dweindl, line 15: No longer required? (also `\` won't work with the comment in the following line)

docker/Dockerfile
```diff
@@ -1,22 +1,36 @@
...
 COPY amici.tar.gz /tmp
 
-RUN pip3 install --upgrade pip build \
-    && mkdir -p /tmp/amici/ \
-    && cd /tmp/amici \
-    && tar -xzf ../amici.tar.gz \
-    && cd /tmp/amici/python/sdist \
-    && python3 -m build --sdist \
-    && pip3 install -v $(ls -t /tmp/amici/python/sdist/dist/amici-*.tar.gz | head -1)[petab] \
-    && rm -rf /tmp && mkdir /tmp
```
dweindl, line 22: I'd keep the `&&` to avoid creating that many layers in the image.

docker/Dockerfile
```diff
@@ -1,22 +1,36 @@
 FROM ubuntu:22.04
 
-RUN apt update \
-    && apt-get install -y \
+# install dependencies 
+
+RUN apt update
+RUN apt-get install -y \
```
dweindl, line 6:
```suggestion
RUN apt-get update && apt-get install -y \
```

apt -> apt-get fixes `#0 0.049 WARNING: apt does not have a stable CLI interface. Use with caution in scripts.`

dweindl (approved):
:+1:

PR #1908, opened 2022-12-11
Update clang-format style, fix clang-format target

dweindl:
Previously, `make clang-format` missed all header files.


FFroehlich (approved):
👍

PR #1911, opened 2022-12-12
Subset ReturnData fields by ID

dweindl:
This adds functionality to access ReturnData fields by ID instead of by index.

### DIFF
--- a/python/sdist/amici/numpy.py
+++ b/python/sdist/amici/numpy.py
@@ -258,7 +258,7 @@ class ReturnDataView(SwigPtrView):
                 f"Subsetting {field} by ID is not implemented or not possible."
             )
         col_index = ids.index(entity_id)
-        return getattr(self, field)[:, col_index]
+        return getattr(self, field)[:, ..., col_index]
 
 
 class ExpDataView(SwigPtrView):
--- a/python/sdist/amici/numpy.py
+++ b/python/sdist/amici/numpy.py
@@ -8,8 +8,9 @@ import numpy as np
 import copy
 import collections
 
+import amici
 from . import ExpDataPtr, ReturnDataPtr, ExpData, ReturnData, Model
-from typing import Union, List, Dict, Iterator
+from typing import Union, List, Dict, Iterator, Literal
 
 
 class SwigPtrView(collections.abc.Mapping):
@@ -238,23 +239,37 @@ class ReturnDataView(SwigPtrView):
             item = 'ts'
         return super(ReturnDataView, self).__getitem__(item)
 
-    def by_id(self, model: Model, field: str, entity_id: str) -> np.array:
+    def by_id(
+            self,
+            entity_id: str,
+            field: str = None,
+            model: Model = None
+    ) -> np.array:
         """
         Get the value of a given field for a named entity.
 
-        :param model: The model from which this ReturnDataView was generated.
-        :param field: The requested field, e.g. 'x' for model states
         :param entity_id: The ID of the model entity that is to be extracted
             from ``field`` (e.g. a state ID).
+        :param field: The requested field, e.g. 'x' for model states. This is
+            optional if field would be one of ``{'x', 'y', 'w'}``
+        :param model: The model from which this ReturnDataView was generated.
+            This is optional if this ReturnData was generated with
+            ``solver.getReturnDataReportingMode() == amici.RDataReporting.full``.
         """
+        if field is None:
+            field = _entity_type_from_id(entity_id, self, model)
+
         if field in {'x', 'x0', 'x_ss', 'sx', 'sx0', 'sx_ss'}:
-            ids = model.getStateIds()
+            ids = (model and model.getStateIds()) or self._swigptr.state_ids
         elif field in {'w'}:
-            ids = model.getExpressionIds()
+            ids = (model and model.getExpressionIds()) \
+                  or self._swigptr.expression_ids
         elif field in {'y', 'sy', 'sigmay'}:
-            ids = model.getObservableIds()
+            ids = (model and model.getObservableIds()) \
+                  or self._swigptr.observable_ids
         elif field in {'sllh'}:
-            ids = model.getParameterIds()
+            ids = (model and model.getParameterIds()) \
+                  or self._swigptr.parameter_ids
         else:
             raise NotImplementedError(
                 f"Subsetting {field} by ID is not implemented or not possible."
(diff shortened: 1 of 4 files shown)


python/sdist/amici/numpy.py
```diff
@@ -238,6 +238,25 @@ def __getitem__(self, item: str) -> Union[np.ndarray, ReturnDataPtr,
...
+
+        :param model: The model from which this ReturnDataView was generated.
+        :param field: The requested field, e.g. 'x' for model states
+        :param entity_id: The ID of the model entity that is to be extracted
+            from ``field`` (e.g. a state ID).
+        """
+        if field in {'x', 'x0', 'sx'}:
+            col_index = model.getStateIds().index(entity_id)
+        elif field in {'w'}:
+            col_index = model.getExpressionIds().index(entity_id)
```
FFroehlich, line 253: `y`/`sy` and `model.getObservableIds()`?

FFroehlich: what about only assigning ids in the if clause, e.g.,
```
ids = model.getExpressionIds()
```
and then doing the indexing outside:
```
col_index = ids.index(entity_id)
```
we are fine with the standard error message if `entity_id` is not a valid id, right?

python/sdist/amici/numpy.py
```diff
@@ -238,6 +238,25 @@ def __getitem__(self, item: str) -> Union[np.ndarray, ReturnDataPtr,
...
+        else:
+            raise NotImplementedError(
+                f"Subsetting {field} by ID is not implemented or not possible."
+            )
+        return getattr(self, field)[:, col_index]
```
FFroehlich, line 258:
```suggestion
        if field.startswith('s'):
            return getattr(self, field)[:, col_index, :]
        else:
            return getattr(self, field)[:, col_index]
```

FFroehlich (approved):
👍

FFroehlich:
Regarding API, those ids are globally unique, so I would argue that we could avoid passing the `field` kwarg and instead pass sensi=True/False and then have a set of literals to differentiate `x`/`x_ss`/`x0`. 

We may also want to think about adding those ids to rdata such that it isn't required to pass the model as argument.

dweindl:
> Regarding API, those ids are globally unique, so I would argue that we could avoid passing the `field` kwarg and instead pass sensi=True/False and then have a set of literals to differentiate `x`/`x_ss`/`x0`.

P.S.: 

I think this gets weird quickly. For example, how to distinguish `y` and `sigmay`? Then we'd need again some extra argument. 

How about keeping it as it is, but making `field` optional. If field is not provided, it will default to `x` if a state ID was given, `y` if an observable ID was given, ... For anything less basic, one needs to specify `field`.

It could also be extended to `entity_id=Tuple(str, str)` and having `field` default to, for example, `sx` if a state ID and a parameter ID are given.

FFroehlich:
> > Regarding API, those ids are globally unique, so I would argue that we could avoid passing the `field` kwarg and instead pass sensi=True/False and then have a set of literals to differentiate `x`/`x_ss`/`x0`.
> 
> Certainly possible. No strong opinion, but I would think it might be easier for the user to stick to the already existing ReturnData fields instead of having to remember an additional set of literals.
> 
> > We may also want to think about adding those ids to rdata such that it isn't required to pass the model as argument.
> 
> Months ago I was looking into implementing something like that on the C++ side, so that each array is aware of the IDs (without creating copies of them). Would require some API changes though. We could just store the model in `ReturnDataView` or `ReturnData`. On the C++ side, this won't work so well, due to potentially dangling pointers, but just doing it on the Python side could be fine. Only that we'd drag a around an additional Model instance per simulation, which might consume a bit of memory for large numbers of conditions.
> 
> Storing just the IDs/names is possible as well. Again some (but fewer) extra memory required.

What about storing the IDs/model in `ReturnDataView`, but only if reporting mode is set to full?

FFroehlich:
> > Regarding API, those ids are globally unique, so I would argue that we could avoid passing the `field` kwarg and instead pass sensi=True/False and then have a set of literals to differentiate `x`/`x_ss`/`x0`.
> 
> P.S.:
> 
> I think this gets weird quickly. For example, how to distinguish `y` and `sigmay`? Then we'd need again some extra argument.
> 
> How about keeping it as it is, but making `field` optional. If field is not provided, it will default to `x` if a state ID was given, `y` if an observable ID was given, ... For anything less basic, one needs to specify `field`.
> 
> It could also be extended to `entity_id=Tuple(str, str)` and having `field` default to, for example, `sx` if a state ID and a parameter ID are given.

I like the idea of making field optional for `x`/`y` etc

python/sdist/amici/numpy.py
```diff
@@ -238,6 +238,28 @@ def __getitem__(self, item: str) -> Union[np.ndarray, ReturnDataPtr,
...
+        :param field: The requested field, e.g. 'x' for model states
+        :param entity_id: The ID of the model entity that is to be extracted
+            from ``field`` (e.g. a state ID).
+        """
+        if field in {'x', 'x0', 'sx'}:
+            ids = model.getStateIds()
+        elif field in {'w'}:
+            ids = model.getExpressionIds()
+        elif field in {'y', 'sy', 'sigmay'}:
+            ids = model.getObservableIds()
```
FFroehlich, line 255:
```suggestion
            ids = model.getObservableIds()
        elif field in {'sllh'}:
            ids = model.getParameterIds()
```

python/sdist/amici/numpy.py
```diff
@@ -238,6 +238,28 @@ def __getitem__(self, item: str) -> Union[np.ndarray, ReturnDataPtr,
...
+    def by_id(self, model: Model, field: str, entity_id: str) -> np.array:
+        """
+        Get the value of a given field for a named entity.
+
+        :param model: The model from which this ReturnDataView was generated.
+        :param field: The requested field, e.g. 'x' for model states
+        :param entity_id: The ID of the model entity that is to be extracted
+            from ``field`` (e.g. a state ID).
+        """
+        if field in {'x', 'x0', 'sx'}:
```
FFroehlich, line 250:
```suggestion
        if field in {'x', 'x0', 'x_ss', 'sx', 'sx0', 'sx_ss'}:
```

python/sdist/amici/numpy.py
dweindl: Should be `return getattr(self, field)[:, :, col_index]`, right? Addressed differently.

dweindl:
I am not sure whether we should add the same functionality based on entity names (vs IDs). For PySB this would be the only way to meaningfully access the species. On the other hand, names may be non-unique. 

So:
- only handle IDs?
- handle names, and just return the first matching entity?
- handle names, and warn if a name is ambiguous?
- ...?

FFroehlich:
> I am not sure whether we should add the same functionality based on entity names (vs IDs). For PySB this would be the only way to meaningfully access the species. On the other hand, names may be non-unique.
> 
> So:
> 
> * only handle IDs?
> * handle names, and just return the first matching entity?
> * handle names, and warn if a name is ambiguous?
> * ...?



> I am not sure whether we should add the same functionality based on entity names (vs IDs). For PySB this would be the only way to meaningfully access the species. On the other hand, names may be non-unique.
> 
> So:
> 
> * only handle IDs?
> * handle names, and just return the first matching entity?
> * handle names, and warn if a name is ambiguous?
> * ...?

I am overall unhappy with the species ids for pysb, would keep as is here and revisit pysb species ids somewhere else.

PR #1915, opened 2022-12-21
PEtab import: fix handling of fixed parameters for rule targets

dweindl:
Previously, targets of initial assignments or rules could have been selected as fixed parameters, which would lead to errors during model import. Those target parameters should be ignored.

### DIFF
--- a/python/sdist/amici/petab_import.py
+++ b/python/sdist/amici/petab_import.py
@@ -176,6 +176,13 @@ def get_fixed_parameters(
                            " model. Ignoring.")
             fixed_parameters.remove(fixed_parameter)
 
+    # exclude targets of rules or initial assignments
+    for fixed_parameter in fixed_parameters.copy():
+        # check global parameters
+        if sbml_model.getInitialAssignmentBySymbol(fixed_parameter)\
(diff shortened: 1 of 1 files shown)


FFroehlich:
> Previously, targets of initial assignments or rules could have been selected as fixed parameters, which would lead to errors during model import. Those target parameters should be ignored.

Hmm, but this would be an invalid petab file then, right?

FFroehlich (approved):

FFroehlich:
> > Hmm, but this would be an invalid petab file then, right?
> 
> No, I don't mean that the targets were specified as non-estimated in the parameters table, but that during amici import most non-estimated parameters are turned into fixed parameters for performance reasons. And the current logic tries to also do that for rule-targets, which is wrong. (Problem introduced in #1810)

Ah, I see, so the parameter table doesn't need to account for all parameters in a model and the rest are assumed to remain at their default values, right?

dweindl:
> Ah, I see, so the parameter table doesn't need to account for all parameters in a model and the rest are assumed to remain at their default values, right?

Correct.

PR #1922, opened 2022-12-28
Fix circular import in amici.petab_import_pysb

dweindl:
`python -c "import amici.petab_import_pysb"` failed with something like `cannot import name 'PysbPetabProblem' from partially initialized module 'amici.petab_import_pysb' (most likely due to a circular import) (/opt/hostedtoolcache/Python/3.9.16/x64/lib/python3.9/site-packages/amici/petab_import_pysb.py)`


FFroehlich (approved):
huh, how did that not pop up in tests?

dweindl:
> huh, how did that not pop up in tests?

Depends on the import order. `python -c "import amici.petab_import; import amici.petab_import_pysb"` worked.

PR #1929, opened 2023-01-13
Make `ReturnDataView.status` return `int`

dweindl:


FFroehlich (approved):
what was previously returned, float?

dweindl:
> what was previously returned, float?

Yes.

PR #1932, opened 2023-01-13
Make model module accessible from model instance

dweindl:
This makes it more convenient to create a new (unmodified) model instance, or to find out where the model files are located.

Also adds a snake_case version of `getModel`.


FFroehlich (approved):
👍 neat, check path (test) for one of the test models?

dweindl:
Added a test.

Ideally, this whole thing would be handled by returning an instance of the ModelXXXX Python class by `ModelModule.getModel`, that would already have the module available. But due to the virtual class hierarchy, this didn't seem to work out of the box. Can be improved later on...

PR #1934, opened 2023-01-14
Skip irrelevant frames in stacktraces

dweindl:
* Add optional startlevel for stacktrace
* Fix maxLevel, which should be interpreted relative to the startlevel

(+ some clang-format)

### DIFF
--- a/include/amici/exception.h
+++ b/include/amici/exception.h
@@ -18,8 +18,9 @@ class AmiException : public std::exception {
   public:
     /**
      * @brief Default ctor.
+     * @param first_frame Index of first frame to include
      */
-    AmiException(int const stacklevel = 3);
+    AmiException(int const first_frame = 3);
 
     /**
      * @brief Constructor with printf style interface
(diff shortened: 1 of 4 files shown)


include/amici/exception.h
```diff
@@ -18,8 +18,9 @@ class AmiException : public std::exception {
   public:
     /**
      * @brief Default ctor.
+     * @param first_frame Index of first frame to include
      */
-    AmiException();
+    AmiException(int const first_frame = 3);
```
FFroehlich, line 23: why not use 4 as this is what's used below (or not have any default)?

FFroehlich (approved):

include/amici/exception.h
dweindl: If you do `throw AmiException()`, you need to drop 3 frames to get to the throwing frame.

If you do `throw AmiException("foo")`, you need to drop 4 frames to get to the throwing frame, because this delegates to `AmiException()`.

PR #1939, opened 2023-01-17
RTD: Add run-in-binder links to notebooks

dweindl:
E.g. https://amici--1939.org.readthedocs.build/en/1939/GettingStarted.html

PR #1940, opened 2023-01-17
RTD: simplify amici installation

dweindl:
after https://github.com/AMICI-dev/AMICI/pull/1939, we can simply install it via rtd_requirements.txt


FFroehlich (approved):
👍

PR #1946, opened 2023-01-18
Doc: hints for fixing simulation failures 

dweindl:
see https://amici--1946.org.readthedocs.build/en/1946/example_errors.html#  / https://mybinder.org/v2/gh/AMICI-dev/AMICI/nb_dbg?labpath=documentation/example_errors.ipynb


python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,977 @@
...
+    "        petab_problem.sample_parameter_startpoints(n_starts=1)[0],\n",
+    "    )\n",
+    ")\n",
+    "res = simulate_petab(\n",
+    "    petab_problem=petab_problem, \n",
+    "    amici_model=amici_model,\n",
+    "    problem_parameters=problem_parameters,\n",
+    "    scaled_parameters=True\n",
+    ")\n",
+    "print(\"Status:\", [amici.simulation_status_to_str(rdata.status) for rdata in res[RDATAS]])"
```
FFroehlich, line 90: assert that condition `condition_step_03_0` fails at `t = 3031.8`

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,977 @@
...
+   "source": [
+    "**What happened?**\n",
+    "\n",
+    "AMICI failed to integrate the forward problem. The problem occurred for only one simulation condition, `condition_step_03_0`. The issue occurred at $t = 3031.8$, where the CVODES reached the maximum number of steps.\n",
+    "\n",
+    "**How to address?**\n",
+    "\n",
+    "The number of steps the solver has to take is closely related to the chosen error tolerance. More accurate results, more steps. Therefore, this problem can be solved in two ways:\n",
+    "\n",
+    "1. Increasing the maximum number of steps via [`amici.Solver.setMaxSteps`](https://amici.readthedocs.io/en/latest/generated/amici.amici.Solver.html#amici.amici.Solver.setMaxSteps). Note that increasing the number of allowed steps increase RAM requirements, so don't set this to ridiculously large values in order to avoid this error.\n",
```
FFroehlich, line 106: Also note that setting high values opens the door to extremely long simulation times (that may still result in an error), which is probably the bigger issue in most practical situations.

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,977 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "53e4b822",
+   "metadata": {},
+   "source": [
+    "**What happened?**\n",
+    "\n",
+    "The forward simulation failed because AMICI the solver exceeded the maximum number of steps. Unlike in the previous case of  `mxstep steps taken before reaching tout` (see above), here we got several additional warnings that the current step size $h$ is numerically zero.\n",
```
FFroehlich, line 199: maybe note that having the current step size $h$ equal to zero is not an issue by itself and that the solver can recover from that. But here the solver does get stuck which results in exceeding the max step number.

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,977 @@
...
+    "        petab_problem.sample_parameter_startpoints(n_starts=1)[0],\n",
+    "    )\n",
+    ")\n",
+    "res = simulate_petab(\n",
+    "    petab_problem=petab_problem, \n",
+    "    amici_model=amici_model,\n",
+    "    problem_parameters=problem_parameters,\n",
+    "    scaled_parameters=True\n",
+    ")\n",
+    "print(\"Status:\", [amici.simulation_status_to_str(rdata.status) for rdata in res[RDATAS]])"
```
FFroehlich, line 287: add assert status code?

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,977 @@
...
```
FFroehlich, line 189: assert status code?

python/examples/example_errors.ipynb
FFroehlich, line 90: assert status?

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,977 @@
...
```
FFroehlich, line 392: add assert?

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,977 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "02f2dd5d",
+   "metadata": {},
+   "source": [
+    "**How to address?**\n",
+    "\n",
+    "The error message already suggests a fix for this situation, so let's try increasing the relative tolerance:"
```
FFroehlich, line 412: that's good maybe we should do the same for other errors?

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,977 @@
...
+    "    )\n",
+    ")\n",
+    "res = simulate_petab(\n",
+    "    petab_problem=petab_problem, \n",
+    "    amici_model=amici_model,\n",
+    "    problem_parameters=problem_parameters,\n",
+    "    scaled_parameters=True,\n",
+    "    solver=amici_solver\n",
+    ")\n",
+    "print(\"Status:\", [amici.simulation_status_to_str(rdata.status) for rdata in res[RDATAS]])"
```
FFroehlich, line 439: assert status?

python/examples/example_errors.ipynb
dweindl: asserted status codes. 

i think the exact time point may vary slightly on different systems, and is currently not easily accessible. maybe something to add...

python/examples/example_errors.ipynb
dweindl: If we can make good suggestions, that would for sure be helpful.

python/examples/example_errors.ipynb
dweindl: Agreed, runtime should me mentioned. 

I was referring more to changing that to something like 10**10, which will afaik require lots of RAM, even if fewer steps are taken.

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,991 @@
...
+   "metadata": {},
+   "source": [
+    "**What happened?**\n",
+    "\n",
+    "The forward simulation failed because AMICI encountered a `NaN` value when simulating condition `model1_data1`.\n",
+    "Then `NaN`s occurred in $\\dot x$ and $w$ (model expressions, such as reaction fluxes or assignment rules). Furthermore, the failure occurred at the first call, so at $t = t_0$ (here: $t = 0$).\n",
+    "\n",
+    "**How to address?**\n",
+    "\n",
+    "The `NaN` in $\\dot x$ is most likely a consequence of the one in $w$. (A subset of) the dependency tree looks something like:\n",
```
FFroehlich, line 497: J, xdot and y also all depend on expressions w

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,991 @@
...
+   "source": [
+    "print(f\"{x0['Z_state']**unscaled_parameter['n_par'] + unscaled_parameter['Kd']**unscaled_parameter['n_par']=}\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6616c2ff",
+   "metadata": {},
+   "source": [
+    "Indeed... No way we can fix this for the given model.\n",
```
FFroehlich, line 603:
```suggestion
    "Indeed, no way we can fix this for the given model.\n",
```

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,991 @@
...
+    "This was most likely a very unrealistic parameter value, originating from a too high upper parameter bound for `n_par`.\n",
```
FFroehlich, line 604:
```suggestion
    "This was most likely an unrealistic parameter value, originating from a too high upper parameter bound for `n_par`.\n",
```

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,991 @@
...
+    "rdata = res[RDATAS][0]\n",
+    "np.linalg.cond(rdata.J)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "da131d3e",
+   "metadata": {},
+   "source": [
+    "Indeed, the condition number shows that the Jacobian is numerically singular. This is usually the result of conserved quantities in the model.\n",
```
FFroehlich, line 690:
```suggestion
    "Indeed, the condition number shows that the Jacobian is numerically singular. If this happens consistently, it is usually due to conserved quantities in the model.\n",
```

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,991 @@
...
+    "\n",
+    "There are two ways we can address that:\n",
+    "\n",
+    "1. Use numerical integration to compute sensitivities, for which a singular Jacobian is not an issue. This is usually slower, though.\n",
```
FFroehlich, line 694:
```suggestion
    "1. Use numerical integration to compute sensitivities, for which a singular Jacobian is not an issue. This is, usually, slower, though.\n",
```

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,991 @@
...
+    "\n",
+    "Always look for the most basic (furthest up) model quantities.\n",
+    "In cases where there non-finite values occur in expressions further down, rerunning the simulation after calling `Model.setAlwaysCheckFinite(True)` may give some further hints on where the issue originates.\n",
+    "\n",
+    "The `NaN` in $w$ occurred for `flux_v7_v_6` (see error log), i.e. when computing the reaction flux for reaction `v7_v_6`. As $w$ only depends on $(t, p, k, x)$ and no non-finite values have been reported for those, the issue has to be in the respective flux equation.\n",
```
FFroehlich, line 504:
```suggestion
    "The `NaN` in $w$ occurred for `flux_v7_v_6` (see error log), i.e., when computing the reaction flux for reaction `v7_v_6`. As $w$ only depends on $(t, p, k, x)$ and no non-finite values have been reported for those, the issue has to be in the respective flux equation.\n",
```

python/examples/example_errors.ipynb
```diff
@@ -0,0 +1,991 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0713b830",
+   "metadata": {},
+   "source": [
+    "**What happened?**\n",
+    "\n",
+    "All given experimental conditions require pre-equilibration, i.e. finding a steady state. AMICI first tries find a steady state using the Newton solver, if that fails, it tries simulating until steady state, if that also failes, it tries the Newton solver from the end of the simulation. In this case, all three failed. Neither Newton's method nor simulation yielded a steadystate satisfying the required tolerances.\n",
```
FFroehlich, line 805:
```suggestion
    "All given experimental conditions require pre-equilibration, i.e., finding a steady state. AMICI first tries find a steady state using the Newton solver, if that fails, it tries simulating until steady state, if that also failes, it tries the Newton solver from the end of the simulation. In this case, all three failed. Neither Newton's method nor simulation yielded a steadystate satisfying the required tolerances.\n",
```

python/examples/example_errors.ipynb
dweindl: Thanks. And a couple more connections were missing.

FFroehlich (approved):
👍

PR #1955, opened 2023-01-24
Always check for finite x0

### LINKED ISSUE #1952: Always check for finite x0
Initial states should always be checked for finite values, not only with `Model.setAlwaysCheckFinite(True)`. As this only happens once per simulation, the overhead is negligible.

dweindl:
Initial states should always be checked for finite values, not only with `Model.setAlwaysCheckFinite(True)`. As this only happens once per simulation, the overhead is negligible.

Closes #1952

### DIFF
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -87,10 +87,10 @@ const std::map<ModelQuantity, std::string> model_quantity_to_str {
  * @param id_name string indicating whether name or id was specified
  * @return value of the selected parameter
  */
-static realtype getValueById(std::vector<std::string> const &ids,
-                             std::vector<realtype> const &values,
-                             std::string const &id, const char *variable_name,
-                             const char *id_name) {
+static realtype getValueById(
+    std::vector<std::string> const& ids, std::vector<realtype> const& values,
+    std::string const& id, char const* variable_name, char const* id_name
+) {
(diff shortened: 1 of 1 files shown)


FFroehlich:
this is an abysmal pain to look at, can we separate clang formats from actual changes?

src/model.cpp
```diff
@@ -1662,11 +1685,7 @@ void Model::fx0(AmiVector &x) {
     ftotal_cl(state_.total_cl.data(), derived_state_.x_rdata_.data(),
               state_.unscaledParameters.data(),
               state_.fixedParameters.data());
-
-    if (always_check_finite_) {
-        checkFinite(derived_state_.x_rdata_, ModelQuantity::x0_rdata);
-        checkFinite(x.getVector(), ModelQuantity::x0);
-    }
+    checkFinite(derived_state_.x_rdata_, ModelQuantity::x0_rdata);
```
dweindl, line 1688: Changed

dweindl:
> this is an abysmal pain to look at, can we separate clang formats from actual changes?

oops. didn't notice there was so much. I'll just clang-format everything later, then that issue should be solved for the near future.

FFroehlich (approved):
👍

PR #1960, opened 2023-01-26
Change default interpolation type from hermite to polynomial

dweindl:
According to [CVODES docs](https://sundials.readthedocs.io/en/latest/cvodes/Mathematics_link.html#checkpointing-scheme), polynomial should be the better choice:

> The variable-degree polynomial is more memory-efficient (it requires only half of the memory storage of the cubic Hermite interpolation) and is more accurate.


FFroehlich (approved):
👍

PR #1964, opened 2023-01-29
Doc: Instructions for AMICI with singularity

dweindl:
* how to create a container
* make Dockerfile compatible with spython/singularity
* mv docker/ container/


FFroehlich (approved):
👍

PR #1968, opened 2023-01-31
Fix PytestReturnNotNoneWarning in test_conserved_quantities_demartino.py

dweindl:
Fixes
```
  C:\hostedtoolcache\windows\Python\3.8.10\x64\lib\site-packages\_pytest\python.py:199: PytestReturnNotNoneWarning: Expected None, but python/tests/test_conserved_quantities_demartino.py::test_compute_moiety_conservation_laws_demartino2014 returned 7.324800299999993, which will be an error in a future version of pytest.  Did you mean to use `assert` instead of `return`?
```


FFroehlich (approved):
👍

PR #1972, opened 2023-01-31
Fix MSVC builds / remove -W* flags

dweindl:
Removes `-W` compiler flags on Windows, which aren't understood by MSVC.

Fixes `cl : Command line error D8021 : invalid numeric argument '/Werror'` (which for some reason currently only occurs on branch `adjoint_event` :man_shrugging:)

See also https://learn.microsoft.com/en-us/cpp/build/reference/wx-treat-linker-warnings-as-errors?view=msvc-170


FFroehlich (approved):
🤷

PR #1973, opened 2023-02-02
Fix assignmentRules2observables - skip non-assignment-rule targets

dweindl:
The filter function should only be called on AssignmentRule targets. Currently it's called on all parameters.


FFroehlich (approved):
👍

PR #1984, opened 2023-02-11
Change default ReturnData::status to AMICI_NOT_RUN

### LINKED ISSUE #1949: Fix ReturnData::status for empty ReturnData
If a simulation fails `runAmiciSimulations(...., failfast=True)` , an empty `ReturnData`  with `status=AMICI_SUCCESS` is returned for the skipped conditions, . This is rather confusing. Either we should add a new status, e.g., `NOT_RUN`, or we just return `None` instead of `ReturnData` for these cases.

dweindl:
So far, if a simulation with `runAmiciSimulations(...., failfast=True)` failed, an empty `ReturnData` with `status=AMICI_SUCCESS` was returned for the skipped conditions. This was rather confusing.

This adds a new status: `AMICI_NOT_RUN`.

Fixes #1949


FFroehlich (approved):
👍

PR #1985, opened 2023-02-11
Print stacktraces only with debug logging level

### LINKED ISSUE #1943: Print stacktraces only with debug logging level
Requires splitting the messages

dweindl:
Closes #1943


FFroehlich (approved):
👍  nice, easier than anticipated :D

PR #1988, opened 2023-02-13
Reduce time-tracking overhead

dweindl:
* Don't check `clock()` on every rhs evaluation, but only every 500.
* Avoid `clock()` completely, if there is no max time specified.

### DIFF
--- a/sbml_test_models/amici_models/piecewise_with_boolean_operations/piecewise_with_boolean_operations_xB.h
+++ b/sbml_test_models/amici_models/piecewise_with_boolean_operations/piecewise_with_boolean_operations_xB.h
@@ -1,0 +1,1 @@
+#define xB0 xB[0]
--- a/sbml_test_models/amici_models/piecewise_with_boolean_operations/piecewise_with_boolean_operations_xB.h
+++ b/sbml_test_models/amici_models/piecewise_with_boolean_operations/piecewise_with_boolean_operations_xB.h
@@ -1,1 +1,0 @@
-#define xB0 xB[0]
--- a/src/solver.cpp
+++ b/src/solver.cpp
@@ -864,7 +864,12 @@ void Solver::startTimer() const
 
 bool Solver::timeExceeded() const
 {
-    auto cputime_exceed = static_cast<double>(std::clock() - starttime_) / CLOCKS_PER_SEC;
+    // 0 means infinte time, don't call `clock()` then (expensive)
+    if(maxtime_.count() == 0)
+        return false;
+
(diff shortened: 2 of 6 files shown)


sbml_test_models/amici_models/piecewise_with_boolean_operations/piecewise_with_boolean_operations_xB.h
```diff
@@ -0,0 +1 @@
+#define xB0 xB[0]
```
FFroehlich, line 1: 🤔

src/solver.cpp
```diff
@@ -864,7 +864,12 @@ void Solver::startTimer() const
 
 bool Solver::timeExceeded() const
 {
-    auto cputime_exceed = static_cast<double>(std::clock() - starttime_) / CLOCKS_PER_SEC;
+    // 0 means infinte time, don't call `clock()` then (expensive)
```
FFroehlich, line 867:
```suggestion
    // 0 means infinite time, don't call `clock()` then (expensive)
```

src/solver_cvodes.cpp
```diff
@@ -1036,8 +1036,14 @@ static int fxdot(realtype t, N_Vector x, N_Vector xdot, void *user_data) {
     auto solver = dynamic_cast<CVodeSolver const *>(typed_udata->second);
     Expects(model);
 
-    if(solver->timeExceeded()) {
-        return AMICI_MAX_TIME_EXCEEDED;
+    // Avoid expensive syscalls by checking only every N rhs evaluations
+    static int eval_counter = 0;
+    int const interval = 1000;
+    if (++eval_counter / interval) {
```
FFroehlich, line 1042: What about rather extending `solver->timeExceeded()` by adding the counter as private member of solver and checking against that in every call?

FFroehlich (approved):

src/solver_cvodes.cpp
dweindl: Thought about it, but it wouldn't work with 

https://github.com/AMICI-dev/AMICI/blob/eba2cadcd4f9439c13cd19644351c6c1b7673c68/src/amici.cpp#L148

There we want to check on every call. Would have to add some force_check=True argument or a separate method.

FFroehlich: > Thought about it, but it wouldn't work with
> 
> https://github.com/AMICI-dev/AMICI/blob/eba2cadcd4f9439c13cd19644351c6c1b7673c68/src/amici.cpp#L148
> 
> There we want to check on every call. Would have to add some force_check=True argument or a separate method.

I think I would prefer that.

Issue #1913, opened 2022-12-14
Reported CPU times are incorrect when used with multi-threading 


dweindl:
Currently, the value of `Solver::getMaxTime()` is checked against `std::chrono::system_clock`, which measures wall time. This is probably not what we want and can cause issues if a program is paused or the cpu is oversubscribed. Should be changed to `std::clock`. The other time measurements reported in `ReturnData` are already using `clock`.

dweindl:
Just using `clock()` won't do the job for parallel simulations. Would need separate Windows- and POSIX-code to get thread-specific CPU time. [boost::chrono::thread_clock](https://www.boost.org/doc/libs/1_73_0/doc/html/chrono/reference.html#chrono.reference.other_clocks.thread_clock_hpp.thread_clock) would work, but we don't want to depend on boost. Probably not worth the trouble at this point.

dweindl:
Well, that also means that all currently reported cpu times obtained from from `runAmiciSimulations(..., nthreads=N)`, `N>1`  are probably garbage. So we should do *something* about it...

PR #1989, opened 2023-02-14
Doc: Add warning regarding cpu time + multi-threading

dweindl:
see https://github.com/AMICI-dev/AMICI/issues/1913#issuecomment-1429265168


FFroehlich (approved):
👍

PR #1992, opened 2023-02-15
CMake-based Python extension builds

### LINKED ISSUE #721: Use CMake or enscons for python package build/installation
[enscons](https://pypi.org/project/enscons/) looks like a viable replacement for setuptools, which still is pain in the ass. This might make it easier to fix #716 #700.

Any experiences with that?

dweindl:
Building non-trivial extensions with setuptools alone is a pain. CMake should make everything way easier and simplify finding relevant libraries. This has also potential to solve conda-issues :crossed_fingers: .

Changes:
* CMake-based build of all python extensions, removes any build_clib steps (Closes #721)
* Removes most of the setuptools/distutils hacks
* Removes custom library-finding-logic that is now handled by CMake
* Enables/disables some additional warnings
* More relocatable CMake exports

A few modifications to SUNDIALS / SuiteSparse were necessary (all marked by some comment containing `AMICI` in the respective files):
* I wasn't able to convince SuiteSparse to find OpenBLAS on windows. Any attempt of doing so has been disabled, and AMICI's logic is used. BLAS with 64bit integer type is assumed.
* Sundials' FindKLU was modified to find static SuiteSparse libraries (or any SuiteSparse libraries at all...)


dweindl:
TODO

- [ ] default to sequential compilation (?)
- [ ] update docs (installation, env vars, parallel compilation, ...)
- [ ] `--no-clibs` options  - still needed?
- [ ] cleanup

scripts/installAmiciSource.sh
```diff
@@ -27,8 +27,10 @@ else
     source ${AMICI_PATH}/build/venv/bin/activate
 fi
 
-pip install --upgrade pip pkgconfig scipy matplotlib coverage pytest pytest-cov
-pip install git+https://github.com/pysb/pysb # pin to develop to fix sympy compatibility
 pip install -U "setuptools<64"
+pip install --upgrade pip wheel
+pip install --upgrade pip scipy matplotlib coverage pytest \
+  pytest-cov cmake_build_extension numpy
+pip install git+https://github.com/pysb/pysb # pin to develop to fix sympy compatibility
```
FFroehlich, line 34: I don't think we need this pin anymore

FFroehlich (approved):
👍  nice!

PR #1996, opened 2023-02-16
Jax example notebook

FFroehlich:
requires #1548


python/examples/example_jax/ExampleJax.ipynb
```diff
@@ -0,0 +1,1087 @@
...
+      "remote: Total 336 (delta 88), reused 216 (delta 39), pack-reused 0\u001b[K\n",
+      "Receiving objects: 100% (336/336), 2.11 MiB | 7.48 MiB/s, done.\n",
+      "Resolving deltas: 100% (88/88), done.\n"
+     ]
+    }
+   ],
+   "source": [
+    "!git clone --depth 1 https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab.git tmp/benchmark-models || (cd tmp/benchmark-models && git pull)\n",
+    "from pathlib import Path\n",
+    "folder_base = Path('.') / \"tmp\" / \"benchmark-models\" / \"Benchmark-Models\""
```
dweindl, line 61: For an individual model, you can also use the URL of the yaml file in Problem.from_yaml to just download that problem.

python/examples/example_jax/ExampleJax.ipynb
```diff
@@ -0,0 +1,1087 @@
...
+     "output_type": "stream",
+     "text": [
+      "2023-02-16 12:37:31.673 - amici.petab_import - INFO - Finished Importing PEtab model                (1.36E+01s)\n",
+      "2023-02-16 12:37:31.684 - amici.petab_import - INFO - Successfully loaded model Boehm_JProteomeRes2014 from /Users/fabian/Documents/projects/AMICI/documentation/amici_models/Boehm_JProteomeRes2014.\n"
+     ]
+    }
+   ],
+   "source": [
+    "from amici.petab_import import import_petab_problem\n",
+    "amici_model = import_petab_problem(petab_problem, force_compile=True)"
```
dweindl, line 573:
```suggestion
    "amici_model = import_petab_problem(petab_problem, force_compile=True, verbose=False)"
```

python/examples/example_jax/ExampleJax.ipynb
```diff
@@ -0,0 +1,1087 @@
...
+      "ld: warning: -undefined dynamic_lookup may not work with chained fixups\n",
```
dweindl, line 558: ?

python/examples/example_jax/ExampleJax.ipynb
```diff
@@ -0,0 +1,1087 @@
...
+   "source": [
+    "# JAX implementation"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e2ef051a",
+   "metadata": {},
+   "source": [
+    "For full jax support, we would have to implement a new [primitive](https://jax.readthedocs.io/en/latest/notebooks/How_JAX_primitives_work.html), which would require quite a bit of engineering, and in the end wouldn't add much benefit since AMICI can't run on GPUs. Instead will interface AMICI using the experimental jax module [`host_callback`](https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html). "
```
dweindl, line 589: verbatim links (`[``host_callback``]`) will probably not render well in the sphinx doc :hankey:

dweindl, line 589:
```suggestion
    "For full jax support, we would have to implement a new [primitive](https://jax.readthedocs.io/en/latest/notebooks/How_JAX_primitives_work.html), which would require quite a bit of engineering, and in the end wouldn't add much benefit since AMICI can't run on GPUs. Instead will interface AMICI using the experimental jax module [host_callback](https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html). "
```

python/examples/example_jax/ExampleJax.ipynb
```diff
@@ -0,0 +1,1087 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6bbf2f06",
+   "metadata": {},
+   "source": [
+    "To do so, we define a base function that only takes a single argument (the parameters) and runs simulation using petab via [`simulate_petab`](https://amici.readthedocs.io/en/latest/generated/amici.petab_objective.html#amici.petab_objective.simulate_petab). To enable gradient computation later on, we create a solver object and set the sensitivity order to first order and pass it to `simulate_petab`. Moreover, `simulate_petab` expects a dictionary of parameters, so we create a dictionary using the free parameter ids from the petab problem. As we want to implement parameter transformation in JAX, we disable parameter scaling in petab by passing `scaled_parameters=True`."
```
dweindl, line 597:
```suggestion
    "To do so, we define a base function that only takes a single argument (the parameters) and runs simulation using petab via [simulate_petab](https://amici.readthedocs.io/en/latest/generated/amici.petab_objective.html#amici.petab_objective.simulate_petab). To enable gradient computation later on, we create a solver object and set the sensitivity order to first order and pass it to `simulate_petab`. Moreover, `simulate_petab` expects a dictionary of parameters, so we create a dictionary using the free parameter ids from the petab problem. As we want to implement parameter transformation in JAX, we disable parameter scaling in petab by passing `scaled_parameters=True`."
```

python/examples/example_jax/ExampleJax.ipynb
```diff
@@ -0,0 +1,1087 @@
...
+    "        solver=amici_solver,\n",
+    "    )"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6f6201e8",
+   "metadata": {},
+   "source": [
+    "Now we can use this base function to create two functions separate functions that compute the log-likelihood (`llh`) and it's gradient (`sllh`) in two individual routines. Note that, as we are using the same base function here, the log-likelihood computation will also run with sensitivities which is not necessary and will add some overhead. This is only out of convenience and should be fixed in an application where efficiency is important."
```
dweindl, line 627:
```suggestion
    "Now we can use this base function to create two separate functions that compute the log-likelihood (`llh`) and its gradient (`sllh`) in two individual routines. Note that, as we are using the same base function here, the log-likelihood computation will also run with sensitivities which is not necessary and will add some overhead. This is only out of convenience and should be fixed in an application where efficiency is important."
```

python/examples/example_jax/ExampleJax.ipynb
```diff
@@ -0,0 +1,1087 @@
...
+    "pd.DataFrame(index=sllh.keys(), data=dict(amici=sllh.values(), jax=np.asarray(sllh_jax)))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "4b00dcb2",
+   "metadata": {},
+   "source": [
+    "We see quite some differences in the gradient calculation. The primary reason is that running JAX in default configuration will use float32 precision for the parameters that are passed to AMICI, which uses float64, and the derivative of the parameter transformation \n",
+    "As AMICI simulations that run on the CPU are the most expensive operation, there is barely any tradeoff for using float32 vs float64 in JAX. Therefore we configure JAX to use float64 instead and rerun simulations."
```
dweindl, line 917:
```suggestion
    "As AMICI simulations that run on the CPU are the most expensive operation, there is barely any tradeoff for using float32 vs. float64 in JAX. Therefore, we configure JAX to use float64 instead and rerun simulations."
```

python/examples/example_jax/ExampleJax.ipynb
```diff
@@ -0,0 +1,1087 @@
...
+       "sd_rSTAT5A_rel       -2.643776e-03 -2.643703e-03"
+      ]
+     },
+     "execution_count": 17,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "pd.DataFrame(index=sllh.keys(), data=dict(amici=sllh.values(), jax=np.asarray(sllh_jax)))"
```
dweindl, line 1062: Add column with relative difference?

dweindl (approved):
:+1:

FFroehlich:
add notebook to `binder/overview.ipynb`

PR #2008, opened 2023-02-21
Fix error logging in steadystate solver

### LINKED ISSUE #1976: Uninformative error if simulation-steady-state solver does not converge
Leads to errors like
```
2023-02-03 13:48:24.088 - amici.swig_wrappers - ERROR - [FORWARD_FAILURE] AMICI forward simulation failed at t = 4.14971e+199: AMICI failed to integrate the forward problem
```
which are difficult to understand by users, instead of processing the error in 
https://github.com/AMICI-dev/AMICI/blob/ed36b2daa214029487588a50b77447d3dac76f78/src/steadystateproblem.cpp#L364

FFroehlich:
fixes #1976

Pretty much completely at loss why there is a difference in error messages for running https://github.com/AMICI-dev/AMICI/blob/master/src/steadystateproblem.cpp#L168 vs  https://github.com/AMICI-dev/AMICI/blob/master/src/steadystateproblem.cpp#L171

### DIFF
--- a/python/tests/test_preequilibration.py
+++ b/python/tests/test_preequilibration.py
@@ -414,17 +414,20 @@ def test_simulation_errors(preeq_fixture):
     model.setSteadyStateSensitivityMode(amici.SteadyStateSensitivityMode.integrationOnly)
     solver.setMaxSteps(1)
 
+    # exceeded maxsteps
     # preeq & posteq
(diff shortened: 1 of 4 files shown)


dweindl (approved):
:+1:

PR #2009, opened 2023-02-21
SWIG __repr__s for different templated vector classes

### LINKED ISSUE #2007: __repr__ for DoubleVector etc
should be easy to implement by using `np.asarray`

FFroehlich:
fixes #2007


dweindl (approved):
:+1: nice

PR #2013, opened 2023-02-22
Matlab: If mex fails, print mex arguments

### LINKED ISSUE #913: Matlab: If mex fails, print mex arguments
For all mex calls, catch exception, print mex invocation, rethrow. Will make bug reports way more informative and simplify debugging.

dweindl:
For all mex calls, catch exception, print mex invocation, rethrow. Will make bug reports way more informative and simplify debugging.

Resolves #913


FFroehlich (approved):
👍

PR #2015, opened 2023-02-23
Remove model name prefix in generated model files

dweindl:
Helps to avoid issues with too long paths on Windows :hankey:


models/model_calvetti/model_calvetti.h
```diff
@@ -70,7 +70,7 @@ class Model_model_calvetti : public amici::Model_DAE {
 
     amici::Model* clone() const override { return new Model_model_calvetti(*this); };
 
-    std::string getAmiciCommit() const override { return "b61abb6cf09df7b484efe45bb8e378ac07d2b33b"; };
+    std::string getAmiciCommit() const override { return ""; };
```
FFroehlich, line 73: 🤔

FFroehlich (approved):

models/model_calvetti/model_calvetti.h
dweindl: Good catch. Doesn't work with git worktrees...

PR #2022, opened 2023-02-25
Extract timing code

dweindl:
This should be more readable, and will make it easier to change to a thread-specific clock.


FFroehlich (approved):
👍

PR #2030, opened 2023-03-03
don't use deprecated code

FFroehlich:
Not sure what this is about? is parameter table intentionally not passed to import model?


dweindl:
> Not sure what this is about? is parameter table intentionally not passed to `import_model`?

Until recently, the parameters table wasn't used for model import. But I don't see why including it would break anything. The failures above were related to the restructured petab test suite.

PR #2031, opened 2023-03-08
Fix HDF5 ambiguous overload 

dweindl:
reported by @dilpath: 
> Fails with HDF5 1.14.0, 1.13.3. Works with 1.12.1

```
      amici/src/hdf5.cpp:949:65: error: call of overloaded ‘H5File(const char*, unsigned int, int)’ is ambiguous
        949 |     H5::H5File file(hdffile.c_str(), H5F_ACC_RDONLY, H5P_DEFAULT);
```


dilpath (approved):
:+1:

PR #2034, opened 2023-03-08
Smith benchmark and SBML initialization fix

FFroehlich:
requires https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/pull/180


python/sdist/amici/petab_import.py
```diff
@@ -671,6 +671,11 @@ def import_model_sbml(
                          'initial values specified by the PEtab problem.')
         formula = f'{PREEQ_INDICATOR_ID} * {init_par_id_preeq} ' \
                   f'+ (1 - {PREEQ_INDICATOR_ID}) * {init_par_id_sim}'
+        if sbml_model.getSpecies(assignee_id).getHasOnlySubstanceUnits():
+            compartment = sbml_model.getCompartment(
+                sbml_model.getSpecies(assignee_id).getCompartment()
+            ).getId()
+            formula = f'({formula}) / {compartment}'
```
dweindl, line 678: Hm, somehow I thought species with hasOnlySubstanceUnits=True should have amounts here, not concentrations.

FFroehlich: I was also a bit confused, but the initial assignments processing code in amici will assume the _formula_ yields a concentration which then needs to be converted to an amount to assign the _species_.

dweindl: Hm, but doesn't that go against the SBML spec?

SBML L3V2 section 4.8.5:
> • In the case of a species, an InitialAssignment sets the referenced species’ initial quantity (concentration or
amount) to the value determined by the formula in math. The unit associated with the value produced by
the math formula should be equal to the unit associated with the species’ quantity. (See Section 4.6.5 on
p. 51 for an explanation of how a species’ quantity is determined.)

4.6.5
> The interpretation is controlled by
the attribute hasOnlySubstanceUnits. If the attribute has the value “false”, then the unit of measurement
associated with the value of the species is {unit of amount}/{unit of size} (i.e., concentration or density). If
hasOnlySubstanceUnits has the value “true”, then the value is interpreted as having a unit of amount only.

tests/benchmark-models/benchmark_models.yaml
```diff
@@ -96,6 +96,13 @@ Sneyd_PNAS2002:
   t_adj: 0.5
   note: benchmark collection reference ignores factor 1/2
 
+Smith_BMCSystBiol2013
```
dweindl, line 99:
```suggestion
Smith_BMCSystBiol2013:
```

dweindl (approved):
:+1:

PR #2035, opened 2023-04-05
Fix Equilibration cpu time computation

plakrisenko:
add time of forward sensitivities computation by solving a linear system (`computeNewtonSensis`) to pre(post)eq_cpu_time


FFroehlich (approved):
👍  thank you!

PR #2038, opened 2023-04-13
Fix doc: inherited members

dweindl:
Closes #2026

[old](https://amici.readthedocs.io/en/develop/generated/amici.de_model.html#module-amici.de_model) [new](https://amici--2038.org.readthedocs.build/en/2038/generated/amici.de_model.html#module-amici.de_model)


FFroehlich (approved):
👍

PR #2039, opened 2023-04-13
Require Python>=3.9

dweindl:
see https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule


FFroehlich:
Latest iPhython no longer supports python 3.8

FFroehlich (approved):

FFroehlich:
needs updated requirements

PR #2051, opened 2023-04-18
#2050: rename measurement column to simulation

### LINKED ISSUE #2050: `PetabSimulator.simulate` should produce simulation data frame
after running 

```python
    petab_problem = Problem.from_yaml(args.yaml_file_name)
    simulator = PetabSimulator(petab_problem)
    sim_df = simulator.simulate()
    sim_df.to_csv(args.output_file, sep='\t', index=False)
```

I expected to get a simulation data frame, that I could use with `petab.visualize`. Specifically there should be a `simulation` column in the file. 

Instead there was a `measurement` column in the file, and `petab.visualize` complained with a KeyError when accessing the `simulation` column. 

**To Reproduce**
*Steps to reproduce the behavior*
*Ideally include minimal code examples here*

see above

 - OS and version: [e.g. Ubuntu, iOS, Windows]
 - AMICI interface: [e.g. Python, Matlab, C++]
 - AMICI version:
 - Additional information: Compiler name and version used, Python/Matlab version, ...

**How to fix**
Do you know how to resolve the problem?
yes, rename the measurement column to 'simulation' before returning from PetabSimulator.simulate. 

Can you submit a pull request?

fbergmann:
This fixes #2050, by just renaming the measurement column before returning the data frame.

### DIFF
--- a/python/sdist/amici/petab_simulate.py
+++ b/python/sdist/amici/petab_simulate.py
@@ -84,9 +84,12 @@ class PetabSimulator(petab.simulate.Simulator):
                 SensitivityMethod_none)
 
         result = _subset_call(simulate_petab, kwargs)
-        return rdatas_to_measurement_df(result[RDATAS],
+        df = rdatas_to_measurement_df(result[RDATAS],
                                         self.amici_model,
                                         self.petab_problem.measurement_df)
+        # rename measurement column to simulation
+        df = df.rename(columns={"measurement": "simulation"})
+        return df
 
 
 def _subset_call(method: Callable, kwargs: dict):


dweindl:
Thanks, @fbergmann. I think there are different use cases for this function: Some use it to generate the simulation table, some use it to generate artificial measurements. While I agree, that for a `PetabSimulator`, `simulation` seems to be the least surprising default, I am wondering if we should better add an option to choose between `simulation` / `measurement`.

CC @dilpath

dweindl:
Waiting for https://github.com/PEtab-dev/libpetab-python/pull/199

dweindl:
Fixed via https://github.com/PEtab-dev/libpetab-python/pull/199 (petab>=0.2.0)

Issue #2012, opened 2023-02-22
Update or remove mingw/Windows installation instructions


dweindl:
Terribly outdated. Most users seem to use MSVC anyways.

PR #2053, opened 2023-04-21
Doc: Update Windows/MSVC installation instructions

Podde1:
Updated the description for installation on Windows following installation of AMICI, on Windows 10 running python 3.9. Partly relating to #2012.
- Added cmake to required softwares
- Added a notation that `compileBLAS.cmd`  needs to point to the Visual Studio installation
- Paths to the OpenBLAS installation were updated to be able to set the environmental variables correctly
- Updated faulty paths for OpenBLAS enviromental variables
- Updated faulty paths for `AMICI_DLL_DIRS` enviromental variable

Additionally, added a check in `amici.i` if `AMICI_DLL_DIRS` enviromental variable is present

### DIFF
--- a/documentation/python_installation.rst
+++ b/documentation/python_installation.rst
@@ -116,6 +116,8 @@ need to be included:
   This is part of multiple packages, including Desktop Development with C++.
 * Windows Universal C Runtime.
   This is an individual component and installs some DLLs that we need.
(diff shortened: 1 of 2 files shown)


dweindl (commented):
Great, thanks!

PR #2058, opened 2023-04-25
Binder Dockerfile

### LINKED ISSUE #2054: Binder & c++17
Binder insists on using ubuntu18.04 as base image, which comes with a rather old g++. AMICI installations fails on `std::reduce`.
We need to check if the available clang better supports c++17, or need to move to a Dockerfile-based build, or hope that Binder/repo2docker becomes more configurable...

dweindl:
Change Dockerfile-based Binder configuration.

The previous setup was cleaner, but repo2docker insists on a 5 year old base image that doesn't let us install a C++17-ready compiler...

Fixes #2054 

[Demo](https://mybinder.org/v2/gh/AMICI-dev/AMICI/binder_dockerfile?labpath=documentation/ExampleSteadystate.ipynb)


FFroehlich (approved):
👍

Issue #2061, opened 2023-04-26
Support execution of SBML comp models


matthiaskoenig:
Currently, AMICI raises an error if the `comp` package appears in models. This should be a warning or there should be an option to force execution with comp. Similar things for the `fbc` package. See example attached which is a completely fine model which can be simulated with roadrunner and copasi, but AMICI complains about it. 

Trying to load the attached `liver_model.xml` 
[icg_liver.zip](https://github.com/AMICI-dev/AMICI/files/11337537/icg_liver.zip)
results in 
```
Traceback (most recent call last):
  File "/home/mkoenig/git/sbmlsim/src/sbmlsim/comparison/example_comparison.py", line 82, in <module>
    simulator = simulator(
  File "/home/mkoenig/git/sbmlsim/src/sbmlsim/comparison/simulate_amici.py", line 31, in __init__
    sbml_importer.sbml2amici(self.mid, model_dir)
  File "/home/mkoenig/.virtualenvs/sbmlsim/lib/python3.9/site-packages/amici/sbml_import.py", line 344, in sbml2amici
    ode_model = self._build_ode_model(
  File "/home/mkoenig/.virtualenvs/sbmlsim/lib/python3.9/site-packages/amici/sbml_import.py", line 416, in _build_ode_model
    self._process_sbml(constant_parameters)
  File "/home/mkoenig/.virtualenvs/sbmlsim/lib/python3.9/site-packages/amici/logging.py", line 199, in wrapper_timer
    rval = func(*args, **kwargs)
  File "/home/mkoenig/.virtualenvs/sbmlsim/lib/python3.9/site-packages/amici/sbml_import.py", line 457, in _process_sbml
    self.check_support()
  File "/home/mkoenig/.virtualenvs/sbmlsim/lib/python3.9/site-packages/amici/sbml_import.py", line 495, in check_support
    raise SBMLException(
amici.sbml_import.SBMLException: Required SBML extension comp is currently not supported!
```

Many comp models can still be simulated, often correctly. Just because the comp package appears in a model does not mean it is not a valid and simulatable model with the SBML core. E.g. comp ports are often used as annotation of interfaces, but do not change the model ODE system.

You could also just call the SBML flattening on the comp model, which is a one-liner in libsbml which converts the model to a valid core model. This is how roadrunner (and I think also COPASI) support comp models. I.e. they detect if the comp package is there (what you already are doing) and flatten the model before execution (this is missing).

[...]

PR #2064, opened 2023-04-27
SBML import: Fix check for required packages

dweindl:
getRequired needs to be checked on XXXSBMLDocumentPlugin, not on XXXPlugin.

Also:
* don't try to sympify empty kinetic laws
* handling of empty SBML test suite settings

Enables importing models using the fbc extension (see #2061).

Tested via the following semantic test cases:
01186
01187
01188
01189
01190
01191
01192
01193
01194
01195
01196
01606
01607
01608
01609
01610
01611
01612
01613
01614
01615
01616
01617
01618
01619
01620
01621
01622
01623
01624
01625
01628
01629
01630


FFroehlich (approved):
Great! Any changes to the number of passed tests?

Issue #2055, opened 2023-04-22
Use `locals` for all sympifications 


dweindl:
Apparently, using `N` in observable expressions causes issues, as it is interpreted as a function by sympy, see https://github.com/ICB-DCM/pyPESTO/issues/1048. It's possible that some `sympify` call lacks `locals`, or that that collection is incomplete. Or maybe we can disable those special names in sympy altogether. Needs further investigation.

dweindl:
See also https://github.com/PEtab-dev/libpetab-python/pull/202

PR #2069, opened 2023-05-08
Fix sympy symbol name clashes during PEtab import

dweindl:
Closes https://github.com/AMICI-dev/AMICI/issues/2055


FFroehlich (approved):

dweindl:
> intentional that this doesn't cover all call to `sympy.sympify`?

Yes. The ones in the SBML and PySB importers should be fine. There, we already gather all model symbols beforehand and pass them as locals.

Issue #1886, opened 2022-11-14
CI: external test files unavailble


dweindl:
These tests regularly fail, because some test data from an unstable URL is required
```
ERROR python/tests/test_conserved_quantities_demartino.py::test_kernel_demartino2014 - urllib.error.URLError: <urlopen error timed out>
ERROR python/tests/test_conserved_quantities_demartino.py::test_fill_demartino2014 - urllib.error.URLError: <urlopen error timed out>
ERROR python/tests/test_conserved_quantities_demartino.py::test_compute_moiety_conservation_laws_demartino2014 - urllib.error.URLError: <urlopen error timed out>
ERROR python/tests/test_conserved_quantities_demartino.py::test_cl_detect_execution_time - urllib.error.URLError: <urlopen error timed out>
```

Tests need to be changed, or the data needs to be uploaded to something more stable.

FFroehlich:
use https://github.com/actions/cache ?

dweindl:
Uploading the respective test files here: 

Supporting Material to https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0100750, © 2014 De Martino et al., used under [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)

* [test-ecoli-met.txt](https://github.com/AMICI-dev/AMICI/files/11430970/test-ecoli-met.txt)
* [DeMartinoDe2014_test-ecoli.dat.gz](https://github.com/AMICI-dev/AMICI/files/11430971/DeMartinoDe2014_test-ecoli.dat.gz)

Original URLs - meanwhile apparently dead - https://chimera.roma1.infn.it/SYSBIO/test-ecoli.dat.gz, https://chimera.roma1.infn.it/SYSBIO/test-ecoli-met.txt.

dweindl:
Solved

PR #2072, opened 2023-05-09
Fix dead URLs in test_conserved_quantities_demartino.py

dweindl:
Files now stored at https://github.com/AMICI-dev/AMICI/issues/1886#issuecomment-1540027102


FFroehlich (approved):
👍

PR #2074, opened 2023-05-09
Add steady-state computation modes

plakrisenko:


plakrisenko:
Should we also change the name of the `SteadyStateSensitivityMode.newtonOnly`  to smth like `linSolve` as Newton's method is only used for steady-state computation?

FFroehlich:
> Should we also change the name of the `SteadyStateSensitivityMode.newtonOnly` to smth like `linSolve` as Newton's method is only used for steady-state computation?
> 
> In general, there is a lot of "Newton" mentions that are not clear. E.g.
> 
> https://github.com/AMICI-dev/AMICI/blob/master/src/steadystateproblem.cpp#L440 https://github.com/AMICI-dev/AMICI/blob/master/src/steadystateproblem.cpp#L455 https://github.com/AMICI-dev/AMICI/blob/master/include/amici/newton_solver.h#L60

see https://github.com/AMICI-dev/AMICI/issues/2080, after introducing `SteadyStateComputationMode` it probably makes sense to consolidate `SteadyStateSensitivityMode` and `SensitivityMethodPreequilibration` into a single option.

dweindl (commented):
Please add/extend a test that checks that no simulation is performed in case of `newtonOnly`, and that no Newton steps are taken with `integrationOnly`.

python/tests/test_preequilibration.py
```diff
@@ -472,6 +474,55 @@ def test_newton_steadystate_check(preeq_fixture):
...
+
+    assert np.all(rdatas[amici.SteadyStateComputationMode.integrationOnly][
+                      'preeq_status'][0] == [0, 1, 0]) and \
+           rdatas[amici.SteadyStateComputationMode.integrationOnly][
+                      'preeq_numsteps'][0][0] == 0
+
+    assert np.all(rdatas[amici.SteadyStateComputationMode.newtonOnly][
+                      'preeq_status'][0] == [1, 0, 0]) and \
+           rdatas[amici.SteadyStateComputationMode.newtonOnly][
+                      'preeq_numsteps'][0][0] > 0
```
dweindl, line 515: Please split the condition into individual `assert`s, this makes it easier to see what's going wrong.

python/tests/test_preequilibration.py
```diff
@@ -472,6 +474,55 @@ def test_newton_steadystate_check(preeq_fixture):
...
+
+    # assert correct results
+    for variable in ["llh", "sllh", "sx0", "sx_ss", "x_ss"]:
+        assert np.isclose(
```
dweindl, line 519: Please use [`numpy.testing.assert_allclose`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html#numpy.testing.assert_allclose) for more readable output in case of failure.

dweindl (approved):
Thanks, looks good to me.

dweindl:
It looks like we have some problem here (https://github.com/AMICI-dev/AMICI/actions/runs/5440997148/jobs/9894461904): 
```
Run check_time.sh forward_steadystate_sensitivities_non_optimal_parameters tests/performance/test.py forward_steadystate_sensitivities_non_optimal_parameters;
numsteps [0]
numstepsB None
numrhsevals [0]
numrhsevalsB None
numerrtestfails [0]
numerrtestfailsB None
numnonlinsolvconvfails [0]
numnonlinsolvconvfailsB None
preeq_cpu_time 0.0
preeq_cpu_timeB 0.0
cpu_time 0.0
cpu_timeB 0.0
posteq_cpu_time 584901.833
posteq_cpu_timeB 0.0
587.70
TOO LONG: 587.70 > 5
Error: Process completed with exit code 1.
```

Previously, Newton's method was successful, so changing the default from `newtonOnly` to `integrateIfNewtonFails` shouldn't be a problem. The parameters did not change. What happened?

plakrisenko:
Maybe it's due to default `NewtonMaxSteps`

dweindl:
> Maybe it's due to default `NewtonMaxSteps`

Where did that change?

PR #2084, opened 2023-05-11
Fix sensitivities for models with events / extend and cleanup event tests

### LINKED ISSUE #2086: Incorrect forward sensitivities in model with events
For some models with events, incorrect forward sensitivities are computed.

See https://github.com/AMICI-dev/AMICI/pull/2084#issuecomment-1545264603 for two examples.

Those examples seem to work with FSA on https://github.com/AMICI-dev/AMICI/pull/1539 though.

dweindl:
Fixes incorrect sensitivities for models with events with state-dependent trigger functions. 

Add three new event test cases and do some cleanup.

Closes #2086


dweindl:
There seems to be some issue with forward sensitivities for `python/tests/test_events.py::test_models[piecewise_plus_event_simple_case`. 

Only for d_x1/d_alpha, rest is okay.
image

FFroehlich (approved):
🙇

FFroehlich:
> There seems to be some issue with forward sensitivities for `python/tests/test_events.py::test_models[piecewise_plus_event_simple_case]`.
> 
> Only for d_x1/d_alpha, rest is okay. image

oh!

stephanmg:
> > There seems to be some issue with forward sensitivities for `python/tests/test_events.py::test_models[piecewise_plus_event_simple_case]`.
> > Only for d_x1/d_alpha, rest is okay. image
> 
> Oi! This is bad, will try to have a look next week!

True. Will also have a look, we should then sync up perhaps.

python/tests/test_events.py
```diff
@@ -298,17 +309,260 @@ def get_early_x(t):
...
+        timepoints,
+        x_expected,
+        sx_expected,
+    )
+
+
+def model_definition_piecewise_plus_event_semi_complicated():
+    """Test model for boolean operations in a piecewise condition, discrete
+    events and a non-vanishing quadrature for the adjoint state.
+    """
```
dweindl, line 412: Currently fails finite differences check. 

Passes with 
```diff
diff --git a/python/tests/util.py b/python/tests/util.py
index d5907da29..e85160a98 100644
--- a/python/tests/util.py
+++ b/python/tests/util.py
@@ -154,4 +154,4 @@ def check_trajectories_with_forward_sensitivities(
     solver.setRelativeToleranceFSA(1e-13)
     rdata = runAmiciSimulation(amici_model, solver=solver)
     _check_close(rdata["x"], result_expected_x, field="x", rtol=1e-10, atol=1e-12)
-    _check_close(rdata["sx"], result_expected_sx, field="sx", rtol=1e-10, atol=1e-9)
+    _check_close(rdata["sx"], result_expected_sx, field="sx", rtol=1e-07, atol=1e-9)
```

Seems a bit loose, but then again, we are comparing to FD...

python/sdist/amici/de_export.py
```diff
@@ -2024,7 +2024,7 @@ def _compute_equation(self, name: str) -> None:
                         # This part only works if we use self.eq('xdot')
                         # instead of self.sym('xdot'). Not immediately clear
                         # why that is.
-                        tmp_dxdp += smart_multiply(self.eq("xdot"), self.sym("stau").T)
+                        tmp_dxdp += smart_multiply(self.eq("xdot_old"), self.sym("stau").T)
```
FFroehlich: This does seem plausible. Would argue though that `self.sym("xdot_old")` is more intuitive (should be the same, no?).

But the fact that this makes a difference does point to a more conceptual problem. If `xdot` needs to be evaluated before the event (which kinda makes sense), the same should also apply to `ddeltaxdt`, `ddeltaxdx` and `ddeltaxdp`. Probably best to add to derived state, add respective methods to model and compute these quantities pre-event. Probably worthwhile to come up with a testcase where this applies beforehand. Should be straightforward to do, use current testcase and change `deltax` such that `ddeltaxdt`, `ddeltaxdx` and `ddeltaxdp` have a dependency on `x`, i.e., are different pre- and post-event.

dweindl: Thanks for looking into that. I will merge this PR, as it clearly improves things already, and keep your suggestions for another one.

dweindl: PS: Some of these cases are already covered:
```
python/tests/test_events.py::test_models[events_plus_heavisides]
ddeltaxdp [Matrix([
[0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0, 0, 0]]), Matrix([
[0, 0, 0, 0, 0, 0, 0, 0,   0],
[0, 0, 0, 0, 0, 0, 0, 0,   0],
[0, 0, 0, 0, 0, 0, 0, 0, 1/3]]), Matrix([
[0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0, 0, 0]]), Matrix([
[0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0, 0, 0]]), Matrix([
[0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0, 0, 0]])]
ddeltaxdx [Matrix([
[0, 0, -1/2],
[0, 0,    0],
[0, 0,    0]]), Matrix([
[0, 0, 0],
[0, 0, 0],
[0, 0, 0]]), Matrix([
[0, 0, 0],
[0, 0, 0],
[0, 0, 0]]), Matrix([
[0, 0, 0],
[0, 0, 0],
[0, 0, 0]]), Matrix([
[0, 0, 0],
[0, 0, 0],
[0, 0, 0]])]


python/tests/test_events.py::test_models[piecewise_plus_event_simple_case]
ddeltaxdp [Matrix([[0, 0, 1, 0]]), Matrix([[0, 0, 0, 0]]), Matrix([[0, 0, 0, 0]]), Matrix([[0, 0, 0, 0]])]
ddeltaxdx [Matrix([[-1]]), Matrix([[0]]), Matrix([[0]]), Matrix([[0]])]

python/tests/test_events.py::test_models[piecewise_plus_event_semi_complicated]
ddeltaxdp [Matrix([
[0, 0, 1, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0]]), Matrix([
[0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0]]), Matrix([
[0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0]]), Matrix([
[0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0]]), Matrix([
[0, 0, 0, 0, 0, 0, 0],
[0, 0, 0, 0, 0, 0, 0]])]
ddeltaxdx [Matrix([
[-1, 0],
[ 0, 0]]), Matrix([
[0, 1],
[0, 0]]), Matrix([
[0, 0],
[0, 0]]), Matrix([
[0, 0],
[0, 0]]), Matrix([
[0, 0],
[0, 0]])]
```

Only non-zero `ddeltaxdt` seems to be missing.

PR #2088, opened 2023-05-16
Model import: Don't create spl.h and sspl.h for models without splines

### LINKED ISSUE #2087: Model import: Don't create spl.h and sspl.h for models without splines
Don't create spl.h and sspl.h for models without splines.

dweindl:
Closes #2087


FFroehlich (approved):
👍

PR #2094, opened 2023-05-17
SBML import - faster _get_list_of_species_references

dweindl:
Avoid iterating over all SBML elements.

Takes only 0.04%-0.4% of the time of the old implementation. Saves some seconds to minutes for larger models.


FFroehlich (approved):
👍

Issue #18, opened 2016-07-18
adjoint sensitivies for models with events


FFroehlich:
this has not been tested thoroughly yet and might fail

PR #2097, opened 2023-05-17
Warn if ASA is used in combination with events

dweindl:
#18

### DIFF
--- a/python/sdist/amici/swig_wrappers.py
+++ b/python/sdist/amici/swig_wrappers.py
@@ -1,6 +1,8 @@
 """Convenience wrappers for the swig interface"""
 import logging
 import sys
+
+import warnings
 from contextlib import contextmanager, suppress
 from typing import Any, Dict, List, Optional, Sequence, Union
 
@@ -102,6 +104,18 @@ def runAmiciSimulation(
     :returns:
         ReturnData object with simulation results
     """
+    if (
+        model.ne > 0
+        and solver.getSensitivityMethod()
+        == amici_swig.SensitivityMethod.adjoint
+        and solver.getSensitivityOrder() == amici_swig.SensitivityOrder.first
+    ):
+        warnings.warn(
+            "Adjoint sensitivity analysis for models with events with parameter-dependent trigger functions has not been thoroughly tested. "
+            "Sensitivities might be wrong. Tracked at https://github.com/AMICI-dev/AMICI/issues/18. "
+            "If your model does not have parameter-dependent trigger functions, you can safely ignore this message."
+        )
+
     with _capture_cstdout():
         rdata = amici_swig.runAmiciSimulation(
             _get_ptr(solver), _get_ptr(edata), _get_ptr(model)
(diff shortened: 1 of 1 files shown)


FFroehlich (commented):
‚ne‘ counts discontinuities in the rhs and x. Discontinuities in the rhs are fine with adjoint sensitivity analysis as long as the trigger function is not parameter dependent.

dweindl:
> ‚ne‘ counts discontinuities in the rhs and x. Discontinuities in the rhs are fine with adjoint sensitivity analysis as long as the trigger function is not parameter dependent.

Ok, thanks. As this is not apparent from the generated model module, and i don't feel like storing it there - how do we want to handle that?

1) change the message and live with the warnings for non-parameter-dependent trigger functions
2) rather warn once during model import 
3) leave it as it was

FFroehlich:
> > ‚ne‘ counts discontinuities in the rhs and x. Discontinuities in the rhs are fine with adjoint sensitivity analysis as long as the trigger function is not parameter dependent.
> 
> Ok, thanks. As this is not apparent from the generated model module, and i don't feel like storing it there - how do we want to handle that?
> 
> 1. change the message and live with the warnings for non-parameter-dependent trigger functions
> 2. rather warn once during model import
> 3. leave it as it was before (no warning)

Hmm rephrase and state that it might not apply to the model at hand?

python/sdist/amici/swig_wrappers.py
```diff
@@ -102,6 +104,18 @@ def runAmiciSimulation(
...
+    if (
+        model.ne > 0
+        and solver.getSensitivityMethod()
+        == amici_swig.SensitivityMethod.adjoint
+        and solver.getSensitivityOrder() == amici_swig.SensitivityOrder.first
+    ):
+        warnings.warn(
+            "Adjoint sensitivity analysis for models with events with parameter-dependent trigger functions has not been thoroughly tested. "
+            "Sensitivities might be wrong. Tracked at https://github.com/AMICI-dev/AMICI/issues/18. "
+            "If your model does not have parameter-dependent trigger functions, you can safely ignore this message."
```
FFroehlich, line 116:
```suggestion
            "Adjoint sensitivity analysis may work if the location of the discontinuity is not parameter-dependent, but we still recommend testing accuracy of gradients."
```

python/sdist/amici/swig_wrappers.py
```diff
@@ -152,6 +166,18 @@ def runAmiciSimulations(
...
     :returns: list of simulation results
     """
+    if (
+        model.ne > 0
+        and solver.getSensitivityMethod()
+        == amici_swig.SensitivityMethod.adjoint
+        and solver.getSensitivityOrder() == amici_swig.SensitivityOrder.first
+    ):
+        warnings.warn(
+            "Adjoint sensitivity analysis for models with events with parameter-dependent trigger functions has not been thoroughly tested. "
```
FFroehlich, line 176:
```suggestion
            "Adjoint sensitivity analysis for models with discontinuous right hand sides (events/piecewise functions) has not been thoroughly tested. "
```

python/sdist/amici/swig_wrappers.py
```diff
@@ -152,6 +166,18 @@ def runAmiciSimulations(
...
+            "Sensitivities might be wrong. Tracked at https://github.com/AMICI-dev/AMICI/issues/18. "
+            "If your model does not have parameter-dependent trigger functions, you can safely ignore this message."
```
FFroehlich, line 178:
```suggestion
            "Adjoint sensitivity analysis may work if the location of the discontinuity is not parameter-dependent, but we still recommend testing accuracy of gradients."
```

FFroehlich (changes requested):

FFroehlich (approved):

PR #2105, opened 2023-05-23
Fix unquoted filename

dweindl:
Fixes issues with paths containing blanks:
`sh: line 1: C:/Program: No such file or directory` https://github.com/AMICI-dev/AMICI/actions/runs/5054632314/jobs/9069784824?pr=2104


FFroehlich (approved):
👍

PR #2111, opened 2023-05-23
Fix compilation error for models with events and xdot=0

### LINKED ISSUE #2110: Compilation error during import of SBML semantic case 01355
When importing SBML semantic case 01355 with enabled sensitivity code, compilation fails with:

```
/home/dweindl/src/AMICI-devel/tests/SBMLTestModels/01355/deltasx.cpp:23:61: error: ‘dA__xdt’ was not declared in this scope; did you mean ‘A__x’?
   23 |                     deltasx[0] = -stau0*xdot_old0 + stau0*(-dA__xdt + xdot_old0) - sx0;
      |                                                             ^~~~~~~
      |                                                             A__x
```

`dA__xdt` should be defined in `xdot.h` which is not generated. Maybe because xdot = [0] ?

dweindl:
Fixes #2110

### DIFF
--- a/python/sdist/amici/de_export.py
+++ b/python/sdist/amici/de_export.py
@@ -1999,9 +1999,10 @@ class DEModel:
 
                 # need to check if equations are zero since we are using
                 # symbols
-                if not smart_is_zero_matrix(self.eq("stau")[ie]):
+                if not smart_is_zero_matrix(self.eq("stau")[ie]) \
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):
if `xdot` is zero, `xdot_old` will also be zero, i.e. we can completely drop that term, thus, would suggest to integrate this with the check whether stau is zero.

PR #2118, opened 2023-06-02
[petab] fix pysb pattern matching

FFroehlich:
can't assume equations have been generated and `SpeciesPatternMatcher` won't work otherwise.


dweindl (approved):
:+1:

PR #2120, opened 2023-06-08
SBML import: Support for rateOf

### LINKED ISSUE #769: Python SBML import: add support for rateOf()
e.g.
```
TestCase 01256 was skipped: Encountered unsupported expression "rateOf" of type "<class 'sympy.core.function.UndefinedFunction'>" as part of a Rule: "rateOf(S1)"!
```

dweindl:
Adds support for SBML's `rateOf(.)` for most not-so-exotic cases.

We currently can't check for functions with `<csymbol encoding="text" definitionURL="http://www.sbml.org/sbml/symbols/rateOf">`, and only check for functions named `rateOf`. This should be safe as long as we use the `SBMLFunctionDefinitionConverter` before. 

Tested in the following SBML semantic test suite cases: 01248,01249,01250,01251,01252,01253,01254,01255,01256,01257,01258,01259,01260,01261,01262,01263,01264,01265,01266,01267,01268,01269,01270,01293,01294,01295,01296,01297,01298,01299,01321,01322,01400,01401,01402,01403,01405,01406,01408,01409,01455,01456,01457,01458,01459,01460,01461,01462,01463,01482,01483,01525,01526,01527,01528,01529,01540,01541,01542,01543

Closes #769


FFroehlich (commented):
looks good.

why don't we have to toposort here? are nested rateOfs not allowed by sbml?

FFroehlich (approved):

dweindl:
> why don't we have to toposort here? are nested rateOfs not allowed by sbml?

You mean higher-order derivatives as in `rateOf(rateOf(x))`? That's explicitly forbidden, yes.

But toposort might be required for xdot, true. Currently it might produce something like `xdot1 = xdot2; xdot2 = 2;`, need to check.

FFroehlich:
> > why don't we have to toposort here? are nested rateOfs not allowed by sbml?
> 
> You mean higher-order derivatives as in `rateOf(rateOf(x))`? That's explicitly forbidden, yes.
> 
> But toposort might be required for xdot, true. Chaining is allowed (e.g. `dA/dt = rateOf(B); dB/dt = rateOf(C); dC/dt = 2`). Currently it might produce something like `xdot1 = xdot2; xdot2 = xdot3; xdot3 = 2;`, need to check.
> 
> I'll have to think about that a bit more.

Yes, worried about chaining. The example above looks okay, but I don't think I fully understand the code yet. Wouldn't that lead to issues as evaluation needs to happen xdot3 -> xdot2 -> xdot1 (all assigned to 2), but would evaluate xdot1->xdot2->xdot3 (0,0,2 when everything is initialized to 0). Will run into similar issues when computing derivatives as, e.g., dxdot1/dx would be `dxdot1/dx + dxdot1/dxdot2*(dxdot2/dx + dxdot2/dxdot3*dxdot3/dx)`, and the current code would only consider `dxdot1/dx + dxdot1/dxdot2*dxdot2/dx`, i.e., a single chaining. Easiest might be to add the respective xdot to expression as we already handle these chained dependencies there.

dweindl:
> The example above looks okay, but I don't think I fully understand the code yet. Wouldn't that lead to issues as evaluation needs to happen xdot3 -> xdot2 -> xdot1 (all assigned to 2), but would evaluate xdot1->xdot2->xdot3 (0,0,2 when everything is initialized to 0).

Yes, it was meant as an example of what could currently go wrong. 

> Easiest might be to add the respective xdot to expression as we already handle these chained dependencies there.

I need to double check, but I think only RateRule processing needs to be adapted for that.  The rest it's already handled via expressions.

PR #2122, opened 2023-06-19
Fix PEtab deprecation warning

dweindl:
Fixes `/home/runner/work/AMICI/AMICI/python/tests/splines_utils.py:265: DeprecationWarning: The `sbml_file` argument is deprecated and will be removed in a future version. Use `model_file` instead.`


FFroehlich (approved):
👍

PR #2124, opened 2023-06-19
Fix sp.Matrix errors with numpy==1.25

dweindl:
and DeprecationWarnings with older numpy (`venv/lib/python3.11/site-packages/sympy/matrices/matrices.py:995: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
    if dat in ([], [[]]):`)


dilpath (approved):

dweindl:
> Currently, `integrate_spline` converts a list to an array, then all instances of its usage in this file have `.tolist()`, so perhaps return the list in `integrate_spline` instead?

True. Thought about it, but wasn't sure if anything relies on on the np array and related kwargs elsewhere.

PR #2125, opened 2023-06-19
Readme: containers

dweindl:
Add info on AMICI+containers to README. This was rather difficult to find so far...


FFroehlich (approved):
👍

PR #2129, opened 2023-06-20
Add min version for pandas

dilpath:
This line is currently a tuple, which is supported by the latest pandas version (2.0.2), but not some earlier versions (e.g. 1.5.3).

https://github.com/AMICI-dev/AMICI/blob/0fc51c0ddcac0661a83931e8b1c4a59f0c6db85c/python/sdist/amici/petab_objective.py#L821


dweindl (approved):

FFroehlich (approved):
👍

PR #2131, opened 2023-06-23
Fix sympy DeprecationWarning

dweindl:
During SBML test case 01288 when handling `true`/`false`:

```
E       sympy.utilities.exceptions.SymPyDeprecationWarning:
E
E       non-Expr objects in a Matrix is deprecated. Matrix represents
E       a mathematical matrix. To represent a container of non-numeric
E       entities, Use a list of lists, TableForm, NumPy array, or some
E       other data structure instead.
E
E       See https://docs.sympy.org/latest/explanation/active-deprecations.html#deprecated-non-expr-in-matrix
E       for details.
E
E       This has been deprecated since SymPy version 1.9. It
E       will be removed in a future version of SymPy.
```


FFroehlich (approved):
👍

PR #2132, opened 2023-06-24
Reduce flakiness of benchmark model tests

dweindl:
Fails too often. Increase repetitions, use min instead of mean.


FFroehlich (approved):
looks like outlier is always first run, probably due to generation/caching of petab mapping. might be sufficient to do 3-4 runs after all.

dweindl:
> looks like outlier is always first run

That's weird. We are only checking C++ times (-> `sum(r.cpu_time + r.cpu_timeB for r in res[RDATAS]) / 1000`)

Reverted to 3 repetitions, let's see...

Issue #1192, opened 2020-08-07
Allow hardcoding of numerical values in pysb models


FFroehlich:
Current workaround is to create a `pysb.Expression` that has the numerical value as value, but that abuses the API. Should be straightforward to allow a list of parameter values that are to be hardcoded as numerical value during model compilation.

dweindl:
- [ ] Do the same for SBML import

PR #2134, opened 2023-06-28
SBML import: Allow hardcoding of numerical values

dweindl:
Allows selecting parameters whose values are to be hard-coded (#1192).

So far, restricted to parameters that aren't targets of rule or initial assignments. 
This can be extended later: lifting those restrictions on parameters, allow hard-coding Species with constant=True, ...


python/sdist/amici/sbml_import.py
```diff
@@ -677,8 +714,19 @@ def _gather_base_locals(self):
...
             if not c.isSetId():
                 continue
-
-            self.add_local_symbol(c.getId(), _get_identifier_symbol(c))
+            if c.getId() in hardcode_symbols:
+                if self.sbml.getRuleByVariable(c.getId()) is not None:
+                    raise ValueError(
+                        f"Cannot hardcode symbol `{c.getId()}` that is a rule target."
+                    )
+                if self.sbml.getInitialAssignment(c.getId()):
```
FFroehlich, line 722: add event assignements. What about assignment rules?

FFroehlich (approved):
👍

python/sdist/amici/sbml_import.py
dweindl: Event assignments were missing, true. Assignment rules were handled by `getRuleByVariable`.  Now changed to checking for the `constant` attribute which covers both.

PR #2137, opened 2023-07-05
prevent the second run of newton's method in case simulation was turn…

plakrisenko:
…ed off

The second attempt only makes sense if simulation was attempted before and the starting point changed


dweindl (approved):
Thanks. Yes.

PR #2142, opened 2023-07-08
Simplify / document antimony import

dweindl:
* Wrapper for importing antimony models, this will allow for more concise test cases in the future
* Remove tellurium dependency, use libantimony directly (so we don't need ncurses and other deps)
* Add antimony example

### DIFF
--- a/.github/workflows/test_python_cplusplus.yml
+++ b/.github/workflows/test_python_cplusplus.yml
@@ -64,7 +64,6 @@ jobs:
             libboost-serialization-dev \
             libboost-chrono-dev \
             libhdf5-serial-dev \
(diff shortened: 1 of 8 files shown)


FFroehlich (approved):
nice! does it make sense to also enable petab import for antimony?

dweindl:
> does it make sense to also enable petab import for antimony?

At least for now, I would not do that. Antimony<->SBML interconversion is rather lossless and fast, so I don't think there is much of an advantage to support antimony directly with PEtab. I would still default to SBML in PEtab unless there is compelling reason to use something else.  If there is a request backed by sufficiently many users, we could add it at some point, though.

PR #2145, opened 2023-07-12
Fix SBML import for events with trigger functions depending on parame…

dweindl:
Fix SBML import for events with trigger functions depending on parameters that are initial assignment targets.

I.e. given the current way of processing initial assignments, events have to be processed first. Otherwise model compilation fails with `use of undeclared identifier`, because the respective parameter was replaced by its initial assignment value.

Credits to @LeaSeep for triggering that issue.


FFroehlich (approved):
👍

Issue #2151, opened 2023-07-26
AMICI installation on Windows


anavalinhas:
Hi there, 

I'm trying to install the amici package on windows using Microsoft Visual Studio. I am following the instructions [here](https://amici.readthedocs.io/en/latest/python_installation.html).

I have installed SWIG v4.1.1 and followed the steps under OpenBLAS to build OpenBLAS binaries from scratch. I've made all the changes in PATH variables as necessary.

I have copied the error message onto a separate file bellow. I would really appreciate any help/suggestions you may have.

[amici_install_windows_error_file.txt](https://github.com/AMICI-dev/AMICI/files/12174802/amici_install_windows_error_file.txt)

dweindl:
Hi @anavalinhas, thanks for the detailed report. Could you please try if changing `/LIBPATH:C:/BLAS/OpenBLAS/lib openblas.lib` to `-LIBPATH:C:/BLAS/OpenBLAS/lib openblas.lib` helps? (in the `BLAS_LIBS` environment variable or in the registry, depending on which option you chose initially)

anavalinhas:
@dweindl I have modified the BLAS_LIBS variable as suggested and I get another error. The system fails to find 'unistd.h'. I've attached a file with the full error message bellow for reference.

[amici_install_windows_error_file_1.txt](https://github.com/AMICI-dev/AMICI/files/12193116/amici_install_windows_error_file_1.txt)

dweindl:
Great, so the first issue is resolved, I will update the documentation accordingly.

Regarding `amici\src\hdf5.cpp(30): fatal error C1083: Cannot open include file: 'unistd.h': No 
such file or directory`: 

It seems like so far, nobody had HDF5 available on Windows and this issue never surfaced. 

If you don't need HDF5 support, you can try installing AMICI after setting an environment variable `AMICI_TRY_ENABLE_HDF5=OFF` - this was only introduced recently and is not yet available on pypi:

* if you have a bash-like shell:
`AMICI_TRY_ENABLE_HDF5=OFF pip3 install -e git+https://github.com/AMICI-dev/amici.git@develop#egg=amici\&subdirectory=python/sdist`
* with cmd.exe, something like
  ```
  set AMICI_TRY_ENABLE_HDF5=OFF
  pip3 install -e git+https://github.com/AMICI-dev/amici.git@develop#egg=amici\&subdirectory=python/sdist
  ```
  
[...]

PR #2153, opened 2023-07-28
Doc: Update Windows installation instructions

dweindl:
`/LIBPATH` doesn't seem to be compatible with ninja. See #2151.


FFroehlich (approved):
👍

PR #2154, opened 2023-07-28
Cleanup includes in hdf5.cpp

dweindl:
Remove unused includes. Leftovers from the early beginnings, it seems ...

In particular, don't use unistd.h for Windows-compatibility #2151


FFroehlich (approved):
👍

Issue #2119, opened 2023-06-06
Single-file models


dweindl:
Refactor model code generation to allow switching between one-file-per-model-quantity and everything-inside-one source file. This should speed up import of small models significantly. (https://github.com/AMICI-dev/AMICI/issues/1453#issuecomment-1551511491)

PR #2159, opened 2023-08-18
Combine code for sparse model functions and their index files

dweindl:
Combine code for sparse model functions and their index files, i.e. generate only a single file instead of 3 individual files for content, rowvals, and colptrs, respectively.

Advantage: Faster import of smaller models and fewer files. For a toy model, this reduced the build steps from 44 to 28, and reduced build time by >20% on my computer.

Disadvantage: None found, so I don't think it worth adding an option for (not) combining those files. For larger models, there shouldn't be any impact. The extra time for compiling the index arrays should be negligible compared to computing the contents.

Related to #2119


FFroehlich (approved):
I think, historically, this slowed down compilation of larger models

dweindl:
> I think, historically, this slowed down compilation of larger models

Will double-check

dweindl:
Here a test for a large model (N=1):

| File         | Size    | Compilation time (s) |
|--------------|--------:|---------------------:|
| dwdx         | 22.4MiB |              3413.64 |
| dwdx_colptrs |  2.0KiB |                 2.79 |
| dwdx_rowvals | 65.6KiB |                 2.66 |
| *combined*   |         |              3416.79 |

I'd consider this time increase negligible.

PR #2166, opened 2023-08-25
CI: Update to jdk 17 to meet changed sonarcloud requirements

dweindl:
https://docs.sonarcloud.io/appendices/scanner-environment/


dweindl:
The sonar scanner package we are using comes with its own jdk. The [warning](https://github.com/AMICI-dev/AMICI/pull/2164#issuecomment-1693037689) doesn't make much sense. We just need to update the scanner package -> #2167

PR #2173, opened 2023-09-19
Doc: Add installation instructions for Arch Linux

stephanmg:
Instructions for archlinux

### DIFF
--- a/documentation/python_installation.rst
+++ b/documentation/python_installation.rst
@@ -66,8 +66,8 @@ Install AMICI:
 
    pip3 install amici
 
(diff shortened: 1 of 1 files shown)


dweindl (commented):
Thanks!

Could you please also include boost? boost-chrono is currently enough. `libboost-chrono` on ubuntu, not sure how it's organized for arch.

stephanmg:
Added base package boost-libs in Archlinux providing boost-chrono.

dweindl (approved):

PR #2181, opened 2023-10-25
Fix misleading preequilibration failure messages

### LINKED ISSUE #2178: Fix misleading preequilibration failure messages
When preequilibration (or finding a steadystate in general) fails, `SteadystateProblem::handleSteadyStateFailure` currently produces `AMICI simulation failed: Steady state computation failed. First run of Newton solver failed. Simulation to steady state failed: No convergence was achieved. Second run of Newton solver failed.`, even in cases where no Newton solve or no simulation was attempted. This is confusing and needs to be changed.

dweindl:
When preequilibration (or finding a steadystate in general) fails, `SteadystateProblem::handleSteadyStateFailure` produced `AMICI simulation failed: Steady state computation failed. First run of Newton solver failed. Simulation to steady state failed: No convergence was achieved. Second run of Newton solver failed.`, even in cases where no Newton solve or no simulation was attempted. This is confusing and is changed here, so that the message now reflects what has actually happened.

Closes #2178

### DIFF
--- a/include/amici/steadystateproblem.h
+++ b/include/amici/steadystateproblem.h
@@ -208,8 +208,15 @@ class SteadystateProblem {
 
     /**
      * @brief Stores state and throws an exception if equilibration failed
+     * @param tried_newton_1 Whether any Newton step was attempted before
+     * simulation
+     * @param tried_simulation Whether simulation was attempted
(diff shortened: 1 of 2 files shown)


FFroehlich (approved):
👍

PR #2193, opened 2023-11-15
Easier access to timepoints via ExpDataView

### LINKED ISSUE #2191: Easier access to timepoints via ExpDataView
ExpDataView provides convenient access to measurements and the like as numpy arrays. However, accessing the associated timepoints is currently only possible via `ExpDataView(...)._swigptr.ts_` which returns an amici.amici.DoubleVector. That's awkward. Instead, we should have one of `ExpDataView(...).{t,ts,timepoints}` that provides a numpy array.

dweindl:
ExpDataView provides convenient access to measurements and the like as numpy arrays. However, accessing the associated timepoints is currently only possible via `ExpDataView(...)._swigptr.ts_` which returns an amici.amici.DoubleVector. That's awkward.

Now:

```python
amici.ExpDataView(amici.ExpData(1, 2, 3, [4, 5, 6])).ts
Out[3]: array([4., 5., 6.])
```

Resolves #2191

### DIFF
--- a/python/sdist/amici/numpy.py
+++ b/python/sdist/amici/numpy.py
@@ -343,6 +343,7 @@ class ExpDataView(SwigPtrView):
     """
 
     _field_names = [
+        "ts",
         "observedData",
         "observedDataStdDev",
         "observedEvents",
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):
there was `edata.getTimepoints()`, but I agree this is more convenient.

PR #2197, opened 2023-11-15
GHA: Modularize actions (apt dependencies, sonar tools)

dweindl:
DRY, use composite actions. More readable + more maintainable.

### DIFF
--- a/.github/workflows/test_install.yml
+++ b/.github/workflows/test_install.yml
@@ -29,8 +29,7 @@ jobs:
       run: |
         sudo apt-get update \
           && sudo apt-get install -y \
-            cmake \
-            g++
+            cmake
 
     - name: Build suitesparse
       run: |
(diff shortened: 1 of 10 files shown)


.github/workflows/test_install.yml
```diff
@@ -22,17 +22,14 @@ jobs:
 
     - run: echo "AMICI_DIR=$(pwd)" >> $GITHUB_ENV
 
-    # install amici dependencies
+    - name: Install apt dependencies
+      uses: ./.github/actions/install-apt-dependencies
+
     - name: apt
       run: |
         sudo apt-get update \
```
FFroehlich, line 30: Skip the update here?

FFroehlich (approved):
Nice!

.github/workflows/test_install.yml
dweindl: Takes only about 1s to run it a second time, I'd leave it here to keep it's safer to shuffle those steps around.

PR #2205, opened 2023-11-18
Refactor: PEtab code to sub-package

dweindl:
Move PEtab-related code to a new `amici.petab` subpackage. An attempt to organize the growing codebase a bit better. More to be done. Unless anybody is using private functions, this should not break anything.

More parts to be moved separately. No change in functionality.


FFroehlich (approved):
👍

PR #2207, opened 2023-11-20
GHA: Fix OpenMP not found on macos

### LINKED ISSUE #2182: GHA: OpenMP not found on macos
e.g. https://github.com/AMICI-dev/AMICI/actions/runs/6637490835/job/18031913365#step:9:853

```
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) 
```

- [ ] ensure cmake can find the libomp installation 
- [ ] update installation guide if relevant
- [ ] adapt test to always expect OpenMP to be available when running on GHA

dweindl:
* Ensure cmake can find OpenMP
* Assert that amici was compiled with OpenMP support
* Update installation guide

Fixes #2182

### DIFF
--- a/.github/workflows/test_python_cplusplus.yml
+++ b/.github/workflows/test_python_cplusplus.yml
@@ -62,6 +62,9 @@ jobs:
     - name: Install python package
       run: scripts/installAmiciSource.sh
 
+    - name: Check OpenMP support
+      run: source build/venv/bin/activate && python -c "import amici; import sys; sys.exit(not amici.compiledWithOpenMP())"
+
     - name: Python tests
       run: |
         source build/venv/bin/activate \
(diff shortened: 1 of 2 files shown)


FFroehlich (approved):
`brew link openmp` and `brew link boost` would also work but adds those paths globally. `$(brew --prefix)` won't work with non-default install directories.

dweindl:
> `brew link openmp` and `brew link boost` would also work but adds those paths globally. 

Okay, will add that to installation guide.

> `$(brew --prefix)` won't work with non-default install directories.

I am not really familiar with brew. Do you mean there are cases where `brew install something` will install the package somewhere else than `$(brew --prefix)`? Or only if other options are passed?

FFroehlich:
> > `$(brew --prefix)` won't work with non-default install directories.
> 
> I am not really familiar with brew. Do you mean there are cases where `brew install something` will install the package somewhere else than `$(brew --prefix)`? Or only if other options are passed?

thought it was possible, but it's actually not. `$(brew --prefix)` should always work.

PR #2212, opened 2023-11-27
Update python_installation.rst

willov:
On Python 3.11 and with openblas 0.3.24-2, the amici installer is unable to find "cblas.h". This is fixed by exporting the openblas path to BLAS_CFLAGS


dweindl:
Thanks @willov.

dweindl:
@willov : Maybe missing some `BLAS_LIBS=-L/path/to/libopenblas.{a,so}`?

willov:
Thanks! You are correct, `export BLAS_LIBS="-L/usr/lib/ -lopenblas"` solved it in my case (also with a `--no-cache-dir`). I will check if this also solved the problems for @Podde1, then I will issue another merge request.

dweindl:
> Thanks! You are correct, `export BLAS_LIBS="-L/usr/lib/ -lopenblas"` solved it in my case (also with a `--no-cache-dir`). I will check if this also solved the problems for @Podde1, then I will issue another merge request.

:+1: 

I will leave this PR open for now. Please update on this branch. Is `-L/usr/lib/` required? I would have expected this directory to be included by default. I.e. is `export BLAS_LIBS=-lopenblas` not sufficient?
Right, `--no-cache-dir` is required with recent pip versions in case of linker errors.

Podde1:
I am have tested this in my environment now aswell. It was indeed sufficient to set `export BLAS_LIBS=-lopenblas` and exclude  `-L/usr/lib/`.

willov:
`export BLAS_LIBS=-lopenblas` was indeed sufficient. I have updated the branch to also include instructions for exporting BLAS_LIBS.

Issue #2185, opened 2023-10-30
Handle events with time-based triggers without rootfinding?


dweindl:
We wouldn't really need any of sundials' rootfinding for handling events with time-based triggers.

Handling those events without rootfinding should be more efficient and solve some of those issues:

> Cvode routine CVode returned a root after reinitialization. The initial step-size after the event or heaviside function is too small. To fix this, increase absolute and relative tolerances! failed with error code 2

Not yet sure how much effort it would be.

dweindl:
- [x] Events that trigger at a single constant (non-parameter-dependent) timepoint (#2227)
- [ ] Events that trigger at a single constant parameter-dependent timepoint
- [ ] ... (probably not worth the extra effort)

PR #2227, opened 2023-12-04
Handle events occuring at fixed timepoints without root-finding

dweindl:
A first attempt towards #2185

For events that occur at known timepoints, we don't need sundials' root-finding. We can just stop the solver at the respective timepoints and handle the events.

Here, events are sorted such that the `ne_solver` events  that require root-finding by the solver come first and the other `ne - ne_solver` events come after that. The solver only tracks  `ne_solver` roots. 

To be extended to parameterized but state-independent trigger functions at some point.


python/sdist/amici/de_model.py
```diff
@@ -713,3 +714,21 @@ def __eq__(self, other):
         return self.get_val() == other.get_val() and (
             self.get_initial_value() == other.get_initial_value()
         )
+
+    def triggers_at_fixed_timepoint(self) -> bool:
+        """Check whether the event triggers at a (single) fixed time-point."""
+        t_root = sp.solve(self.get_val(), amici_time_symbol)
```
FFroehlich, line 720: Does it make sense to cache the value of t_root at initialization? `get_trigger_time` alone computes this twice.  Would also reduce some redundancy.

src/forwardproblem.cpp
```diff
@@ -109,25 +130,65 @@ void ForwardProblem::workForwardProblem() {
...
-            while (t_ < nextTimepoint) {
-                int status = solver->run(nextTimepoint);
-                solver->writeSolution(&t_, x_, dx_, sx_, dx_);
+        if (next_t_out > model->t0()) {
+            // Solve for next output timepoint
+            while (t_ < next_t_out) {
+                if (is_next_t_too_close(t_, next_t_out)) {
+                    // next timepoint is too close to current timepoint
+                    // update `t_`, required by `handleDataPoint`
+                    t_ = next_t_out;
```
FFroehlich, line 157: isnt the value of `t_` always overwritten by `solver->writeSolution`?

src/forwardproblem.cpp
```diff
@@ -156,13 +217,9 @@ void ForwardProblem::handleEvent(
     /* store Heaviside information at event occurrence */
     model->froot(t_, x_, dx_, rootvals_);
 
-    /* store timepoint at which the event occurred*/
+    /* store timepoint at which the event occurred */
     discs_.push_back(t_);
 
-    /* extract and store which events occurred */
```
FFroehlich, line 162: Why was this removed?

FFroehlich (approved):
👍

python/sdist/amici/de_model.py
dweindl: Can be done. I don't think computing it twice matters performance-wise, but the redundancy is not so nice, agreed.

src/forwardproblem.cpp
dweindl: Correct. But `solver->writeSolution` is only called for the next timepoint once `handleDataPoint` finishes. By then it's fine that it gets overwritten.

src/forwardproblem.cpp
dweindl: I lifted that up into `ForwardProblem::workForwardProblem`. This is the only place where `handleEvent` is called with `!seflag && !initial_event`. We need it out there, because we need to modify `roots_found_` *after* calling `getRootInfo`.

src/forwardproblem.cpp
FFroehlich: That‘s not evident from the code here though

PR #2230, opened 2023-12-11
buildAmici.sh: Only try installing numpy if not yet available

dweindl:
Makes buildAmici.sh work with system-python. Before, it would fail if we don't have write permissions to system-packages even if numpy was already installed.

### DIFF
--- a/scripts/buildAmici.sh
+++ b/scripts/buildAmici.sh
@@ -25,7 +25,7 @@ else
 fi
 
 # required for build swig interface
-python3 -m pip install numpy
+pip show numpy > /dev/null || python3 -m pip install numpy
 
 ${cmake} \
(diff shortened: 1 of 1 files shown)


dweindl:
> check against some minimum version?

Wouldn't hurt, but it's a bit of a pain. We'll either have to hardcode minimum numpy versions for different python versions and platforms here, or we need to parse the constraints from pyproject.toml.
It would be easiest if pip would support installing only build-dependencies, but that's not possible unfortunately.

I would rather address that some other time. This PR doesn't really change anything in that respect. Before this change, numpy wouldn't have been upgraded either. With system-pip, `pip install numpy` would already fail before even checking if there is any numpy installed.

PR #2232, opened 2023-12-12
Fix potential out-of-bounds read in Model::checkFinite

dweindl:
Fix potential out-of-bounds read and incorrect parameter IDs in error messages in Model::checkFinite: `dwdp_` has size `nw` x `np`, not `nw` x `nplist`.


FFroehlich (approved):
👍

PR #2234, opened 2023-12-12
Fix piecewise/Heaviside handling

### LINKED ISSUE #2231: NaNs in model expressions with DiracDelta due to unfortunate term ordering
Encountered the following problem with https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/tree/d7c52f90b15ca83a3d3f832f08989f1448fbd01e/Benchmark-Models/Bertozzi_PNAS2020:

Some expressions in dwdp that involve DiracDeltas may become NaN when evaluated for DiracDelta(0). DiracDelta(0) is currently implemented as DBL_MAX, not infinity. Therefore, `a * DiracDelta(0)` is finite for `a` in [-1, 1]. 
For the Bertozzi_PNAS2020 model, we have something like `a * DiracDelta(0) * b`, which is non-finite if `a` is outside [-1, 1], even if `a * b` would be in [-1, 1] (in this specific case, a*b == 0). Reordering the multiplicands fixes that problem.

Not sure how to handle that robustly. In this case, just moving the DiracDelta as far to the right as possible fixes the issue. However, in other situations it might be beneficial if a DiracDelta(x != 0) occurs further left.

Any thoughts?

dweindl:
* Fixes a bug where Heaviside functions weren't correctly substituted, potentially resulting in DiracDelta's in the model equations. The problem was, the substitution targets were collected from an expanded expression, but the actual substitution was attempted on the non-expanded expression, which did not always succeed.

  Fixed by *not* expanding the expressions beforehand. Closes #2231
  
* Fixes redundant trigger-function processing


dweindl:
Okay, this might have been a more widespread problem - this also fixes frequent Brannmark_JBC2010 issues (i.e. breaks my example at https://amici.readthedocs.io/en/latest/example_errors.html#Steady-state-computation-failed :-|)

dweindl:
Worth checking whether this change also affects the issues encountered in #1539

dweindl:
> Okay, this might have been a more widespread problem - this also fixes frequent Brannmark_JBC2010 issues (i.e. breaks my example at https://amici.readthedocs.io/en/latest/example_errors.html#Steady-state-computation-failed :-|)

No. For Brannmark_JBC2010 didn't have any stray Heavisides. Maybe just calling `sympy.expand()` did the trick there...

dweindl:
```
=========================== short test summary info ============================
FAILED test_petab_benchmark.py::test_benchmark_gradient[Giordano_Nature2020-True] - AssertionError:                                                          direction  ...                       analysis_results
  directi...   [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, ...  ...  Empty DataFrame
  Columns: []
  Index: []
  
  [50 rows x 6 columns]
assert False
============= 1 failed, 36 passed, 7 skipped in 941.58s (0:15:41) ==============
```

Unable to reproduce locally :roll_eyes:

python/sdist/amici/de_export.py
```diff
@@ -2767,9 +2767,11 @@ def _process_heavisides(
         if heavisides:
             # only apply subs if necessary
             for heaviside_sympy, heaviside_amici in heavisides:
-                dxdt = dxdt.subs(heaviside_sympy, heaviside_amici)
+                dt_expanded = dt_expanded.subs(
```
FFroehlich, line 2770: was the issue that `expand` led to manipulation of the trigger function which was then missed during this substitution?
 
 Not super happy that we manipulate the rhs by applying expand here as I think we discussed at some point we want to keep simplification optional. I don't see why the expand is necessary here in the first place. Looks like this was introduced in https://github.com/AMICI-dev/AMICI/pull/1352

dweindl: > was the issue that `expand` led to manipulation of the trigger function which was then missed during this substitution?

Correct.

> Not super happy that we manipulate the rhs by applying expand here as I think we discussed at some point we want to keep simplification optional. I don't see why the expand is necessary here in the first place. Looks like this was introduced in #1352

```
        # expanding the rhs will in general help to collect the same
        # heaviside function
```

So I guess, not expanding the expression may lead to more root functions to be tracked than strictly necessary in certain cases. I don't know how common this issue really is, and how much of a performance impact each trigger function has.

I am open to dropping `expand`.

dweindl: > Looks like this was introduced in #1352

Specifically in https://github.com/AMICI-dev/AMICI/pull/1352/commits/f27004fd029671d5b862378adace3de7ba248d3b
Doesn't really provide more context, though.

FFroehlich: > > was the issue that `expand` led to manipulation of the trigger function which was then missed during this substitution?
> 
> Correct.
> 
> > Not super happy that we manipulate the rhs by applying expand here as I think we discussed at some point we want to keep simplification optional. I don't see why the expand is necessary here in the first place. Looks like this was introduced in #1352
> 
> ```
>         # expanding the rhs will in general help to collect the same
>         # heaviside function
> ```
> 
> So I guess, not expanding the expression may lead to more root functions to be tracked than strictly necessary in certain cases. I don't know how common this issue really is, and how much of a performance impact each trigger function has.
> 
> I am open to dropping `expand`.

But we are anyways calling `sp.simplify` in `_get_unique_root`. If expand really makes a difference, we could do `sp.simplify(root_found.expand() - root.get_val())`, but I would be surprised if that's really the case. We can map multiple Heaviside functions to the same root finding function, so I would prefer keeping all symbolic manipulations in `_get_unique_root`.

dweindl: Well, I don't think the `expand` ever really improved anything in terms of reducing the number of root functions. Either it would simply have had no effect at all (i.e. the same roots would have been extracted with or without `expand`), or it would have prevented replacing the Heavisides because the to-be-substituted expressions weren't found in the original expression.

FFroehlich (approved):

PR #2240, opened 2023-12-15
code-gen: Simplify `switch`

dweindl:
Improves model code generation by collapsing cases with identical statements.

I.e.
```
switch(a)
 case b:
 case c:
  statements;
  break;
```
instead of

```
switch(a)
 case b:
  statements;
  break;
 case c:
  statements;
  break;
```

For my current model of interest, containing many events, this significantly reduces the generated code:

E.g.:
```
 16K my_model/deltasx.cpp
6,6M my_model_old/deltasx.cpp
```

Overall, for this model, I got from 204201 LOC down to 7936 LOC (i.e. -96%).

### DIFF
--- a/python/sdist/amici/cxxcodeprinter.py
+++ b/python/sdist/amici/cxxcodeprinter.py
@@ -303,7 +303,9 @@ def get_switch_statement(
     indentation_step: Optional[str] = " " * 4,
 ):
     """
-    Generate code for switch statement
+    Generate code for a C++ switch statement.
+
+    Generate code for a C++ switch statements with a ``break`` after each case.
 
     :param condition:
         Condition for switch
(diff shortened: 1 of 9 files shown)


python/sdist/amici/cxxcodeprinter.py
```diff
@@ -303,7 +303,9 @@ def get_switch_statement(
     indentation_step: Optional[str] = " " * 4,
 ):
     """
-    Generate code for switch statement
+    Generate code for a C++ switch statement.
+
+    Generate code for a C++ switch statements with a ``break`` after each case.
```
dilpath, line 308:
```suggestion
    Generate code for a C++ switch statement with a ``break`` after each case.
```

python/sdist/amici/cxxcodeprinter.py
```diff
@@ -321,26 +323,43 @@ def get_switch_statement(
...
-
-    return lines
+            case_code = f"{indent1}case {expression}:"
+
+            try:
+                # there is already a case with the same statement, append
+                cases_map[statement_code].append(case_code)
+            except KeyError:
+                # add new case + statement
+                cases_map[statement_code] = [case_code]
```
dilpath, line 351:
```suggestion
            cases_map[statement_code] = cases_map.get(statement_code, []) + [case_code]
```

python/sdist/amici/cxxcodeprinter.py
```diff
@@ -321,26 +323,43 @@ def get_switch_statement(
...
+    def get_lines():
+        for statements, case_code in cases_map.items():
+            yield from case_code
+            yield from statements
+
+    return [
+        f"{indent0}switch({condition}) {{",
+        *(get_lines()),
+        indent0 + "}",
+    ]
```
dilpath, line 365: Same or different?

```suggestion
    return [
        f"{indent0}switch({condition}) {{",
        *[code for k, v in cases_map.items() for code in [k,v]],
        indent0 + "}",
    ]
```

dilpath (commented):
Nice :+1:

python/sdist/amici/cxxcodeprinter.py
dweindl: Same result, but constructs potentially plenty of temporary lists. Preferring the former.

dilpath: Ah, how about

```suggestion
    return [
        f"{indent0}switch({condition}) {{",
        *[code for codes in cases_map.items() for code in codes],
        indent0 + "}",
    ]
```

FFroehlich (approved):
nice 👍

python/sdist/amici/cxxcodeprinter.py
dweindl: Would work with changing `in codes` to `in chain.from_iterable(codes)`.

python/sdist/amici/cxxcodeprinter.py
dweindl: creates an unnecessary list each time

PR #2249, opened 2023-12-18
 Use proper labels for plotting if IDs are available in ReturnData

### LINKED ISSUE #2175: Use proper labels for plotting if IDs are available in ReturnData
`amici.plotting.*` uses some default labels (e.g. x1, x2, x3) if `model=None`. However, in some cases (ReturnDataReportingMode == full), we can still access IDs via the provided `rdata` argument, similar to what is done in 
https://github.com/AMICI-dev/AMICI/blob/847a5966cad44a4c3b353d1cdca520b7e83c4da9/python/sdist/amici/numpy.py#L297
In such a case, those IDs should be used to create more informative plots.

dweindl:
`amici.plotting.*` uses some default labels (e.g. x1, x2, x3) if `model=None`. However, in some cases (ReturnDataReportingMode == full), we can still access IDs via the provided `rdata` argument. Try that.

Closes #2175

### DIFF
--- a/python/sdist/amici/plotting.py
+++ b/python/sdist/amici/plotting.py
@@ -56,7 +56,7 @@ def plot_state_trajectories(
     elif model is not None and prefer_names:
         labels = np.asarray(model.getStateNames())[list(state_indices)]
         labels = [
-            l if l else model.getStateNames()[ix]
+            l if l else model.getStateIds()[ix]
             for ix, l in enumerate(labels)
         ]
     elif model is not None:
@@ -116,7 +116,7 @@ def plot_observable_trajectories(
             list(observable_indices)
         ]
         labels = [
-            l if l else model.getObservableNames()[ix]
+            l if l else model.getObservableIds()[ix]
             for ix, l in enumerate(labels)
         ]
     elif model is not None:
--- a/python/sdist/amici/plotting.py
+++ b/python/sdist/amici/plotting.py
@@ -6,6 +6,7 @@ Plotting related functions
 from typing import Iterable, Optional, Sequence, Union
 
 import matplotlib.pyplot as plt
+import numpy as np
 import pandas as pd
 import seaborn as sns
 from matplotlib.axes import Axes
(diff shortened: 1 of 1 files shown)


python/sdist/amici/plotting.py
```diff
@@ -64,38 +78,54 @@ def plot_observable_trajectories(
...
+
+    if marker is None:
+        # Show marker if only one time point is available,
+        #  otherwise nothing will be shown
+        marker = "o" if len(rdata.t) == 1 else None
+
+    if model is None and rdata.ptr.observable_ids is None:
+        labels = [f"$y_{{{iy}}}$" for iy in observable_indices]
+    elif model is not None and prefer_names:
+        labels = np.asarray(model.getObservableNames())[
```
FFroehlich, line 115: unindexed_observables = model.getObservableNames() if prefer_names else model.getObservableIds()

FFroehlich, line 115: `unindexed_observables = model.getObservableNames() if prefer_names else model.getObservableIds()` to merge `elif model is not None and prefer_names` and `elif model is not None` cases?

python/sdist/amici/plotting.py
```diff
@@ -64,38 +78,54 @@ def plot_observable_trajectories(
...
+            list(observable_indices)
+        ]
+        labels = [
+            l if l else model.getObservableNames()[ix]
```
FFroehlich, line 119:
```suggestion
            l if l else model.getObservableIds()[ix]
```

python/sdist/amici/plotting.py
```diff
@@ -16,42 +17,55 @@
...
+
+    if marker is None:
+        # Show marker if only one time point is available,
+        #  otherwise nothing will be shown
+        marker = "o" if len(rdata.t) == 1 else None
+
+    if model is None and rdata.ptr.state_ids is None:
+        labels = [f"$x_{{{ix}}}$" for ix in state_indices]
+    elif model is not None and prefer_names:
+        labels = np.asarray(model.getStateNames())[list(state_indices)]
```
FFroehlich, line 57: see below

python/sdist/amici/plotting.py
```diff
@@ -16,42 +17,55 @@
...
+        labels = [
+            l if l else model.getStateNames()[ix]
```
FFroehlich, line 59:
```suggestion
            l if l else model.getStateIds()[ix]
```

FFroehlich (approved):
👍

python/sdist/amici/plotting.py
dweindl: I would rather keep it separate as it's already sufficiently confusing what's to be used :sweat_smile:.
We'd still need some distinction between the two cases to fall back to IDs if the name is ''.

PR #2260, opened 2024-01-05
GHA: Split macOS C++ / Python test jobs

dweindl:
Currently, macOS jobs run for >1h, this can be rather annoying. 
Splitting current tests into two separate jobs. Now we have 33' + 23' instead of 1h20'.

Also:
* cleanup some bash scripts
* manage dependencies via pyproject.toml
* move h5py dependency to extras
* parallelize model compilation on macos

### DIFF
--- a/.github/workflows/test_python_cplusplus.yml
+++ b/.github/workflows/test_python_cplusplus.yml
@@ -12,6 +12,8 @@ jobs:
   ubuntu-cpp-python-tests:
     name: C++/Python tests Ubuntu
     runs-on: ubuntu-22.04
+    env:
+      AMICI_PARALLEL_COMPILE: ""
 
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


.github/workflows/test_python_cplusplus.yml
```diff
@@ -254,8 +256,57 @@ jobs:
...
+    steps:
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.9
+
+    - uses: actions/checkout@v3
+    - run: git fetch --prune --unshallow
+
+    - run: echo "AMICI_DIR=$(pwd)" >> $GITHUB_ENV
```
FFroehlich, line 285: add shell script to avoid code duplication?

FFroehlich (approved):

.github/workflows/test_python_cplusplus.yml
dweindl: done

PR #2255, opened 2024-01-03
Add PetabProblem class for handling PEtab-defined simulation con…

dweindl:
…ditions

Makes it a bit easier to work with PEtab problems interactively or when implementing some PEtab-based objective function (#962).

PR #2265, opened 2024-01-16
simulate_petab: Avoid warning when simulating with default parameters

dweindl:
Fixes a "unused parameters" warning when using simulate_petab with default parameters. By default, fixed parameters are hard-coded during parameter mapping, and thus, must not be passed to `fill_in_edatas` later on. This warning was swallowed until recently (#2243).


dweindl:
Reviewed as part of #2255

PR #2280, opened 2024-01-29
Fix amici.ExpData()

dweindl:
Fix `IndexError: tuple index out of range` when calling `amici.ExpData` without any argument.


FFroehlich (approved):
👍

PR #2289, opened 2024-02-14
User option to fail if model needs to be compiled

dilpath:
e.g. in model selection, often a single superset model is compiled in advance, then many pyPESTO problems/AMICI models are loaded during model selection, by loading and customizing the superset model. If the model is not compiled in advance, this can lead to many models being compiled simultaneously into the same output directory.

The job usually fails immediately, and is easy to fix. Recently, a user reported that they forgot to compile in advance, but the jobs didn't fail immediately. Instead, the job failed after the first iteration of model selection, which took some hours. Somehow, each of their 8 parallel model selection jobs all compiled the AMICI model separately into the same directory, and were able to complete model selection, but then the next set of jobs failed, due to model import/compilation issues.

This PR allows users to set `compile_=False`, in the case that the user knows there must be an error if the model needs to be compiled.


dilpath:
I think the errors are unrelated, e.g. the 4 failing `C++/Python` tests are all due to
```
FAILED python/tests/test_bngl.py::test_compare_to_pysb_simulation[Motivating_example_cBNGL] - ValueError: ComplexPatterns that span multiple volume compartments must have these volume compartments separated by a single surface compartment.
```

FFroehlich (approved):
👍

FFroehlich:
> > I think the errors are unrelated, e.g. the 4 failing `C++/Python` tests are all due to
> > ```
> > FAILED python/tests/test_bngl.py::test_compare_to_pysb_simulation[Motivating_example_cBNGL] - ValueError: ComplexPatterns that span multiple volume compartments must have these volume compartments separated by a single surface compartment.
> > ```
> 
> Yes, those are unrelated, will have a closer look at what's going on.

hopefully fixed in https://github.com/pysb/pysb/pull/570/commits/52570c3ae68cfac90ee7e24cd27e690f68182b06

dilpath:
> hopefully fixed in [pysb/pysb@52570c3](https://github.com/pysb/pysb/commit/52570c3ae68cfac90ee7e24cd27e690f68182b06)

Oh nice, thanks for the quick fix there! Fine with me to leave this open until the next PySB comes out, although I'm unfamiliar with the dev cycle there.

PR #2291, opened 2024-02-19
Fix missing toposort after rateOf-substitutions in `w`

### LINKED ISSUE #2290: Incorrect simulation results when using rateOf in expressions
Amici may produce incorrect simulation results when using `rateOf` in `w` (mostly assignment rules) due to missing toposort.

dweindl:
Fixes potentially incorrect simulation results when using rateOf in `w` where the rates depend on `w`.

Fixes #2290


FFroehlich (approved):
👍

PR #2298, opened 2024-02-20
Fix redundant calls to Model::fdwdx from Model_ODE::fJ

dweindl:
Remove an extraneous call to Model::fdwdx, the result of which will be overridden without use by Model_ODE::fJSparse.


FFroehlich (approved):
🙈

PR #2303, opened 2024-02-22
Split expressions into static and dynamic

### LINKED ISSUE #1269: Split expressions into static and dynamic
static expressions don't depend on time/states and only need to be evaluated after reinitialization, not at every timepoint.

dweindl:
Split expressions in `w` into dynamic (explicitly or implicitly time-dependent) and static ones.
Evaluate static ones only when needed, i.e. after (re)initializing x_rdata or parameters.

See #1269


dweindl:
On second thought, I'll directly include dwd* here as well.

python/sdist/amici/de_export.py
```diff
@@ -1638,6 +1646,132 @@ def free_symbols(self) -> set[sp.Basic]:
...
+            dwdw = self.eq("dwdw")
+            w = self.eq("w")
+
+            # check for indirect state and time dependency
+            # to avoid lengthy symbolic computations,
+            # we only check if we have any non-zeros in hierarchy,
+            # we currently neglect the case where different hierarchy
+            # levels may cancel out. treating a static expression as
+            # dynamic in such rare cases shouldn't be a problem.
+            nonzero_dwdx = np.asarray(
```
FFroehlich, line 1674: rename `nonzero_dwdx` to something like `dynamic_dependency` and explicitly state in the comments that we are looking for direct dependency (via `t`) and indirect dependency (via `x`).

python/sdist/amici/de_export.py
```diff
@@ -1638,6 +1646,132 @@ def free_symbols(self) -> set[sp.Basic]:
...
+                        ]
+                    )[:, np.newaxis],
+                )
+            )
+
+            nonzero_dwdw = np.asarray(
+                dwdw.applyfunc(lambda x: int(not x.is_zero))
+            ).astype(np.int64)
+
+            tmp = nonzero_dwdx
```
FFroehlich, line 1712: add comment that this unravels the hierarchical structure of `w` and that `np.sum(tmp, axis=1)>0` checks for direct/indirect time dependency?

python/sdist/amici/de_export.py
```diff
@@ -1638,6 +1646,132 @@ def free_symbols(self) -> set[sp.Basic]:
...
+                *self.sym("x"),
+                *self.sym("h"),
+                # FIXME see spline comment above
+                # *(self.sym("spl") if name in ("dwdw", "dwdx") else ()),
+                # *(self.sym("sspl") if name == "dwdp" else ()),
+            ]
+            dynamic_syms = sp.Matrix(dynamic_syms)
+            rowvals = self.rowvals(name)
+            sparseeq = self.sparseeq(name)
+            self._static_indices[name] = [
```
FFroehlich, line 1741: add some comments to explain what's being done here?

python/sdist/amici/de_export.py
```diff
@@ -1638,6 +1646,132 @@ def free_symbols(self) -> set[sp.Basic]:
...
+                i
+                for i, r in enumerate(rowvals)
+                if r in static_indices_w
```
FFroehlich, line 1744: 👍

FFroehlich (approved):
👍 neat!

src/model.cpp
```diff
@@ -2901,25 +2930,29 @@ void Model::fdwdp(realtype const t, realtype const* x) {
...
         fdwdx_rowvals(dwdx_hierarchical_.at(0));
         fdwdx(
             dwdx_hierarchical_.at(0).data(), t, x,
             state_.unscaledParameters.data(), state_.fixedParameters.data(),
             state_.h.data(), derived_state_.w_.data(), state_.total_cl.data(),
-            state_.spl_.data()
+            state_.spl_.data(), include_static
         );
 
         for (int irecursion = 1; irecursion <= w_recursion_depth_;
```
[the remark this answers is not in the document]
dweindl: Right, but the default value is `include_static=True`, so it's not necessary to specify. I would hope that we rather remove matlab code generation than implementing the static/dynamic distinction in matlab.

python/sdist/amici/de_export.py
```diff
@@ -1638,6 +1646,132 @@ def free_symbols(self) -> set[sp.Basic]:
...
+                tmp = nonzero_dwdw.dot(tmp)
+                res += np.sum(tmp, axis=1)
+            self._static_indices[name] = (
+                np.argwhere(res == 0).flatten().tolist()
+            )
+
+            return self._static_indices[name]
+
+        if name in ("dwdw", "dwdx", "dwdp"):
+            static_indices_w = set(self.static_indices("w"))
```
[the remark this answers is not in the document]
dweindl: Currently they are cached, but I didn't profile it yet.

python/sdist/amici/de_export.py
```diff
@@ -3605,9 +3739,61 @@ def _get_function_body(
...
+                                "    // static expressions",
+                                "    if (include_static) {",
+                                *tmp_lines,
+                                "    }",
+                            ]
+                        )
+
+                # dynamic expressions
+                if len(dynamic_idxs := self.model.dynamic_indices(function)):
+                    tmp_symbols = sp.Matrix(
```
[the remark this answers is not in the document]
dweindl: True. But currently I don't think that moving those lines to some separate function will make it clearer or more maintainable, rather the opposite.

dweindl:
Still something wrong here. The number of nonlinear convergence failures is significantly higher than before this change. :/

dweindl:
> For a model with nw: 227, nx: 67, ndwdx: 759, ndwdw: 471:
> 
>     * determining static expressions add 10 seconds to the previous 5 seconds for full code generation :(
> 
>     * no observable difference in simulations times
> 
> 
> Looks like some further profiling/optimization is required...

Okay, for that model, I am now down to no noticeable overhead and would merge those changes.

FFroehlich (approved):
👍

Issue #2306, opened 2024-02-23
Refactor de_export.py


dweindl:
Currently 4K+ lines and somewhat difficult to follow through. Some functionality should be moved to separate modules.

E.g.:

* DEModel -> de_model.py (possibly de_model.py -> de_model_components.py)
* various sympy -> sympy_utils.py

PR #2311, opened 2024-02-23
Refactor DEExporter/DEModel/csc_matrix

dweindl:
Reduce unnecessary coupling:

* `csc_matrix` as free function - removes the need for the codeprinter in `DEModel`
* Move the codeprinter from `DEModel` to `DEExporter` where it's actually needed

Not other changes in functionality.

Related to #2306

### DIFF
--- a/python/sdist/amici/pysb_import.py
+++ b/python/sdist/amici/pysb_import.py
@@ -178,6 +178,10 @@ def pysb2amici(
         compiler=compiler,
         generate_sensitivity_code=generate_sensitivity_code,
     )
+    # Sympy code optimizations are incompatible with PySB objects, as
+    #  `pysb.Observable` comes with its own `.match` which overrides
+    #  `sympy.Basic.match()`, breaking `sympy.codegen.rewriting.optimize`.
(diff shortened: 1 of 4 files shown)


python/sdist/amici/pysb_import.py
```diff
@@ -178,6 +178,10 @@ def pysb2amici(
         compiler=compiler,
         generate_sensitivity_code=generate_sensitivity_code,
     )
+    # Sympy code optimizations are incompatible with PySB objects, as
+    #  `pysb.Observable` comes with its own `.match` which overrides
+    #  `sympy.Basic.match()`, breaking `sympy.codegen.rewriting.optimize`.
+    exporter._code_printer._fpoptimizer = None
```
FFroehlich, line 184: :( sounds like this is something we want to get fixed. We could convert everything to expressions.

FFroehlich (approved):

python/sdist/amici/pysb_import.py
dweindl: Ideally there wouldn't be any PySB objects anymore at the stage of code generation.

PR #2313, opened 2024-02-23
Cache remote files for tests

dweindl:
Cache remote files for tests locally via pooch, and cache the pooch cache for GitHub actions.

Today, I had dozens of workflow failures due to rate-limiting or unavailability of biomodels. This is avoidable.


FFroehlich (approved):
👍

PR #2338, opened 2024-03-04
Fix uncaught C++ exception in runAmiciSimulation

### LINKED ISSUE #1882: Uncaught C++ exceptions crash python (sometimes)
It happened a couple of times to me, that an uncaught C++ exception crashed python. Happened for sigma <= 0 errors. Need to check whether this occurs specifically with OpenMP or this is a general problem, but needs to be checked.

dweindl:
Previously, an exception during `ReturnData::processSimulationObjects` would result in program termination.

Fixes #1882.

### DIFF
--- a/src/amici.cpp
+++ b/src/amici.cpp
@@ -220,10 +220,20 @@ std::unique_ptr<ReturnData> runAmiciSimulation(
         );
     }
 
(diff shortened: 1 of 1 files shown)


FFroehlich (approved):
🤔

Issue #2327, opened 2024-02-28
Handling (non-negativity) constraints


dweindl:
In many applications, our models are subject to physical constraint, such that the model state has to be non-negative. Due to numerical issues, this is however not always ensured in practice. Currently, this is optionally handled via `Model.{setStateIsNonNegative|allStatesNonNegative}`, which boils down to taking `max(0, state)`.

As of [v6.2.0](https://github.com/LLNL/sundials/releases/tag/v6.2.0)[1], SUNDIALS supports handling such [constraints via projection functions](https://sundials.readthedocs.io/en/latest/cvode/Mathematics_link.html#ivps-with-constraints). This would be the preferable mechanism for handling non-negativity- or more general constraints.

~~This will, however, first require upgrading SUNDIALS (#1565).~~

~~[1] Note: relevant memory leaks fixed in [v6.4.0](https://github.com/LLNL/sundials/releases/tag/v6.4.0)~~

dweindl:
Okay, after a brief test, this won't help with cases where the Jacobian cannot be evaluated for negative `x`.

dweindl:
Note: 
> CVODES in function cvInitialSetup : Constraints can not be enforced while forward sensitivity is used with simultaneous method

PR #2340, opened 2024-03-04
Enable setting constraints on state variables

dweindl:
In addition to the current `Model.setStateIsNonNegative`, this adds the option to set additional (non)negativity/positivity constraints to be enforced by the solver.

See [CVodeSetConstraints](https://sundials.readthedocs.io/en/latest/cvode/Usage/index.html#c.CVodeSetConstraints) for details.

Related to #2327.

### DIFF
--- a/python/tests/test_sbml_import.py
+++ b/python/tests/test_sbml_import.py
@@ -722,3 +722,49 @@ def test_hardcode_parameters(simple_sbml_model):
             constant_parameters=["p1"],
             hardcode_symbols=["p1"],
         )
+
+
+def test_constraints():
+    """Test non-negativity constraint handling."""
+    from amici.antimony_import import antimony2amici
+
+    ant_model = """
+    model test_non_negative_species
+        species A = 10
+        species B = 0
+        # R1: A => B; k1f * sqrt(A)
+        R1: A => B; k1f * max(0, A)
+        k1f = 1e10
+    end
+    """
+    module_name = "test_non_negative_species"
+    with TemporaryDirectory(prefix=module_name) as outdir:
+        antimony2amici(
+            ant_model,
+            model_name=module_name,
+            output_dir=outdir,
+        )
+        model_module = amici.import_model_module(
+            module_name=module_name, module_path=outdir
+        )
+        amici_model = model_module.getModel()
+        amici_model.setTimepoints(np.linspace(0, 100, 200))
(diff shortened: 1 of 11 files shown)


python/tests/test_sbml_import.py
```diff
@@ -722,3 +722,50 @@ def test_hardcode_parameters(simple_sbml_model):
...
+        amici_model.setTimepoints(np.linspace(0, 100, 200))
+        amici_solver = amici_model.getSolver()
+        rdata = amici.runAmiciSimulation(amici_model, amici_solver)
+        assert rdata.status == amici.AMICI_SUCCESS
+        # should be non-negative in theory, but is expected to become negative
+        #  in practice
+        assert np.any(rdata.x < 0)
+
+        amici_solver.setRelativeTolerance(1e-14)
+        amici_solver.setConstraints([1.0, 1.0])
```
FFroehlich, line 761: can we add an enum for `0.0`/`1.0`/`2.0`/`-1.0`/`-2.0` values for a bit more user convenience??

FFroehlich: also, probably makes sense to automatically add constraints when using `Model::setStateIsNonNegative` and add this behaviour to the documentation of setConstraints and setStateIsNonNegative.

FFroehlich (approved):

python/tests/test_sbml_import.py
dweindl: Added a python enum for now.

dweindl: > can we add an enum for `0.0`/`1.0`/`2.0`/`-1.0`/`-2.0` values for a bit more user convenience??

done

> also, probably makes sense to automatically add constraints when using `Model::setStateIsNonNegative` and add this behaviour to the documentation of setConstraints and setStateIsNonNegative.

I am not completely sure about this one. While I agree that it makes sense to imply the constraints in `setStateIsNonNegative` `True`, this would lead to some funny behavior when disabling non-negativity. 

```
# I want to enable constraints:
model.setConstraints([non_negative])
# I want to enable the max(0, state) in addition:
model.setStateIsNonNegative([True])
# I decide to disable the latter
model.setStateIsNonNegative([False]) # implies setConstraints([none]) ?!
# now it also cleared my original constraint
```

I think I'd prefer to keep them independent for now.

PR #2345, opened 2024-03-04
Fix petab import: initial assignment targets as constant parameters

dweindl:
During PEtab import, parameters that are targets of initial assignments have so far not been turned into constant parameters, because they didn't exist in the amici model (see #2304).
Now that those parameters remain in the model, they should be turned into constant parameters, unless specified otherwise.

Requires https://github.com/PEtab-dev/libpetab-python/pull/248, otherwise PEtab parameter mapping will provide potentially incorrect values for those parameters.

### DIFF
--- a/python/sdist/amici/petab/sbml_import.py
+++ b/python/sdist/amici/petab/sbml_import.py
@@ -520,16 +520,9 @@ def _get_fixed_parameters_sbml(
 
     # exclude targets of rules or initial assignments that are not numbers
     sbml_model = petab_problem.model.sbml_model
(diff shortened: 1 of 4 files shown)


FFroehlich (approved):

FFroehlich:
On second though, maybe add the libsbml parser settings as module variable and import rather than duplicating?

dweindl:
> On second thought, maybe add the libsbml parser settings as module variable and import rather than duplicating?

Would like to, but it takes a specific model as first argument. The best thing I could do, would be adding a function that creates a parser, I think, to not duplicate the settings.

Issue #2246, opened 2023-12-18
In case of simulation failures, store time point of failure in ReturnData


dweindl:
Makes live easier when trying to analyze failures. Distinguish between fwd and bwd failures.

EDIT: it might be sufficient to make ExpData::t_ public.

dweindl:
Also, the corresponding model state would ideally be accessible for debugging.

dweindl:
Maybe via storing `SimulationState` and `ModelStateDerived` in `ReturnData`.

PR #2346, opened 2024-03-04
Add Solver::get_current_time

dweindl:
... to get the current integration time from the solver.

(also fixes a couple of incorrect exception types and messages)

Prerequisite for #2246.


dweindl:
Wait, there is already `Solver::gett`. This should produce the same results.

PR #2356, opened 2024-03-05
Fix venv related installation issues

### LINKED ISSUE #2355: Fix buildAmici.sh & installAmiciSource.sh to work with system-python
Currently, if `Python3_EXECUTABLE` points to the system-python, there is a problem with installing packages. We need to create a venv from CMake initially and after that, only use the python executable from there.

FFroehlich:
- creates a venv before running cmake to avoid issues in externally managed environments, fixes #2355
- replaces a stray `pip` with `pip3`

### DIFF
--- a/.github/workflows/test_petab_test_suite.yml
+++ b/.github/workflows/test_petab_test_suite.yml
@@ -57,7 +57,7 @@ jobs:
 
       - name: Install petab
         run: |
-          source ./build/venv/bin/activate \
+          source ./venv/bin/activate \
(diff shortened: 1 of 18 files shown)


dweindl (approved):
The current cmake changes will break installation for everybody relying on specifying BLAS settings via environment variables.

I think I'd prefer fixing the CMake stuff in a separate PR (-> https://github.com/AMICI-dev/AMICI/pull/2357).

dweindl:
> The current cmake changes will break installation for everybody relying on specifying BLAS settings via environment variables.

See Windows tests.

PR #2362, opened 2024-03-06
Fix Solver::dky_ initialization

dweindl:
`Solver::dky_` is always empty. Any call to `CVodeGetDky` should fail. Seems like it's never used.


FFroehlich (approved):
🤔

PR #2400, opened 2024-04-16
Fix ModuleNotFoundError in tox -e doc

dweindl:
`import exhale_multiproject_monkeypatch` fails on some systems due to unclear sys.path issues. Therefore, we need to adjust sys.path manually.


FFroehlich (approved):
🙈

PR #2412, opened 2024-04-23
zip(... , strict=True)

### LINKED ISSUE #2410: zip(... , strict=True)
With Python3.10 (#2039) we can use `zip(... , strict=True)`. We should use that everywhere. I'm not aware of a case where we don't assume equal lengths.

dweindl:
Closes #2410.


FFroehlich (approved):

dweindl:
Okay, in some places we seem to rely on different lengths. Related to DAEs and not well documented... :grimacing: TODO

PR #2421, opened 2024-05-02
Add stacklevel to warnings

dweindl:
Make it easier to find out where warnings originate from.

### DIFF
--- a/python/sdist/amici/logging.py
+++ b/python/sdist/amici/logging.py
@@ -175,7 +175,8 @@ def get_logger(
     elif kwargs:
         warnings.warn(
             "AMICI logger already exists, ignoring keyword "
-            "arguments to setup_logger"
+            "arguments to setup_logger",
+            stacklevel=2,
         )
 
     logger = logging.getLogger(logger_name)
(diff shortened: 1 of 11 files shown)


FFroehlich:
Hmm, is it not possible to manage this using a context manager? Seems rather difficult to keep stack levels cohesive throughout the package.

dweindl:
> Hmm, is it not possible to manage this using a context manager? Seems rather difficult to keep stack levels cohesive throughout the package.

Finding the precise frame / code location where some deprecated argument is introduced is overkill in my opinion.

For the module-level deprecation warnings level=2 always makes sense, because this will show the location of the deprecated import statement.

FFroehlich:
> > Hmm, is it not possible to manage this using a context manager? Seems rather difficult to keep stack levels cohesive throughout the package.
> 
> Finding the precise frame / code location where some deprecated argument is introduced is overkill in my opinion.
> 
> For the module-level deprecation warnings, `stacklevel=2` always makes sense, because this will show the location of the deprecated import statement.

Yes, my concern is more that the next time we add a warning, we will forget to add the `stacklevel=2` argument.

dweindl:
> my concern is more that the next time we add a warning, we will forget to add the `stacklevel=2` argument.

Oh, no worries, I'm happy to add a reminder: https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/

FFroehlich (approved):

PR #2423, opened 2024-05-06
Remove wurlitzer dependency

### LINKED ISSUE #2409: Remove wurlitzer dependency
Afaik, when using the Python interface, we don't rely on C output anywhere anymore. If so, we don't need wurlitzer anymore. Double-check.

dweindl:
When using the Python interface, we don't rely on C(++) output anywhere anymore, and thus, we don't need wurlitzer anymore.

Closes #2409


FFroehlich (approved):
👍

PR #2425, opened 2024-05-06
Support event-assignments to compartments that aren't rate rule targets

### LINKED ISSUE #2424: Support event-assignments to compartments
Not supported yet:

https://github.com/AMICI-dev/AMICI/blob/8c3b59ac9f47f798d07cdeef79ddf6df6c36d5b4/python/sdist/amici/sbml_import.py#L1630-L1635

Test case: https://github.com/sbmlteam/sbml-test-suite/blob/release/cases/semantic/00945/

dweindl:
Event-assignments to compartments weren't supported yet, now they are.

Compartments that are targets of event assignments are now processed as "species", the same way as it is already done for parameters that are targets of event assignments.

Closes #2424.

### DIFF
--- a/python/sdist/amici/sbml_import.py
+++ b/python/sdist/amici/sbml_import.py
@@ -1644,7 +1644,9 @@ class SbmlImporter:
             # parse the boluses / event assignments
             bolus = [get_empty_bolus_value() for _ in state_vector]
             event_assignments = event.getListOfEventAssignments()
-            compartment_event_assignments = set()
+            compartment_event_assignments: set[tuple[sp.Symbol, sp.Expr]] = (
+                set()
+            )
             for event_assignment in event_assignments:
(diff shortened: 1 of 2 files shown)


python/sdist/amici/sbml_import.py
```diff
@@ -1631,7 +1662,7 @@ def get_empty_bolus_value() -> sp.Float:
                         "Could not process event assignment for "
                         f"{str(variable_sym)}. AMICI currently only allows "
                         "event assignments to species; parameters; or, "
-                        "compartments with rate rules, at the moment."
+                        "compartments, at the moment."
```
FFroehlich, line 1665:
```suggestion
                        "compartments."
```
already have "currently" at the beginning of the sentence.

FFroehlich (approved):

PR #2431, opened 2024-05-07
Fix execution of multiple event assignments from the same event

### LINKED ISSUE #2430: Execution of multiple event assignments from the same event
Multiple event assignments from the same event should be executed simultaneously. From the SBML L3V2 specs:

>  In other words, a single event cannot have multiple EventAssignment children
    assigning the same variable. (All of them would be performed at the same time, when that particular
    Event triggers, resulting in indeterminacy.)

However, that doesn't seem to be the case:

https://github.com/AMICI-dev/AMICI/blob/da19a550688e8232b23ace6d06722c66deb2703f/python/sdist/amici/sbml_import.py#L1670-L1673

I'd say, in the `else` case, that assignment should be ignored.

Example: Assuming a model with a concentration-based species S in compartment C, and an event with two event assignments – EA_S assigning to S and EA_C assigning to C. 
In this case, S should only change due to EA_S, but not due to the volume change caused by EA_C. Currently, first EA_S is applied, then EA_C. 

Correct?

dweindl:
~~Fixes #2430~~

### DIFF
--- a/python/sdist/amici/sbml_import.py
+++ b/python/sdist/amici/sbml_import.py
@@ -1689,7 +1689,7 @@ class SbmlImporter:
 
             # Update the concentration of species with concentration units
             # in compartments that were affected by the event assignments.
-            for compartment_sym, formula in compartment_event_assignments:
+            for compartment_sym, new_size in compartment_event_assignments:
                 for species_sym in concentration_species_by_compartment[
                     compartment_sym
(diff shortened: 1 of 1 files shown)


dweindl:
Invalid. See https://github.com/AMICI-dev/AMICI/issues/2430#issuecomment-2098463685

PR #2443, opened 2024-05-14
CMake: Fix MKL detection when not using environment modules

### LINKED ISSUE #2441: Compilation fails with Intel OneApi MKL 2024.1
I expected the build script to find the includes and libraries to link, including the linking commands.

```
/usr/bin/ld: /home/lars/Git/AMICI/build/libamici.a(cblas.cpp.o): in function `amici::amici_dgemm(amici::BLASLayout, amici::BLASTranspose, amici::BLASTranspose, int, int, int, double, double const*, int, double const*, int, double, double*, int)':
/home/lars/Git/AMICI/src/cblas.cpp:27:(.text+0x5): undefined reference to `cblas_dgemm'
...
```

**To Reproduce**
*Steps to reproduce the behavior*

```
sudo apt install intel-oneapi-mkl-devel
./scripts/buildSuiteSparse.sh
./scripts/buildSundials.sh
./scripts/buildAmici.sh
```

 - OS and version: Ubuntu 24.04
 - AMICI interface: C++, but actually I want to build the amici Python Interface via `pip install amici`
 - AMICI version: master / v0.25.0 (tested on both)
 - Additional information: gcc (Ubuntu 13.2.0-23ubuntu4) 13.2.0, python 3.12, intel oneapi mkl (blas)
 
```
MKLROOT=/opt/intel/oneapi/mkl/2024.1
LIBRARY_PATH=/opt/intel/oneapi/tbb/2021.12/env/../lib/intel64/gcc4.8:/opt/intel/oneapi/mkl/2024.1/lib/:/opt/intel/oneapi/compiler/2024.1/lib
LD_LIBRARY_PATH=/opt/intel/oneapi/tbb/2021.12/env/../lib/intel64/gcc4.8:/opt/intel/oneapi/mkl/2024.1/lib:/opt/intel/oneapi/compiler/2024.1/opt/compiler/lib:/opt/intel/oneapi/compiler/2024.1/lib:/usr/local/cuda-12.4/lib64
CPATH=/opt/intel/oneapi/tbb/2021.12/env/../include:/opt/intel/oneapi/mkl/2024.1/include:/opt/intel/oneapi/mkl/2024.1/include:/opt/intel/oneapi/tbb/2021.12/env/../include:/opt/intel/oneapi/mkl/2024.1/include:

 ```

**How to fix**
*Do you know how to resolve the problem?*

```
find_package(MKL CONFIG REQUIRED PATHS $ENV{MKLROOT})
  target_link_libraries(BLAS INTERFACE
    mkl_scalapack_ilp64
    mkl_cdft_core
    mkl_intel_ilp64
    mkl_intel_thread
    mkl_core
    mkl_blacs_intelmpi_ilp64
    iomp5
  ) 
```

*Can you submit a pull request?*
I don't know enough about the amici library and the reasons for the design of the `AmiciFindBLAS.cmake` file to open a pull request. Maybe after some feedback, I could open one. I would try to rely more on
`find(BLAS ...)`[https://cmake.org/cmake/help/latest/module/FindBLAS.html] or `find(MKL ...)`[https://www.intel.com/content/www/us/en/docs/onemkl/developer-guide-linux/2023-2/cmake-config-for-onemkl.html]

### PR DESCRIPTION
**Pull request by dweindl**
If MKLROOT is set (e.g., `source /opt/intel/oneapi/setvars.sh` will do that), first try FindBLAS instead of just relying on MKL_INCDIR and MKL_LIB.

Fixes #2441.

### DIFF
--- a/cmake/AmiciFindBLAS.cmake
+++ b/cmake/AmiciFindBLAS.cmake
@@ -12,18 +12,32 @@ set_property(CACHE BLAS PROPERTY STRINGS "CBLAS" "MKL" "ACCELERATE")
 
 if(${BLAS} STREQUAL "MKL" OR DEFINED ENV{MKLROOT})
   if(DEFINED ENV{MKLROOT})
-    # This is set by Environment Modules
-    message(STATUS "Using MKL_INCDIR and MKL_LIB from environment module")
     set(BLAS
         "MKL"
         CACHE STRING "BLAS library to use" FORCE)
-    set(BLAS_INCLUDE_DIRS
-        "$ENV{MKL_INCDIR}"
-        CACHE STRING "" FORCE)
-    set(BLAS_LIBRARIES
-        "$ENV{MKL_LIB}"
-        CACHE STRING "" FORCE)
+
+    # Was MKLROOT set by /opt/intel/oneapi/setvars.sh? then cmake will find it
+    message(STATUS "Trying to find BLAS based on MKLROOT=$ENV{MKLROOT}")
+    # give the user the option to override the BLA_VENDOR
+    if(NOT DEFINED BLA_VENDOR)
+      set(BLA_VENDOR Intel10_64lp)
+    endif()
+    find_package(BLAS)
+    if(BLAS_FOUND)
+      message(STATUS "Found BLAS via FindBLAS and MKLROOT")
+    else()
+      # This is set by Environment Modules and might not be compatible with
+      # FindBLAS
+      message(STATUS "Using MKL_INCDIR and MKL_LIB from environment module")
+      set(BLAS_INCLUDE_DIRS
+          "$ENV{MKL_INCDIR}"
+          CACHE STRING "" FORCE)
+      set(BLAS_LIBRARIES
+          "$ENV{MKL_LIB}"
+          CACHE STRING "" FORCE)
+    endif()
   else()
+    message(STATUS "BLAS is set to MKL, but MKLROOT is not set.")
     set(BLAS_INCLUDE_DIRS
         ""
         CACHE STRING "")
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: cmake/AmiciFindBLAS.cmake [hunk 1]
```diff
@@ -12,18 +12,32 @@ set_property(CACHE BLAS PROPERTY STRINGS "CBLAS" "MKL" "ACCELERATE")
...
-        "$ENV{MKL_INCDIR}"
-        CACHE STRING "" FORCE)
-    set(BLAS_LIBRARIES
-        "$ENV{MKL_LIB}"
-        CACHE STRING "" FORCE)
+
+    # Was MKLROOT set by /opt/intel/oneapi/setvars.sh? then cmake will find it
+    message(STATUS "Trying to find BLAS based on MKLROOT=$ENV{MKLROOT}")
+    # give the user the option to override the BLA_VENDOR
+    if(NOT DEFINED BLA_VENDOR)
```
FFroehlich, line 22: Maybe add message about assumed default here?

cmake/AmiciFindBLAS.cmake
```diff
@@ -12,18 +12,32 @@ set_property(CACHE BLAS PROPERTY STRINGS "CBLAS" "MKL" "ACCELERATE")
...
+      set(BLA_VENDOR Intel10_64lp)
+    endif()
+    find_package(BLAS)
```
FFroehlich, line 25: Is this really only going to work if `MKLROOT` is set?

cmake/AmiciFindBLAS.cmake
dweindl: The default is that BLA_VENDOR is undefined and that any BLAS is accepted.
I add a message with the current value.

cmake/AmiciFindBLAS.cmake
dweindl: `find_package(BLAS)`? No, generally MKLROOT does not need to be set. But to use the Intel MKL, it is most likely required. I'd have to check the full cmake module to see which other environment variables might be considered, but with a default installation, it only works after setting certain env variables, and MKLROOT is one of them.

The `find_package(BLAS)` case without MKLROOT is handled further down.

FFroehlich: ah, missed the `if(DEFINED ENV{MKLROOT})`

FFroehlich (approved):

Issue #2444, opened 2024-05-14
`Benchmark Collection` workflow fails for Lang_PLOSComputBiol2024


dweindl:
`Benchmark Collection` workflow [fails](https://github.com/AMICI-dev/AMICI/actions/runs/9073914733) with: 
```
FAILED test_petab_benchmark.py::test_benchmark_gradient[Lang_PLOSComputBiol2024-True] - RuntimeWarning: divide by zero encountered in log
FAILED test_petab_benchmark.py::test_benchmark_gradient[Lang_PLOSComputBiol2024-False] - RuntimeWarning: divide by zero encountered in log
```

https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/tree/master/Benchmark-Models/Lang_PLOSComputBiol2024

Check why we try to compute `log(0)` there.

dweindl:
> Check why we try to compute `log(0)` there.

Some log-scale parameter has a nominal value of 0 (https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/blob/e50299b0ca8642a2939aabb448ab3ee7d1b5ae1f/Benchmark-Models/Lang_PLOSComputBiol2024/parameters_Lang_PLOSComputBiol2024.tsv).

dweindl:
Needs to be checked whether this is an issue in amici or petab:

```python
/home/runner/.local/lib/python3.11/site-packages/amici/petab/simulations.py:204: in simulate_petab
    fill_in_parameters(
/home/runner/.local/lib/python3.11/site-packages/amici/petab/conditions.py:76: in fill_in_parameters
    fill_in_parameters_for_condition(
/home/runner/.local/lib/python3.11/site-packages/amici/petab/conditions.py:148: in fill_in_parameters_for_condition
    map_sim_fix = {
/home/runner/.local/lib/python3.11/site-packages/amici/petab/conditions.py:149: in <dictcomp>
    key: _get_par(key, val, map_sim_fix)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

model_par = 'initial__Cyt__CCND_sim', value = 'ic_wt_Cyt__CCND'
mapping = {'_InitialConc1': 60220000000.0, '_InitialConc10': 50629691000.0, '_InitialConc11': 4648479700.0, '_InitialConc12': 242609110000.0, ...}

[...]

PR #2449, opened 2024-05-15
PEtab: fix missing parameters in fill_in_parameters

dweindl:
Allows providing only a subset of parameter to simulate_petab. The missing parameters are taken from nominalValues in the parameter table.

Fixes https://github.com/AMICI-dev/AMICI/issues/2444. Nevertheless, exclude `Lang_PLOSComputBiol2024` to save time.


FFroehlich (approved):
👍

PR #2464, opened 2024-06-25
Deprecate passing individual tables to `amici_import_petab`

### LINKED ISSUE #2460: Replace deprecated `petab.Problem.from_files` in `amici.petab.cli.import_petab._main`
Replace deprecated `petab.Problem.from_files` in `amici.petab.cli.import_petab._main`.

Also deprecate passing individual PEtab tables to `amici_import_petab` on the command line (who is doing that anyways?), but while it's still in place, check that the parameter table is present.

dweindl:
and replace deprecated `petab.Problem.from_files` in `amici.petab.cli.import_petab._main`.

Fixes #2460

### DIFF
--- a/python/sdist/amici/petab/cli/import_petab.py
+++ b/python/sdist/amici/petab/cli/import_petab.py
@@ -3,6 +3,7 @@ import argparse
 import petab
 
 from ..petab_import import import_model_sbml
+from petab.models.sbml_model import SbmlModel
 
 
 def _parse_cli_args():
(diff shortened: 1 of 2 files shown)


python/sdist/amici/petab/cli/import_petab.py
```diff
@@ -102,18 +115,42 @@ def _parse_cli_args():
     )
 
     args = parser.parse_args()
```
FFroehlich, line 117: don't we need a a `group.parse_args()` as well?

python/sdist/amici/petab/cli/import_petab.py
```diff
@@ -102,18 +115,42 @@ def _parse_cli_args():
...
-
-    if not args.yaml_file_name and not all(
+    if any(
+        [
+            args.sbml_file_name,
```
FFroehlich, line 120: parser doesn't have them as options?

FFroehlich (approved):
Maybe add tests for both calls?

python/sdist/amici/petab/cli/import_petab.py
dweindl: No, that's happening automatically

python/sdist/amici/petab/cli/import_petab.py
dweindl: It does, all arguments from `parser.add_argument_group().add_argument()` will be available through that  `parser`.

FFroehlich: ah :)

dweindl:
> Maybe add tests for both calls?

the yaml-based one is used for the benchmark collection, added one for the deprecated call

Issue #2458, opened 2024-06-16
Remove deprecated functionality for PEtab import from individual files instead of petab.Problem


dweindl:
Remove deprecated functionality for PEtab import from individual files instead of petab.Problem:

https://github.com/AMICI-dev/AMICI/blob/3fbb90be64a1aafd8d0ed61404a60f867a42a584/python/sdist/amici/petab/sbml_import.py#L32-L37

#2456 showed that we need the parameter table for proper PEtab import, but there is no option to pass the parameter table as individual file (since in the early days, we didn't (think we'd) need it). To avoid any surprises and wrong gradients there, I'd suggest to remove this functionality as part of fixing #2455. 

There has been a DeprecationWarning in place for almost 2 years (https://github.com/AMICI-dev/AMICI/pull/1810) and I think this isn't used anyways.

Independent of this specific issue, I think it would be good to allow removing deprecated functionality not only in major releases, but also in minor releases after a certain transition period (maybe 1y after the first release containing the deprecation warning? after 3 minor releases? ...? See also, for example, [numpy's policy](https://numpy.org/neps/nep-0023-backwards-compatibility.html#implementing-deprecations-and-removals)). Opinions? @FFroehlich?

PR #2465, opened 2024-06-25
Add deprecation policy

dweindl:
As suggested in https://github.com/AMICI-dev/AMICI/issues/2458.

Other opinions?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,6 @@
 # Changelog
 
-## Deprecation policy
-
-AMICI aims to provide a stable API for users. However, not all features can be
-maintained indefinitely. We will deprecate features in minor releases and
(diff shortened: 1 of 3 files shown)


dilpath (approved):
Looks good, could also add this outside of the changelog, e.g. https://amici.readthedocs.io/en/latest/development.html

dweindl:
> could also add this outside of the changelog, e.g. https://amici.readthedocs.io/en/latest/development.html

Yes, I wasn't sure where to put it. That document says `This document contains information for AMICI developers, not too relevant to regular users.`, so I don't think it's the right place. In the changelog it's the most visible, but I can also add a separate document on versioning, deprecation, ... policy.

FFroehlich (approved):

PR #2483, opened 2024-07-08
Handle reserved names during code-printing

### LINKED ISSUE #2461: Handle reserved names in PEtab parameter mapping
Some model entity IDs are replaced by `amici_{actual_id}` during model import, but the current PEtab parameter mapping generated by `amici.petab.parameter_mapping.create_parameter_mapping` will contain the original IDs instead of the amici-model-IDs. This needs to be changed.

See https://github.com/ICB-DCM/pyPESTO/issues/1418

dweindl:
Avoid symbolic replacements for handling reserved symbols in amici.
Instead, only handle them during code printing. This has the advantage that
users won't be bothered by changed IDs.
This, however, makes it a breaking change for anybody currently relying on "amici_*" entity IDs.
I don't expect this to be much of a problem, since probably most users would have renamed their model
entities to avoid this "amici_"-prefixing.

Fixes https://github.com/AMICI-dev/AMICI/issues/2461.


dweindl:
This won't work for `t` (amici's time symbol) :(

PR #2503, opened 2024-09-25
ReturnData: remove protected data members

### LINKED ISSUE #2348: ReturnData: remove protected data members
Currently, ReturnData has a number of protected data members that are used as temporary storage for model quantities during collecting the output data. It's currently rather intransparent which values are used when. They also waste some memory.
I'd suggest to remove them and rather pass `SimulationState` through a couple of functions.

dweindl:
Removes most protected  data members from `ReturnData`, in particular sundials objects. 

Makes the code slightly easier to follow and we won't have to worry about managing SUNContext outside of runAmiciSimulation with sundials>=6.0.

Also saves some unnecessary copying of various data.

Closes #2348


FFroehlich (approved):
👍

Issue #1164, opened 2020-07-19
Use SUNLinSolWrapper in NewtonSolver


dweindl:
We have C++ wrappers for the Sundials linear solvers, but the `NewtonSolver`s still use the raw objects. This makes those classes unnecessarily complex. Should use those wrappers there.

paulstapor:
That one is already part of #1073 , but makes sense to keep this in an issue of its own, as this one is larger.

PR #2521, opened 2024-09-30
Refactor NewtonSolver 

dweindl:
Use SUNLinSolWrapper. Get rid of subclasses. Simplify.

Closes https://github.com/AMICI-dev/AMICI/issues/1164.

We should also be able to get rid of the remaining data members,
but that's for another time...

### DIFF
--- a/src/newton_solver.cpp
+++ b/src/newton_solver.cpp
@@ -10,56 +10,31 @@
 
 namespace amici {
 
-NewtonSolver::NewtonSolver(Model const& model)
+NewtonSolver::NewtonSolver(Model const& model, LinearSolver linsol_type)
     : xdot_(model.nx_solver)
     , x_(model.nx_solver)
     , xB_(model.nJ * model.nx_solver)
-    , dxB_(model.nJ * model.nx_solver) {}
-
-std::unique_ptr<NewtonSolver>
-NewtonSolver::getSolver(Solver const& simulationSolver, Model const& model) {
-
-    std::unique_ptr<NewtonSolver> solver;
-
-    switch (simulationSolver.getLinearSolver()) {
-
-    /* DIRECT SOLVERS */
-    case LinearSolver::dense:
-        solver.reset(new NewtonSolverDense(model));
-        break;
-
-    case LinearSolver::band:
-        throw NewtonFailure(AMICI_NOT_IMPLEMENTED, "getSolver");
-
-    case LinearSolver::LAPACKDense:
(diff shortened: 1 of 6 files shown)


src/newton_solver.cpp
```diff
@@ -105,144 +78,71 @@ void NewtonSolver::computeNewtonSensis(
...
+    auto status = linsol_->solve(rhs.getNVector(), rhs.getNVector(), 0.0);
     if (status != SUNLS_SUCCESS)
-        throw NewtonFailure(status, "SUNLinSolSolve_Dense");
+        throw NewtonFailure(status, "SUNLinSolSolve");
 }
 
-void NewtonSolverDense::reinitialize() {
-    /* dense solver does not need reinitialization */
+void NewtonSolver::reinitialize() {
+    // dense solver does not need reinitialization
```
FFroehlich, line 123:
```suggestion
    // direct solvers does not need reinitialization
```
what about iterative solvers?

src/newton_solver.cpp
```diff
@@ -105,144 +78,71 @@ void NewtonSolver::computeNewtonSensis(
...
 };
 
-bool NewtonSolverDense::is_singular(Model& model, SimulationState const& state)
+bool NewtonSolver::is_singular(Model& model, SimulationState const& state)
     const {
+    if (auto s = dynamic_cast<SUNLinSolKLU const*>(linsol_.get())) {
+        return s->is_singular();
+    }
+
     // dense solver doesn't have any implementation for rcond/condest, so use
```
FFroehlich, line 140: what about other direct/implicit solvers?

FFroehlich (approved):
neat!

src/newton_solver.cpp
dweindl: They are already filtered out in the constructor: https://github.com/AMICI-dev/AMICI/blob/12ee3ebbcb35e7536cd4d5c3b6447efc93a79d9b/src/newton_solver.cpp#L19-L30

Some methods will have to be updated when more solvers are to be supported.

src/newton_solver.cpp
dweindl: As above -- they haven't been supported before this PR, and they are still not supported yet. Although it would be pretty easy to extend that now....

src/newton_solver.cpp
dweindl: Added comment+exception regarding the need for updating those two functions to support additional solvers.

PR #2553, opened 2024-10-19
add 3.13 to test matrix

FFroehlich:
python 3.13 was released on Oct 7, 2024: https://docs.python.org/3.13/whatsnew/3.13.html


FFroehlich:
I believe we should also drop python 3.9 support, right?

dweindl (approved):
:+1:

dweindl:
> I believe we should also drop python 3.9 support, right?

We did, didn't we? If there's anything left -- yes we should.

PR #2555, opened 2024-10-19
Fix crashes when errors occur at output timepoints

### LINKED ISSUE #2491: Assertion error in ForwardProblem::getSimulationStateTimepoint
I got the following assertion error in `ForwardProblem::getSimulationStateTimepoint` when simulating [Weber_BMC2015](https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/tree/master/Benchmark-Models/Weber_BMC2015) with AMICI 0.26.1:

```
AMICI/include/amici/forwardproblem.h:200: const amici::SimulationState& amici::ForwardProblem::getSimulationStateTimepoint(int) const: Assertion `map_iter != timepoint_states_.end()' failed.
```

Needs further investigation.

Seems to be non-deterministic, but happens repeatedly.

Might be some memory corruption issue, potentially caused outside amici.

dweindl:
Fixes a bug that lead to program termination if a root-after-reinitialization error (potentially also others) occurred at an output timepoint, because an non-existing/invalid SimulationState for that timepoint was accessed. See #2491 for further details.

Fixes #2491.

Also avoid some unnecessary copying (during which previously the segfault occurred if this bug triggered in non-debug builds).

### DIFF
--- a/include/amici/forwardproblem.h
+++ b/include/amici/forwardproblem.h
@@ -444,16 +444,15 @@ class FinalStateStorer : public ContextManager {
     ~FinalStateStorer() {
         if (fwd_) {
             fwd_->final_state_ = fwd_->getSimulationState();
-            // if there is an associated data point, also store it in
+            // if there is an associated output timepoint, also store it in
             // timepoint_states if it's not present there.
             // this may happen if there is an error just at
             // (or indistinguishably before) an output timepoint
             auto final_time = fwd_->getFinalTime();
-            if (!fwd_->timepoint_states_.count(final_time) && fwd_->edata
-                && std::find(
-                       fwd_->edata->ts_.cbegin(), fwd_->edata->ts_.cend(),
-                       final_time
-                   ) != fwd_->edata->ts_.cend()) {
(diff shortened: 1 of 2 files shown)


FFroehlich (approved):

include/amici/forwardproblem.h
```diff
@@ -441,8 +442,21 @@ class FinalStateStorer : public ContextManager {
...
+            // if there is an associated data point, also store it in
+            // timepoint_states if it's not present there.
+            // this may happen if there is an error just at
+            // (or indistinguishably before) an output timepoint
+            auto final_time = fwd_->getFinalTime();
+            if (!fwd_->timepoint_states_.count(final_time) && fwd_->edata
+                && std::find(
+                       fwd_->edata->ts_.cbegin(), fwd_->edata->ts_.cend(),
+                       final_time
+                   ) != fwd_->edata->ts_.cend()) {
```
dweindl, line 456: Wait, ExpData is not sufficient, we need to check the model timepoints

PR #2556, opened 2024-10-19
Fix dangling pointer in SolverTest.SettersGettersWithSetup

dweindl:
Fixes a dangling pointer issue in SolverTest.SettersGettersWithSetup.
The SUNContext was destroyed before the associated AmiVectors which resulted in invalid reads.

### DIFF
--- a/.github/workflows/test_valgrind.yml
+++ b/.github/workflows/test_valgrind.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - master
       - update_valgrind_supp
+      - fix_valgrind_tests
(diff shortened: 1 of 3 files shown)


.github/workflows/test_valgrind.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - master
       - update_valgrind_supp
+      - fix_valgrind_tests
```
dweindl, line 7:
```suggestion
```

FFroehlich (approved):

Issue #2565, opened 2024-10-24
segfaults with swig4.3 and PYTHONFAULTHANDLER=1 


dweindl:
`scripts/run-python-tests.sh: line 22:  7262 Segmentation fault: 11  pytest --ignore-glob=*petab* --ignore-glob=*test_splines.py --durations=10 $@`

Happens randomly. So far, twice this week and only on macos. Unclear.

Being updated:

* https://github.com/AMICI-dev/AMICI/actions/runs/11496295512/job/31997628356?pr=2564
* https://github.com/AMICI-dev/AMICI/actions/runs/11483086913/job/31957694802
* https://github.com/AMICI-dev/AMICI/actions/runs/11496540268/job/31998356075?pr=2566
* https://github.com/AMICI-dev/AMICI/actions/runs/11496295512/job/31997628356
* https://github.com/AMICI-dev/AMICI/actions/runs/11496295512/job/31997630402

dweindl:
Another crash, now for `python/examples/example_splines_swameye//ExampleSplinesSwameye2003.ipynb` on Ubuntu: https://github.com/AMICI-dev/AMICI/actions/runs/11495925466/job/31996499560?pr=2563 :-/

May or may not be related.

dweindl:
This happens already during import of the amici-extension. I would currently guess that this is related to some external changes (runner image, brew packages, ...)

https://github.com/dweindl/AMICI/actions/runs/11500013663/job/32009264065:
```
Run scripts/run-python-tests.sh \
Fatal Python error: Segmentation fault

Current thread 0x00000001ef9a0f40 (most recent call first):
  File "<frozen importlib._bootstrap>", line 241 in _call_with_frames_removed
  File "<frozen importlib._bootstrap_external>", line 1233 in create_module
  File "<frozen importlib._bootstrap>", line 573 in module_from_spec
  File "<frozen importlib._bootstrap>", line 676 in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1147 in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 1176 in _find_and_load
  File "<frozen importlib._bootstrap>", line 241 in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1232 in _handle_fromlist
  File "/Users/runner/work/AMICI/AMICI/python/sdist/amici/amici.py", line 28 in <module>
  File "<frozen importlib._bootstrap>", line 241 in _call_with_frames_removed
[...]

PR #2572, opened 2024-11-05
GHA: Workaround for macos segfaults

dweindl:
Homebrew serves swig 4.3.0 which is incompatible with `filterwarnings = error` in pytest resulting in segfaults (#2565,  https://github.com/swig/swig/issues/3061). 
For now, the easiest workaround seems to be just ignoring any warnings during macos runs.

### DIFF
--- a/.github/workflows/test_python_cplusplus.yml
+++ b/.github/workflows/test_python_cplusplus.yml
@@ -293,7 +293,7 @@ jobs:
 
     - name: Python tests
       run: |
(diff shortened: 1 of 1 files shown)


dweindl:
> For now, the easiest workaround seems to be just ignoring any warnings during macos runs.

If this does not get fixed soon, we need to modify the swig-generated cpp files (base + model) to undefine `SWIG_HEAPTYPES`.

FFroehlich (approved):

PR #2611, opened 2024-11-30
add nan safe log&divide

FFroehlich:
Add guards to log & divide functions to avoid nans during gradient computation

### DIFF
--- a/python/sdist/amici/jax.template.py
+++ b/python/sdist/amici/jax.template.py
@@ -1,7 +1,8 @@
+# ruff: noqa: F401, F821, F841
 import jax.numpy as jnp
 from interpax import interp1d
 
-from amici.jax.model import JAXModel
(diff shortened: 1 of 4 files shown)


dweindl (approved):
Makes sense to have that, but I am slightly worried that this may cover up potential issues in the model.

Depending on how common these issues are, make it optional? Also fine as is for now.

FFroehlich:
> Makes sense to have that, but I am slightly worried that this may cover up potential issues in the model.
> 
> Depending on how common these issues are, make it optional? Also fine as is for now.

Rather common, happens with Bachmann, Laske, Lucarelli and Weber.

For jax we (currently) have to evaluate all possible observables for all timepoints and all loss functions as well as avoid nan values from ever occurring. There might be a smarter way of going about this, this seems reasonable to me. You will still end up with pretty big values (rather than infs/nans) anyways.

PR #2617, opened 2024-12-04
Separate pre-equilibration and dynamic simulation in jax

FFroehlich:
Prepares for timeseries implementation. Also fixes some issues with jax compilation that snuck into the develop branch.

TODO:
- [x] repair symlink (thanks for nothing pycharm)
- [x] update jax petab notebook
- [x] properly separate jax & cpp modules


dweindl (approved):
Looks good to me.

PR #2618, opened 2024-12-05
Mark JAX experimental

FFroehlich:


python/sdist/amici/jax/__init__.py
```diff
@@ -1,6 +1,21 @@
...
+experimental, the API may substantially change in the future. Use at your own risk and do not expect backward
+compatibility.
+"""
+
+from warnings import warn
 
 from amici.jax.petab import JAXProblem, run_simulations
 from amici.jax.model import JAXModel
 
+warn(
```
dweindl, line 15: Needs to be added to the ignore list in `pytest.ini`, I think.

dweindl (approved):
:+1:

PR #2619, opened 2024-12-05
State reinitialisation in JAX

FFroehlich:


python/sdist/amici/jax/petab.py
```diff
@@ -176,26 +202,53 @@ def _get_measurements(
...
                 ]
             )
+            if (
+                petab.OBSERVABLE_TRANSFORMATION
+                in self._petab_problem.observable_df
+            ):
+                trafo_map = {
+                    petab.LIN: 0,
+                    petab.LOG: 1,
+                    petab.LOG10: 2,
```
dweindl, line 227: It would be great to have some enum/constants here, or at least have a reference to where the values are defined/used.

python/sdist/amici/jax/petab.py
```diff
@@ -292,6 +345,79 @@ def load_parameters(
...
+                        ],
+                        Number,
+                    )
+                    and np.isnan(xval)
+                )
+                else xval
+                if isinstance(
+                    xval := self._petab_problem.condition_df.loc[
+                        simulation_condition, xname
+                    ],
```
dweindl, line 401:
```suggestion
                    xval,
```
?

dweindl (approved):

Issue #11, opened 2016-07-12
Separate output parameter and dynamic parameters


FFroehlich:
Currently, sensitivities are computed for all model parameters and not only for those that have nonzero state sensitivities. This does not matter too much for small systems but can become problematic for larger systems.

PR #2641, opened 2025-01-31
No flattening of timepoint specific overrides in jax

FFroehlich:


python/sdist/amici/de_model_components.py
```diff
@@ -607,6 +607,46 @@ def __init__(
...
+            unique)
+        """
+        super().__init__(identifier, name, 0.0)
+
+
+class ObservableParameter(ModelQuantity):
+    """
+    A NoiseParameter is an input variable for the computation of ``y`` that can be specified in a data-point specific
+    manner, abbreviated by ``op``. Only used for jax models.
+    """
```
dweindl, line 634: With respect to https://github.com/AMICI-dev/AMICI/issues/11, we might want to use `ObservableParameter` also for non-jax models at some point, without necessarily being able do specify them in a time-point specific manner. How about adding a `is_timepoint_specific` attribute already? To soon to worry about that?

python/sdist/amici/de_model.py
```diff
@@ -1087,6 +1103,21 @@ def _generate_symbol(self, name: str) -> None:
         """
         if name in self._variable_prototype:
             components = self._variable_prototype[name]()
+            # ensure placeholder parameters are consistently and correctly ordered
+            if name == "op":
+                components = sorted(
+                    components,
+                    key=lambda x: int(
+                        x.get_name().replace("observableParameter", "")
```
dweindl, line 1111: Shouldn't that be `get_id`? Why is the replacement necessary for sorting?

python/sdist/amici/jax/petab.py
```diff
@@ -249,94 +306,184 @@ def _get_measurements(
...
+                if col not in m or m[col].isna().all():
+                    mat_numeric = jnp.ones((len(m), n_pars[col]))
+                    par_mask = np.zeros_like(mat_numeric, dtype=bool)
+                    par_index = np.zeros_like(mat_numeric, dtype=int)
+                elif np.issubdtype(m[col].dtype, np.number):
+                    mat_numeric = np.expand_dims(m[col].values, axis=1)
+                    par_mask = np.zeros_like(mat_numeric, dtype=bool)
+                    par_index = np.zeros_like(mat_numeric, dtype=int)
+                else:
+                    split_vals = m[col].str.split(";")
```
dweindl, line 335:
```suggestion
                    split_vals = m[col].str.split(petab.C.PARAMETER_SEPARATOR)
```
?

python/sdist/amici/jax/petab.py
```diff
@@ -208,9 +236,38 @@ def _get_measurements(
...
+                if np.issubdtype(
+                    self._petab_problem.measurement_df[col].dtype, np.number
+                ):
+                    n_pars[col] = 1 - int(
+                        self._petab_problem.measurement_df[col].isna().all()
+                    )
+                else:
+                    n_pars[col] = (
+                        self._petab_problem.measurement_df[col]
+                        .str.split(";")
```
dweindl, line 262:
```suggestion
                        .str.split(petab.C.PARAMETER_SEPARATOR)
```

python/sdist/amici/pysb_import.py
```diff
@@ -646,17 +704,22 @@ def _add_expression(
...
+        y = sp.Symbol(name)
         trafo = noise_distribution_to_observable_transformation(noise_dist)
-        obs = Observable(y, name, sym, transformation=trafo)
-        ode_model.add_component(obs)
-
-        sigma_name, sigma_value = _get_sigma_name_and_value(
-            pysb_model, name, sigmas
+        # note that this is a bit iffy since we are potentially using the same symbolic identifier in expressions (w)
+        # and observables (y). This is not a problem as there currently are no model functions that use both. If this
+        # changes, I would expect symbol redefinition warnings in CPP models and overwriting in JAX models, but as both
```
dweindl, line 711: I think these warnings are easily overlooked. Maybe we should turn them into errors. But this will break the import of at least one benchmark collection model that has something named `NULL`. Fine for now.

python/sdist/amici/pysb_import.py
```diff
@@ -696,26 +759,17 @@ def _get_sigma_name_and_value(
...
-            )
-        except StopIteration:
-            raise ValueError(
-                f"value of sigma {obs_name} is not a " f"valid expression."
-            )
-        sigma_value = sigma_expr.expand_expr()
+        if sigma_name in pysb_model.expressions.keys():
+            return pysb_model.expressions[sigma_name]
+        raise ValueError(
+            f"value of sigma {obs_name} is not a " f"valid expression."
```
dweindl, line 769:
```suggestion
            f"value of sigma {obs_name} is not a valid expression."
```

dweindl (approved):
Thanks. I'd didn't check super thoroughly, but looks good to me.

python/sdist/amici/de_model_components.py
FFroehlich: Would follow YAGNI here

python/sdist/amici/de_model.py
FFroehlich: We need to convert to `int` otherwise `11` will come before `2`.

pytest.ini
```diff
@@ -26,5 +26,8 @@ filterwarnings =
     ignore:divide by zero encountered in log:RuntimeWarning
     # ignore jax deprecation warnings
     ignore:jax.* is deprecated:DeprecationWarning
+    # pandas
+    ignore:Setting an item of incompatible dtype:FutureWarning
+
```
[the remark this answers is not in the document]
FFroehlich: I wasn't able to reproduce this issue locally, in parts due to https://github.com/AMICI-dev/AMICI/issues/2565. Pretty sure this is due some third party library (likely petab)

python/sdist/amici/de_model.py
dweindl: Ah, I completely missed the `int` part.

PR #2645, opened 2025-02-16
Disable CvodeF checkpointing

dweindl:
amici wants to avoid sundials' checkpointing, that's why we set `steps` (the number of integration steps at which a checkpoint is created) for `CVodeAdjInit` to the maximum number of integration steps (`mxsteps`) for the forward problem.

However, the checkpoint is created at step `steps`, not at step `steps + 1`.

Therefore, if the forward integration takes exactly `mxsteps`, a checkpoint is still created.

This is not a problem per se, but it may trigger segfaults under circumstances not fully understood (-> https://github.com/LLNL/sundials/issues/49). Although unlikely, it still occasionally crashes long running optimizations.

This change should make sure that checkpointing never occurs.

### DIFF
--- a/src/solver_cvodes.cpp
+++ b/src/solver_cvodes.cpp
@@ -716,7 +716,7 @@ void CVodeSolver::adjInit() const {
         status = CVodeAdjReInit(solver_memory_.get());
     } else {
         status = CVodeAdjInit(
-            solver_memory_.get(), static_cast<int>(maxsteps_),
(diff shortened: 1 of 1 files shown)


dweindl:
Although I don't quite understand the details, https://github.com/LLNL/sundials/issues/49 is possibly related to the state reinitialization in amici, which may not be compatible with checkpointing.

Basically, the segfault occurs if the forward integration during the backward pass (CVodeB->CVode) takes more steps than the forward integration during the forward pass (CVodeF->CVode).

PR #2677, opened 2025-03-31
Fix incorrect sympification of '=='

dweindl:
Don't evaluate during `sympy.sympify` to prevent incorrect handling of `==`.

Workaround for https://github.com/sympy/sympy/issues/26227.

```python
In [1]: sympify('a == 1')
Out[1]: False
In [2]: sympify('a == 1', evaluate=False)
Out[2]: a = 1
```


dweindl:
`evaluate=False` breaks n-ary relational operators... :/

Issue #2146, opened 2023-07-12
Refactor SBML sympification 


dweindl:
At some point it would be convenient to replace the current SBML-sympification by https://github.com/dweindl/sbmlmath/. 
This gives us better control over the whole process and can replace the awkward `log` and `piecewise` handling, handling of SBML constants and special functions. And the initial collection of all SBML symbols would become obsolete.
Not expecting a relevant speed gain, but improved maintainability.

dweindl:
Furthermore, I think the current sympification is problematic because of differences in operator precedence between [SBML L3 formula strings](https://synonym.caltech.edu/software/libsbml/5.18.0/docs/formatted/python-api/libsbml-math.html#math-l3) and sympy (which is based on [Python](https://docs.python.org/3/reference/expressions.html#operator-precedence) rules) that are ignored in amici.

PR #2681, opened 2025-04-01
Use sbmlmath package for sympification of SBML math constructs

dweindl:
Directly parse the SBML MathML constructs without going through SBML L3 formula strings and `sympy.sympify()` due to various issues (https://github.com/AMICI-dev/AMICI/issues/2146).

### DIFF
--- a/python/sdist/amici/sbml_import.py
+++ b/python/sdist/amici/sbml_import.py
@@ -288,7 +288,7 @@ class SbmlImporter:
         compute_conservation_laws: bool = True,
         simplify: Callable | None = _default_simplify,
         cache_simplify: bool = False,
-        log_as_log10: bool = True,
+        log_as_log10: bool = None,
         generate_sensitivity_code: bool = True,
         hardcode_symbols: Sequence[str] = None,
     ) -> None:
@@ -397,9 +397,8 @@ class SbmlImporter:
                 see :meth:`amici.DEModel.__init__`
 
         :param log_as_log10:
-            If ``True``, log in the SBML model will be parsed as ``log10``
-            (default), if ``False``, log will be parsed as natural logarithm
-            ``ln``.
+            This option is deprecated and will be removed in a future version.
+            Also, this option never had any effect on model import.
 
         :param generate_sensitivity_code:
             If ``False``, the code required for sensitivity computation will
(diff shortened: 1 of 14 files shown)


python/sdist/amici/sbml_import.py
```diff
@@ -2785,58 +2801,156 @@ def _clean_reserved_symbols(self) -> None:
...
-                else:
-                    raise
-        except (sp.SympifyError, TypeError, ZeroDivisionError) as err:
-            raise SBMLException(
-                f'{ele_name} "{math_string}" '
-                "contains an unsupported expression: "
-                f"{err}."
+            except (sp.SympifyError, TypeError, ZeroDivisionError) as err:
+                raise SBMLException(
+                    f'{ele_name} "{var_or_math}" '
```
FFroehlich, line 2878: this will never be triggered from ele_name == "SymPy expression", right? Should the try claus also include "isinstance(var_or_math, sp.Basic)"?

python/sdist/amici/sbml_import.py
```diff
@@ -2785,58 +2801,156 @@ def _clean_reserved_symbols(self) -> None:
...
+            ele_name = sbml_obj.getId()
+
+            mathml = libsbml.writeMathMLWithNamespaceToString(
+                ast_node,
+                libsbml.SBMLNamespaces(
+                    self.sbml.getLevel(), self.sbml.getVersion()
+                ),
             )
+            try:
+                expr = self._mathml_parser.parse_str(mathml)
```
FFroehlich, line 2900: same as above, make this one big try clause?

FFroehlich (approved):

python/sdist/amici/sbml_import.py
dweindl: > this will never be triggered from ele_name == "SymPy expression", right? 

right. that ele_name can be removed.

> Should the try claus also include "isinstance(var_or_math, sp.Basic)"?

I don't think so. I don't recall errors during `.subs()`

python/sdist/amici/sbml_import.py
dweindl: I don't think `.subs()` will raise, I'd leave it out of there.

FFroehlich (approved):

PR #2730, opened 2025-05-16
Support for event priorities

### LINKED ISSUE #2718: SBML import: support for event priorities
So far, AMICI does not support SBML event priorities. Event priorities will be required for a robust event-based implementation of PEtab v2 timecourses (https://github.com/AMICI-dev/AMICI/issues/2662), though.

dweindl:
Support plain numeric priorities of SBML events (not the arbitrary expressions that SBML would allow).

If there are multiple event assignments to be executed at the same model time, the ones with higher priority values come first.

Closes #2718.

Also:
* remove an incorrect check for whether we support `useValueFromTriggerTime=True`

### DIFF
--- a/include/amici/event.h
+++ b/include/amici/event.h
@@ -1,6 +1,7 @@
 #ifndef AMICI_EVENT_H
 #define AMICI_EVENT_H
 
+#include "amici/defines.h"
 #include <functional>
 #include <list>
 #include <stdexcept>
@@ -26,10 +27,12 @@ class Event {
      * @brief Event constructor
      * @param id ID of the event
      * @param initial_value Initial value of the root function
+     * @param priority Priority of the event assignments
      */
-    Event(std::string id, bool initial_value)
+    Event(std::string id, bool initial_value, realtype priority)
         : id_(id)
-        , initial_value_(initial_value) {}
+        , initial_value_(initial_value)
+        , priority_(priority) {}
 
     /**
      * @brief Get the ID of the event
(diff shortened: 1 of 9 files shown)


include/amici/event.h
```diff
@@ -96,7 +116,13 @@ class EventQueue {
             throw std::runtime_error("No pending events");
         }
 
-        // TODO: FIFO for now. Implement priority handling here.
+        // Sort events by priority from high to low.
+        pending_events_.sort([](Event const& a, Event const& b) {
+            // The priority is NaN if not defined. In this case, the execution
+            // order is undefined, so this does not need any special treatment.
+            return a.get_priority() > b.get_priority();
```
FFroehlich, line 123: what about mixed defined and undefined event priories (sbml spec says we can do whathever we want with undefined prios). Also in the case of equal priority SBML spec says events should be picked randomly.

python/sdist/amici/de_model_components.py
```diff
@@ -736,6 +737,13 @@ def __init__(
         self._state_update = state_update
         self._initial_value = initial_value
 
+        if priority is not None and not priority.is_Number:
+            raise NotImplementedError(
+                "Only fixed numbers are currently supported as event priority."
```
FFroehlich, line 742:
```suggestion
                "Currently, only numeric values are supported as event priority."
```

FFroehlich (approved):

include/amici/event.h
dweindl: > what about mixed defined and undefined event priories (sbml spec says we can do whathever we want with undefined prios). 

That is already the case here.

> Also in the case of equal priority SBML spec says events should be picked randomly.

Ah, you are right. This will probably not work well with gradient-based optimization, but if SBML demands it, let's do it. It's the user's problem then.

PR #2739, opened 2025-05-18
Remove unused `ForwardProblem::dx_old_`

dweindl:
This isn't used anywhere. Should it have been used?


FFroehlich (approved):

FFroehlich:
> This isn't used anywhere. Should it have been used?

Don't think so.

PR #2741, opened 2025-05-18
 Support for `use_values_from_trigger_time=true` 

dweindl:
Implements full support for SBML's `use_values_from_trigger_time=true`.

* Add attribute to the different event classes
* Remove the old checks for whether we can ignore `use_values_from_trigger_time=true`
* Refactor EventQueue to store the model state from trigger time along with the event.


python/tests/test_events.py
```diff
@@ -985,3 +985,73 @@ def test_random_event_ordering():
...
+    """For simultaneously executed events, check that values from trigger
+    times are used to compute the state update."""
+    from amici.antimony_import import antimony2amici
+
+    model_name = "test_event_vals_trig_time"
+    with TemporaryDirectory(prefix=model_name) as outdir:
+        antimony2amici(
+            r"""
+            some_time = time
+            some_time' = 1
```
FFroehlich, line 1001: unused?

FFroehlich (approved):

python/tests/test_events.py
dweindl: Used. `some_time` is a RateRule target instead of an AssignmentRule target (the line above just just sets the initial value and is not an AssignmentRule), . Either would work, but this way, `some_time` will be in `x` instead of in `w`.

Issue #2745, opened 2025-05-20
Segfault when computing forward sensitivities for model with no state variables


dweindl:
SBML test case 01240 segfaults when requesting forward sensitivities:

```c
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007fff59291428 in N_VScaleVectorArray (nvec=1, c=0x6419800, X=0x62c1340, Z=0x0) at amici/ThirdParty/sundials/src/sundials/sundials_nvector.c:654
654       if (Z[0]->ops->nvscalevectorarray != NULL)
(gdb) bt
#0  0x00007fff59291428 in N_VScaleVectorArray (nvec=1, c=0x6419800, X=0x62c1340, Z=0x0) at amici/ThirdParty/sundials/src/sundials/sundials_nvector.c:654
#1  0x00007fff5920b7f3 in amici::CVodeSolver::sensReInit (this=0x64623e0, yyS0=...) at amici/src/solver_cvodes.cpp:603
#2  0x00007fffbc1fac8e in amici::ForwardProblem::handleEvent (this=0x12e8a00, tlastroot=@0x12e8eb0: 5.5, initial_event=false) at amici/src/forwardproblem.cpp:314
#3  0x00007fffbc1fa38b in amici::ForwardProblem::workForwardProblem (this=0x12e8a00) at amici/src/forwardproblem.cpp:189
#4  0x00007fffbc172945 in amici::runAmiciSimulation (solver=..., edata=0x0, model=..., rethrow=false) at amici/src/amici.cpp:109
```

PR #2749, opened 2025-05-20
Don't attempt FSA when nx==0

FFroehlich:
## Summary
- avoid segfault when computing forward sensitivities on models without state variables

## Testing
- `This environment doesn't have network access after setup, so Codex couldn't run certain commands. Consider configuring a setup script in your Codex environment to install dependencies.`


dweindl (approved):
Thanks, fixes #2745 indeed.

PR #2751, opened 2025-05-20
Implement SBML useValuesFromTriggerTime support

FFroehlich:
## Summary
- expose `use_values_from_trigger_time` flag in Event classes
- pass attribute when importing SBML
- drop old check rejecting such models
- evaluate event updates at trigger time when requested
- extend solver interface accordingly

## Testing
- `pytest` *(fails: command not found)*


dweindl:
Obsoleted by #2741.

Issue #2386, opened 2024-03-25
Handle events during pre-equilibration, simulation, and post-equilibration


dweindl:
During preequilibration and presimulation, amici only handles events at t=t0, the others are ignored. For debugging preequilibration failures, it would be convenient if one could switch off event handling also for regular simulations. The alternative would be re-importing the model without events, which is often inconvenient.

dweindl:
For PEtab 2.0 we will require events to be processed during preequilibration. For PEtab 1.0, this wasn't explicitly specified, but there wasn't really a reason to assume that they should not be processed.
If we want to keep backward compatibility, we'll have to add such an option to enable/disable event processing.
However, I think, it would be less surprising, and thus, preferable to change the default. I am not sure how many users were aware of events not being processed during preequilibration.

FFroehlich:
sounds fair

dweindl:
Event-handling and applying the currently implemented steady-state simplifications are mutually exclusive.
Once event-handling during pre-equilibration is implemented, and a model has events and a user wants to use Newton's method - how do we handle that? Silently ignore? Disallow? Emit a warning?
I would take that as "ignore events" (initial events should still be handled, though).
Disallowing it doesn't seem to make much sense. Those events might not even trigger during preequilibration, we can't know that, but the user may.
A warnings during every simulation will be a pain during parameter estimation.

FFroehlich:
> Event-handling and applying the currently implemented steady-state simplifications are mutually exclusive. Once event-handling during pre-equilibration is implemented, and a model has events and a user wants to use Newton's method - how do we handle that? Silently ignore? Disallow? Emit a warning? I would take that as "ignore events" (initial events should still be handled, though). Disallowing it doesn't seem to make much sense. Those events might not even trigger during preequilibration, we can't know that, but the user may. A warnings during every simulation will be a pain during parameter estimation.

Emit warning when setting that option?

dweindl:
I think this is done.

PR #2755, opened 2025-05-22
Add warnings for presimulation and steady state simulation with events

dweindl:
Currently, events aren't handled during presimulation and steady state simulation. This will change. See https://github.com/AMICI-dev/AMICI/issues/2386.


src/steadystateproblem.cpp
```diff
@@ -111,6 +111,13 @@ void SteadystateProblem::workSteadyStateProblem(
 void SteadystateProblem::workSteadyStateBackwardProblem(
     Solver const& solver, Model& model, BackwardProblem const* bwd
 ) {
+    if (model.ne > 0) {
```
FFroehlich: why specifically in the backward problem (only adjoint)?

dweindl: Yes.

PR #2772, opened 2025-06-06
Support for scipy-openblas64

### LINKED ISSUE #2736: Consider using OpenBLAS PyPI wheels
https://github.com/MacPython/openblas-libs provides [scipy-openblas64](https://pypi.org/project/scipy-openblas64/). This is used, or planned to be used, by both numpy and scipy. It would be convenient for users if AMICI could use that as well and wouldn't require an external BLAS.

I guess, on Mac, we still might want to use Accelerate, and on HPC platforms, we would still want to have the option to use some potentially more optimized BLAS.

dweindl:
Simplify amici installation. Use the BLAS library from `scipy-openblas64` (https://github.com/MacPython/openblas-libs) that is also used by numpy and scipy. No need to pre-install BLAS libraries anymore.

* On macOS, Accelerate is used and scipy-openblas64 will not be installed.
* Elsewhere, users can opt out of using scipy-openblas64, by setting `AMICI_BLAS_USE_SCIPY_OPENBLAS=FALSE`. The package will be installed in the build environment anyways, but won't be used.

Closes #2736.

### DIFF
--- a/cmake/AmiciFindBLAS.cmake
+++ b/cmake/AmiciFindBLAS.cmake
@@ -11,11 +11,15 @@ set(BLAS
 set_property(CACHE BLAS PROPERTY STRINGS "CBLAS" "MKL" "ACCELERATE")
 
 if(DEFINED ENV{AMICI_BLAS_USE_SCIPY_OPENBLAS})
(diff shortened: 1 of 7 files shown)


FFroehlich (approved):

dweindl:
Okay, probably not a good idea (yet): https://github.com/MacPython/openblas-libs/issues/203

FFroehlich (approved):
🙏

Issue #2775, opened 2025-06-09
Event-handling during preequilibration, presimulation and post-equilibration


dweindl:
For more consistency (#2386) and for PEtab v2 support (#2662), we want event-handling during preequilibration, presimulation and post-equilibration.

Rough plan:

- [x] Refactor ForwardProblem: extract some EventHandlingSimulator (class or function) that can be used in all four periods (if simulation-based) that does the event and (optionally) measurement processing. Also extract a SimulationWorkspace for the data structures currently stored in ForwardProblem that can then be passed to the different EventHandlingSimulators. (Started in #2777 for presimulation)
- [x] Do the same for BackwardProblem (Started in #2780)
- [x] Enable presimulation with ASA (#2786)
- [x] Event-handling during pre-equilibration
  - [x]  FSA, FSA (#2859)
  - [x] FSA, ASA (#2881)
  - [x] ASA, ASA (#2889)
- [ ] Event-handling during post-equilibration
  - [x] FSA (#2920)
  - [ ] ASA (#2925)

dweindl:
How do we want to deal with the case of steady-state simulations and events with known trigger times?
At least for events with a single trigger time, I'd say we should at least simulate until that event triggered.
For other events it's more difficult, since they might in principle trigger infinitely many times (say, `tan(t)`). However, they aren't handled yet as fixed-time events, so we can probably ignore that for now.

FFroehlich:
> How do we want to deal with the case of steady-state simulations and events with known trigger times? At least for events with a single trigger time, I'd say we should at least simulate until that event triggered. For other events it's more difficult, since they might in principle trigger infinitely many times (say, `tan(t)`). However, they aren't handled yet as fixed-time events, so we can probably ignore that for now.
> 
> Related to [#2673](https://github.com/AMICI-dev/AMICI/issues/2673).

If steadystate is reached before events are triggered, I‘d say stop there. Otherwise you end up with simulation routines that depend on how well your symbolic engine works, which I‘d prefer to avoid. But sounds we may want to say a word about this in the PEtab spec.

dweindl:
> If steadystate is reached before events are triggered, I‘d say stop there.

That depends on the definition of steady state 😅, but I see your point.

> But sounds we may want to say a word about this in the PEtab spec.

👍

FFroehlich:
> > If steadystate is reached before events are triggered, I‘d say stop there.
> 
> That depends on the definition of steady state 😅, but I see your point.

[...]

PR #2777, opened 2025-06-10
Handle events during presimulation

dweindl:
Related to #2775.

Implement event-handling during presimulation:
* Extract event- and data-handling from from `ForwardProblem` to a new `EventHandlingSimulator`
* Extract data members from `ForwardProblem` to a new `FwdSimWorkspace`
* Use them for simulation and event handling during both presimulation and the main simulation
* Simplify initialization logic before pre- and main simulation
* Replace some uses of `Model::t0()` by explicit time parameters

### DIFF
--- a/include/amici/forwardproblem.h
+++ b/include/amici/forwardproblem.h
@@ -5,6 +5,7 @@
 #include "amici/edata.h"
 #include "amici/misc.h"
 #include "amici/model.h"
+#include "amici/solver.h"
 #include "amici/steadystateproblem.h"
 #include "amici/vector.h"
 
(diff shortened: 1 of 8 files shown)


include/amici/forwardproblem.h
```diff
@@ -63,6 +64,257 @@ struct Discontinuity {
...
+        return gsl::narrow<int>(result.event_states_.size()) - 1;
+    }
+
+    /**
+     * @brief Creates a carbon copy of the current simulation state variables
+     * @return state
+     */
+    SimulationState getSimulationState();
+
+    /** Results for the currentb simulation period. */
```
FFroehlich:
```suggestion
    /** Results for the current simulation period. */
```

include/amici/model.h
```diff
@@ -260,16 +262,20 @@ class Model : public AbstractModel, public ModelDimensions {
 
     /**
      * @brief Initialize initial states.
+     * @param t Timepoint
```
FFroehlich: Rename to t0? Same below

src/forwardproblem.cpp
```diff
@@ -41,29 +42,118 @@ ForwardProblem::ForwardProblem(
...
+
+    // handle initial events
+    if (model_->ne && std::ranges::any_of(ws_->roots_found, [](int rf) {
+            return rf == 1;
+        })) {
+        handleEvent(true, edata);
+    }
+
+    // store initial state and sensitivity
+    result.initial_state_ = getSimulationState();
```
FFroehlich: Would argue that we should store initial state post assignment but pre event execution. Different situation with PEtab converted timecourses, but we may want to have special handling/reporting there anyways.

FFroehlich (approved):

include/amici/model.h
dweindl: `Model::{x0,sx0}` use `t`. No strong opinion.

src/forwardproblem.cpp
dweindl: I also stumbled across that, but this is how it was always done, it's only more obvious now. I mainly didn't change it because I wasn't sure whether it breaks something downstream.

I think one can argue for either. On the one hand, if the initial state is defined as $x(t_0)$, it should be post-event. It may be somewhat confusing if $x_0 = x(t_0)$ and "initial state" differ. On the other hand, we have `Model::fx0`, which in the case of events may not yield the actual `x_0`, and get/setInitialState which is currently pre-event.
It feels more intuitive to consider everything that happens at $t = t_0$ as initialization, but this is not really compatible with the rest...

No strong opinion. I can change it and try to clarify that in the documentation.

FFroehlich: okay let's keep it as is for now.

PR #2780, opened 2025-06-11
Refactor BackwardProblem for presimulation

dweindl:
Refactor BackwardProblem to implement presimulation including event-handling (#2775).

Similar to #2777 for the ForwardProblem:

* Extract some state from BackwardProblem
* Extract a EventHandlingBwdSimulator that handles simulation and discontinuities
* Those will be reused for presimulation shortly
* Remove some unused getters from BackwardProblem

### DIFF
--- a/include/amici/backwardproblem.h
+++ b/include/amici/backwardproblem.h
@@ -183,8 +183,6 @@ class BackwardProblem {
     realtype t_;
     /** sensitivity state vector array */
     AmiVectorArray sx0_;
-    /** array of number of found roots for a certain event type */
-    std::vector<int> nroots_;
     /** array containing the time-points of discontinuities*/
     std::vector<Discontinuity> discs_;
 
--- a/include/amici/backwardproblem.h
+++ b/include/amici/backwardproblem.h
@@ -8,12 +8,131 @@
 #include <vector>
 
 namespace amici {
-
 class ExpData;
 class Solver;
 class Model;
 class ForwardProblem;
 class SteadystateProblem;
+
+/**
+ * @brief The BwdSimWorkspace class is used to store temporary simulation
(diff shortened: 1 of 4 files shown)


include/amici/backwardproblem.h
```diff
@@ -8,89 +8,107 @@
...
+    /** quadrature state vector */
+    AmiVector xQB_;
 
-    /**
-     * @brief Accessor for dJydx
-     * @return dJydx
-     */
-    std::vector<realtype> const& getdJydx() const { return dJydx_; }
+    /** array of number of found roots for a certain event type */
+    std::vector<int> nroots_;
```
FFroehlich, line 42: Can we drop this in favour of computing nroots_ online from discs_? would guess compute is negligible.

include/amici/backwardproblem.h
```diff
@@ -8,89 +8,107 @@
...
+class EventHandlingBwdSimulator {
+  public:
     /**
-     * @brief Accessor for xB
-     * @return xB
+     * @brief EventHandlingBwdSimulator constructor.
+     * @param model The model to simulate.
+     * @param solver The solver to use for the simulation.
+     * @param ws The workspace to use for the simulation.
+     * (dimension nJ x nx x nMaxEvent, ordering =?)
```
FFroehlich: Does ws really have a dim?

dweindl: No... leftover...

include/amici/backwardproblem.h
dweindl: Fine for me. ~~I'd prefer doing that after merging #2777 and #2780 though, for both fwd and bwd together.~~ Nah, I'll just do it here for BackwardProblem.

PR #2782, opened 2025-06-12
Fix sensitivities for the total abundances of conservation laws

dweindl:
Fix incorrect argument order when calling `Model::fdtotal_cldx_rdata`.

This bug most probably led to incorrect sensitivities for models with conservation laws.
This should have affected only PySB models, and SBML models if the `AMICI_EXPERIMENTAL_SBML_NONCONST_CLS` environment variable was set.


FFroehlich (approved):
:(

PR #2793, opened 2025-06-17
Separate t_0 for pre-equilibration

### LINKED ISSUE #2762: Separate t_0 for pre-equilibration
So far, pre-equilibration starts at Model::t0 / ExpData::t0. In many cases the initial time won't matter. 
However, when implementing event handling during pre-equilibration (#2386), it might be more relevant to specify a different initial time for pre-equilibration and the main simulation. This will also be required for full PEtab v2 support.

dweindl:
So far, pre-equilibration started at `Model::t0`. This change allows specifying a different start time for pre-equilibration via `Model::setT0Preeq()`. The default remains starting pre-equilibration at `Model::t0`.

Closes #2762.

### DIFF
--- a/python/tests/test_preequilibration.py
+++ b/python/tests/test_preequilibration.py
@@ -698,7 +698,7 @@ def test_preequilibration_t0(tempdir):
         preeq_indicator = 0
         t0_preeq = time * preeq_indicator
         # we need some state variable for simulation to work
-        T = time
+        T = 0
         # at some large value of T, we will "reach steady state" due to
         # the tiny relative change
         T' = 1
@@ -716,10 +716,10 @@ def test_preequilibration_t0(tempdir):
     )
     amici_model = model_module.getModel()
     edata = amici.ExpData(amici_model)
-    edata.setTimepoints([0.0, 10.0])
+    edata.setTimepoints([0.0, 10_000.0])
     edata.fixedParametersPreequilibration = [1.0]
     edata.fixedParameters = [0.0]
-    amici_model.setT0Preeq(-10.0)
+    amici_model.setT0Preeq(-10_000.0)
     amici_model.setT0(-2.0)
     amici_solver = amici_model.getSolver()
     amici_solver.setRelativeToleranceSteadyState(1e-5)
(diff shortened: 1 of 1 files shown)


python/tests/test_preequilibration.py
```diff
@@ -687,3 +687,46 @@ def test_partial_eq():
...
+    )
+    model_module = amici.import_model_module(
+        module_name=module_name, module_path=tempdir
+    )
+    amici_model = model_module.getModel()
+    edata = amici.ExpData(amici_model)
+    edata.setTimepoints([0.0, 10.0])
+    edata.fixedParametersPreequilibration = [1.0]
+    edata.fixedParameters = [0.0]
+    amici_model.set_t0_preeq(-10.0)
```
FFroehlich: Can we stick to setT0Preeq for the moment?

python/tests/test_preequilibration.py
```diff
@@ -687,3 +687,46 @@ def test_partial_eq():
...
+    amici_model.setT0(-2.0)
+    amici_solver = amici_model.getSolver()
+    amici_solver.setRelativeToleranceSteadyState(1e-5)
+    amici_model.setSteadyStateComputationMode(
+        amici.SteadyStateComputationMode.integrationOnly
+    )
+
+    rdata = amici.runAmiciSimulation(amici_model, amici_solver, edata)
+    assert rdata.status == amici.AMICI_SUCCESS
+    assert set(rdata.by_id("t0_preeq")) == {-10.0}
```
FFroehlich: This only checks whether the right value was stored, but as t0_preeq doesn't influence simulation we don't really check whether that's the case. Maybe add event at -5 and ensure it fires during preeq but not simulation?

dweindl: Unfortunately, we don't have event- or piecewise-handling during preequilibration implemented yet.
I agree that better testing can be performed once that's in place.
The current test at least checks at which time the model was initialized for preequilibration.

FFroehlich: > Unfortunately, we don't have event- or piecewise-handling during preequilibration implemented yet. I agree that better testing can be performed once that's in place. The current test at least checks at which time the model was initialized for preequilibration.

Ah, how about integrating with dx/dt = 1 then and checking the integral?

dweindl: Good idea, that should work.

PR #2796, opened 2025-06-18
Fix indexing errors for adjoint quadratures during event updates

dweindl:
* Fix indexing error in xQB (related to #18)
   Now this is in line with the indexing in ReturnData https://github.com/AMICI-dev/AMICI/blob/ce28017159c9172a93b3b5ad6c602ebd3c0cd9d3/src/rdata.cpp#L497
* Fix incorrect documentation of dimensions
* Fix error message for finiteness check for deltaqB


FFroehlich (approved):
ouch :(

PR #2797, opened 2025-06-18
Fix (some) ASA event handling

dweindl:
* Fix backward simulation: In the forward simulation, measurements are handled after events. So for the backward integration, events should be handled after data points.
* Make Heaviside updates more visible
* Store Heaviside values and additional other pre-event state during event handling for the backward simulation
* Remove the now obsolete `Model::updateHeavisideB`

### DIFF
--- a/src/backwardproblem.cpp
+++ b/src/backwardproblem.cpp
@@ -129,7 +129,11 @@ void EventHandlingBwdSimulator::handleEventB(
         ws_->nroots_[ie]--;
     }
 
-    model_->updateHeavisideB(disc.root_info.data());
+    // apply pre-event state
+    auto state = model_->getModelState();
+    state.h = disc.h_pre;
+    state.total_cl = disc.total_cl_pre;
+    model_->setModelState(state);
 }
 
 void EventHandlingBwdSimulator::handleDataPointB(
@@ -218,16 +222,16 @@ void EventHandlingBwdSimulator::run(
             );
         }
 
-        // handle discontinuity
-        if (!ws_->discs_.empty() && tnext == ws_->discs_.back().time) {
-            handleEventB(ws_->discs_.back(), dJzdx);
-            ws_->discs_.pop_back();
-        }
-
         // handle data-point
         if (it >= 0 && tnext == timepoints[it]) {
             handleDataPointB(it, dJydx);
             it--;
+        }
+
+        // handle discontinuity
+        if (!ws_->discs_.empty() && tnext == ws_->discs_.back().time) {
+            handleEventB(ws_->discs_.back(), dJzdx);
+            ws_->discs_.pop_back();
         }
 
         // reinitialize state
--- a/include/amici/forwardproblem.h
+++ b/include/amici/forwardproblem.h
@@ -25,27 +25,39 @@ struct Discontinuity {
      * @brief Constructor.
      * @param time
      * @param root_info
+     * @param x_pre
      * @param xdot_pre
      * @param x_post
      * @param xdot_post
+     * @param h_pre
+     * @param total_cl_pre
      */
     explicit Discontinuity(
         realtype time, std::vector<int> const& root_info = std::vector<int>(),
+        AmiVector const& x_pre = AmiVector(),
         AmiVector const& xdot_pre = AmiVector(),
         AmiVector const& x_post = AmiVector(),
-        AmiVector const& xdot_post = AmiVector()
+        AmiVector const& xdot_post = AmiVector(),
+        std::vector<realtype> const& h_pre = std::vector<realtype>(),
+        std::vector<realtype> const& total_cl_pre = std::vector<realtype>(0)
     )
         : time(time)
         , x_post(x_post)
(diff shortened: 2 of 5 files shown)


src/backwardproblem.cpp
```diff
@@ -98,6 +98,8 @@ void BackwardProblem::handlePostequilibration() {
 void EventHandlingBwdSimulator::handleEventB(
     Discontinuity const& disc, std::vector<realtype> const* dJzdx
 ) {
+    model_->updateHeavisideB(disc.root_info.data());
```
FFroehlich: This shouldn't matter and might point to a more fundamental issue. All computations below should not depend on heaviside states and if they do that information needs to be either stored with discs (preferable) or we need heaviside updates between events (in backwards execution). Looking at `EventHandlingSimulator::handle_event`, I think we are missing heaviside updates for secondary events?

src/backwardproblem.cpp
```diff
@@ -216,18 +216,18 @@ void EventHandlingBwdSimulator::run(
             );
         }
 
-        // handle discontinuity
```
FFroehlich, line 221: in the forward pass we are processing data points and then events, right? In the backward pass we should invert order so the original code looks fine to me.

dweindl: > in the forward pass we are processing data points and then events, right?

No, the other way round: https://github.com/AMICI-dev/AMICI/blob/4eebf1613352303a97a26f393471c14924dd71aa/src/forwardproblem.cpp#L137-L141

src/backwardproblem.cpp
dweindl: Hm. At least `fdeltaqB` in `addAdjointQuadratureEventUpdate` and `fdeltaxB` in `addAdjointStateEventUpdate` have `h` in their parameter list.

dweindl: > Looking at `EventHandlingSimulator::handle_event`, I think we are missing heaviside updates for secondary events?

I don't think so. But I agree that it currently happens in an unexpected place and should probably be changed:
https://github.com/AMICI-dev/AMICI/blob/4eebf1613352303a97a26f393471c14924dd71aa/src/forwardproblem.cpp#L308-L310

dweindl: Ah, but during forward simulation it is updated before event execution, so in backward mode it was probably correct.

FFroehlich: > Hm. At least `fdeltaqB` in `addAdjointQuadratureEventUpdate` and `fdeltaxB` in `addAdjointStateEventUpdate` have `h` in their parameter list.

Yes, but I think we should provide the input `h` from model state in discs rather than spooling back updated in the backwards pass (which will fail with combinations of `useValuesFromTriggerTimer` or whatever that option is called)

> > Looking at `EventHandlingSimulator::handle_event`, I think we are missing heaviside updates for secondary events?
> 
> I don't think so. But I agree that it currently happens in an unexpected place and should probably be changed:
> 
> https://github.com/AMICI-dev/AMICI/blob/4eebf1613352303a97a26f393471c14924dd71aa/src/forwardproblem.cpp#L308-L310

Is that function called recursively? To me it looks like we only update after the first event.

dweindl: > Yes, but I think we should provide the input `h` from model state in discs rather than spooling back

Makes sense.

> Is that function called recursively? To me it looks like we only update after the first event.

Once before the first event is handled (top), then iteratively, whenever a secondary event was triggered (bottom):
https://github.com/AMICI-dev/AMICI/blob/afc84d923cdef77a6befe46b9e9a6b98910a3411/src/forwardproblem.cpp#L324-L388  


Semi-related: I am not completely sure, but my gut feeling is that the current way the discontinuities are stored will cause issues with sensitivity analysis in case of simultaneously triggered primary events, because only the combined pre/post event state (combined deltax) is available.

dweindl: > Yes, but I think we should provide the input `h` from model state in discs rather than spooling back

Done. I also already added `x_pre` and `total_cl_pre` which will be required for implementing deltaxB and deltaqB for Python-models.

FFroehlich: > > Yes, but I think we should provide the input `h` from model state in discs rather than spooling back
> 
> Makes sense.
> 
> > Is that function called recursively? To me it looks like we only update after the first event.
> 
> Once before the first event is handled (top), then iteratively, whenever a secondary event was triggered (bottom):
> 
> https://github.com/AMICI-dev/AMICI/blob/afc84d923cdef77a6befe46b9e9a6b98910a3411/src/forwardproblem.cpp#L324-L388
> 
> Semi-related: I am not completely sure, but my gut feeling is that the current way the discontinuities are stored will cause issues with sensitivity analysis in case of simultaneously triggered primary events, because only the combined pre/post event state (combined deltax) is available.

Yes that sounds plausible.

dweindl: I'll keep that for a separate PR (tracked as #2805).

FFroehlich (approved):

PR #2813, opened 2025-06-20
pre-commit updates

dweindl:
Update pre-commit hooks and re-run.

### DIFF
--- a/python/sdist/amici/numpy.py
+++ b/python/sdist/amici/numpy.py
@@ -399,7 +399,7 @@ class ExpDataView(SwigPtrView):
         """
         if not isinstance(edata, (ExpDataPtr | ExpData)):
             raise TypeError(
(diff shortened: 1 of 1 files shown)


python/sdist/amici/numpy.py
```diff
@@ -399,8 +399,7 @@ def __init__(self, edata: ExpDataPtr | ExpData):
         """
         if not isinstance(edata, (ExpDataPtr | ExpData)):
             raise TypeError(
-                f"Unsupported pointer {type(edata)}, must be"
-                f"amici.ExpDataPtr!"
+                f"Unsupported pointer {type(edata)}, must beamici.ExpDataPtr!"
```
FFroehlich:
```suggestion
                f"Unsupported pointer {type(edata)}, must be amici.ExpDataPtr!"
```

PR #2815, opened 2025-06-21
Track discontinuities in JAX & fix gradient for models with parameter dependent roots

FFroehlich:
TODO 
- [x] wait for new diffrax release, see https://github.com/patrick-kidger/diffrax/issues/654 or temp pin to main
- [x] add util wrapper to model to avoid using private API
- [x] add rootfinding for everything else or error for parameter dependent events that require rootfinding


python/sdist/amici/jax/ode_export.py
```diff
@@ -243,6 +248,9 @@ def _generate_jax_code(self) -> None:
             # tuple of variable names (ids as they are unique)
             **_jax_variable_ids(self.model, ("p", "k", "y", "w", "x_rdata")),
             "P_VALUES": _jnp_array_str(self.model.val("p")),
+            "ROOTS": _jnp_array_str(
+                {root for e in self.model._events for root in e._t_root}
```
```suggestion
                {root for e in self.model._events for root in e.get_trigger_times()}
```
?

dweindl (approved):
Looks good

PR #2826, opened 2025-06-26
Simplify SteadystateProblem

dweindl:
More transparent logic for whether we need forward sensitivities

* Remove SteadystateProblem::requires_state_sensitivities. Minimal duplication, but simpler logic in each of the two use cases.
* Get rid of SteadyStateContext


FFroehlich (approved):
🙏

PR #2840, opened 2025-06-26
Avoid events for smooth piecewise

### LINKED ISSUE #2049: piecewise without events
Currently, all piecewise functions translate to events. This wouldn't be necessary for continuous expressions and continuous 1st derivative. 

See https://stackoverflow.com/questions/55881078/how-to-know-whether-a-function-is-continuous-with-sympy and discussion in  #2048.

FFroehlich:
## Summary
- detect continuously differentiable piecewise expressions
- check smoothness w.r.t. given parameters only
- keep these piecewise expressions instead of converting to Heaviside
- extend tests for constant piecewise expressions

## Testing
- `pre-commit run --files python/sdist/amici/import_utils.py python/tests/test_heavisides.py`
- `pytest python/tests/test_heavisides.py::test_parse_piecewise_c1_no_heaviside python/tests/test_heavisides.py::test_parse_piecewise_discontinuous_to_heaviside python/tests/test_heavisides.py::test_parse_piecewise_c1_constant_zero -q`

fixes #2049 

------
https://chatgpt.com/codex/tasks/task_b_685d399bee14832b899b26d9136e2fda


FFroehlich:
currently don't see a good way of making this work

key issue is that it makes it no longer safe to substitute True/False to 1.0/0.0. don't see a nice way of fixing that since even conditions may include boolean expressions that should be interpreted as floats and only eventually get converted to booleans. 

could be workable with some creative use of try/except, @dweindl any ideas?

FFroehlich:
Comment on why this hasn't been merged yet: I currently don't really trust the C1 continuity check as assignments etc are not flattened out. This might imply that we could inadvertently miss discontinuities, but maybe it's all fine as we can assume continuity of the respective terms. Haven't thought about it enough.

Issue #2226, opened 2023-12-04
Don't handle model entity names via preprocessor


dweindl:
The preprocessor defines we are currently using to have more readable model equations are potentially problematic because they aren't scoped. This can lead to issues like: 

```
In file included from /home/runner/work/AMICI/AMICI/test_bmc/Smith_BMCSystBiol2013/stau.cpp:8:
/home/runner/work/AMICI/AMICI/test_bmc/Smith_BMCSystBiol2013/x.h:1: error: "NULL" redefined [-Werror]
    1 | #define NULL x[0]
      | 
In file included from /usr/include/stdio.h:33,
                 from /usr/include/c++/11/cstdio:42,
                 from /usr/include/c++/11/ext/string_conversions.h:43,
                 from /usr/include/c++/11/bits/basic_string.h:6608,
                 from /usr/include/c++/11/string:55,
                 from /usr/include/c++/11/bits/locale_classes.h:40,
                 from /usr/include/c++/11/bits/ios_base.h:41,
                 from /usr/include/c++/11/ios:42,
                 from /home/runner/.local/lib/python3.9/site-packages/amici/include/gsl/gsl-lite.hpp:26,
                 from /home/runner/work/AMICI/AMICI/test_bmc/Smith_BMCSystBiol2013/stau.cpp:5:
/usr/lib/gcc/x86_64-linux-gnu/11/include/stddef.h:392: note: this is the location of the previous definition
  392 | #define NULL __null
      | 
```

Would be good to come up with some other solution that allows compile-time mapping for readable names to array locations.

Related to #1544

dweindl:
One alternative could be generating, for example, a class `Parameters` that wraps the plain parameter array and has as members the names of all parameters that map to the respective array location. So what is currently written as `p1 + p2` would become `p.p1 + p.p2`. The main advantage would be, that we have everything inside the class namespace, so it's safe if somebody wants to use a parameter named `p`. Slightly more verbose, but on the positive side, it would be obvious that those are parameters. It still wouldn't work for c++ reserved names, so no parameter named `int` etc. would be supported. If we want that, then I guess there is no way around using a string-indexed map.
However, I don't know if it's possible to resolve some string-based mapping at compile time (I could imagine it's feasible, though). Furthermore, that would make the code less readable, e.g. `p["p1"] + p["p2"]`, which would be nice to avoid.

dweindl:
[...]

PR #2853, opened 2025-06-28
Handle reserved symbol names during code generation

FFroehlich:
## Summary
- avoid reserved name renaming during SBML import
- drop reserved name check in `ModelQuantity`
- sanitize symbols when printing C++ and JAX code
- adjust sbml tests for new behavior
- fix reserved symbol handling during code generation
- allow BNGL conversion for empty-compartment tests
- fix JAX SBML test module import
- revert SBMLSuiteJax import path
- use `semantic-results-jax` directory for JAX SBML suite results
- fix parameter ID names
- fix parameter ID handling
- fix JAX sanitization of variable names

## Testing
- `pre-commit run --files python/sdist/amici/jax/ode_export.py`
- `pytest python/tests/test_sbml_import.py::test_import_same_model_name -vv -s`
- `pytest python/tests/test_events.py::test_handling_of_fixed_time_point_event_triggers -q`
- `pytest tests/benchmark_models/test_petab_benchmark.py -k Boehm_JProteomeRes2014 -q`
- `pytest tests/benchmark_models/test_petab_benchmark_jax.py -k Boehm_JProteomeRes2014 -q`
- `pytest tests/sbml/testSBMLSuiteJax.py::test_sbml_testsuite_case_jax[00525] -q`

------
https://chatgpt.com/codex/tasks/task_b_685d2802e714832b968ced9c529d9411

### DIFF
--- a/python/sdist/amici/cxxcodeprinter.py
+++ b/python/sdist/amici/cxxcodeprinter.py
@@ -10,6 +10,8 @@ from collections.abc import Iterable
 import sympy as sp
 from sympy.codegen.rewriting import Optimization, optimize
 from sympy.printing.cxx import CXX11CodePrinter
+
+from .import_utils import RESERVED_SYMBOLS
 from sympy.utilities.iterables import numbered_symbols
 from toposort import toposort
 
@@ -49,6 +51,12 @@ class AmiciCxxCodePrinter(CXX11CodePrinter):
             self._fpoptimizer = lambda x: optimize(x, self.optimizations)
         else:
             self._fpoptimizer = None
+
+    def _print_Symbol(self, expr: sp.Symbol) -> str:
+        name = super()._print_Symbol(expr)
+        if name in RESERVED_SYMBOLS and name != "t":
+            return f"amici_{name}"
+        return name
 
     def doprint(self, expr: sp.Expr, assign_to: str | None = None) -> str:
         if self._fpoptimizer:
--- a/python/sdist/amici/jax/ode_export.py
+++ b/python/sdist/amici/jax/ode_export.py
@@ -28,6 +28,7 @@ from amici.de_model import DEModel
 
 from amici.de_export import is_valid_identifier
 from amici.import_utils import (
+    RESERVED_SYMBOLS,
     strip_pysb,
 )
 from amici.logging import get_logger, log_execution_time, set_log_level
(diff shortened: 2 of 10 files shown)


tests/sbml/testSBMLSuiteJax.py
```diff
@@ -25,7 +25,7 @@
 
 @pytest.fixture(scope="session")
 def result_path_jax() -> Path:
-    return Path(__file__).parent / "amici-semantic-results-jax"
+    return Path(__file__).parent / "semantic-results-jax"
```
FFroehlich:
```suggestion
    return Path(__file__).parent / "amici-semantic-results-jax"
```

scripts/run-SBMLTestsuite.sh
```diff
@@ -26,7 +26,7 @@ else
 fi
 
 if [[ "$BACKEND" == "jax" ]]; then
-  RESULT_DIR=tests/sbml/amici-semantic-results-jax
+  RESULT_DIR=tests/sbml/semantic-results-jax
```
FFroehlich:
```suggestion
  RESULT_DIR=tests/sbml/amici-semantic-results-jax
```

python/sdist/amici/jax/ode_export.py
```diff
@@ -63,11 +71,11 @@ def _jax_variable_equations(
     return {
         f"{eq_name.upper()}_EQ": "\n".join(
             code_printer._get_sym_lines(
-                (str(strip_pysb(s)) for s in model.sym(eq_name)),
+                (_sanitize(str(strip_pysb(s))) for s in model.sym(eq_name)),
                 model.eq(eq_name).subs(subs),
                 indent,
             )
-        )[indent:]  # remove indent for first line
+        )[indent:]
```
FFroehlich:
```suggestion
        )[indent:]  # remove indent for first line
```

python/sdist/amici/cxxcodeprinter.py
```diff
@@ -50,6 +52,12 @@ def __init__(self):
         else:
             self._fpoptimizer = None
 
+    def _print_Symbol(self, expr: sp.Symbol) -> str:
+        name = super()._print_Symbol(expr)
+        if name in RESERVED_SYMBOLS and name != "t":
```
dweindl, line 57: What happens if a model has a parameter `t` that is not time?

dweindl (commented):
I'd be happy to handle reserved symbols only during code generation, but I don't think this will work well in general. The problem is that we don't know anymore whether the given symbol is a user-defined symbol or an amici-internal symbol. There are quite a number of (currently unhandled) names that will cause problems, that we can't handle here (`d{}dt`, `xdot{}`, `stau{}`, any C++ keywords and typenames, any parameter names from amici's model functions, ...).
I think the most and only sustainable solution would be some string-based indexing that is evaluated at compile-time that avoids any collisions (see also https://github.com/AMICI-dev/AMICI/issues/2226).

PR #2859, opened 2025-06-30
Handle events during equilibration

dweindl:
Handle events during equilibration.

* Handle initial events independent of steady-state computation mode.
* Handle other events during integration
* TODO: ASA (separate PR)

### DIFF
--- a/python/tests/test_preequilibration.py
+++ b/python/tests/test_preequilibration.py
@@ -12,6 +12,13 @@ from amici.testing import (
     TemporaryDirectoryWinSafe as TemporaryDirectory,
     skip_on_valgrind,
 )
+from amici.gradient_check import check_derivatives
+
+pytestmark = pytest.mark.filterwarnings(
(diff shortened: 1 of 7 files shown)


python/tests/test_preequilibration.py
```diff
@@ -734,3 +741,105 @@ def test_preequilibration_t0(tempdir):
...
+    amici_solver.setSensitivityOrder(amici.SensitivityOrder.first)
+
+    for sensi_meth in [
+        amici.SensitivityMethod.forward,
+        # amici.SensitivityMethod.adjoint,
+    ]:
+        amici_solver.setSensitivityMethod(sensi_meth)
+
+        # amici_model.requireSensitivitiesForAllParameters()
+        # FIXME: sensitivities w.r.t. trigger time are off
```
FFroehlich, line 827: for forward sensi or just adjoint?

dweindl: This was for forward sensitivities. Adjoints yet to be tested.

Issue #2861, opened 2025-06-30
Revisit root-finding


dweindl:
When simulating models with discontinuities/events, one often encounters root-after-reinitialization errors, even for seemingly trivial examples (see, e.g., #2804). This should be fixable.
For example, try deactivating root functions after they triggered until we had another zero crossing in the opposite direction.

PR #2862, opened 2025-06-30
Mask root functions

dweindl:
Mask root functions once a root was found until we encounter a negative value. Related to #2861.

WIP


dweindl:
Superseded by #3266.

PR #2886, opened 2025-07-07
CI: Update model_constant_species

dweindl:
* Convert to antimony
* Parameterize initial concentration (`init_substrate`) to test non-zero initial state sensitivities


dweindl:
> @dweindl also remove the actual model file in the examples directory?

It's currently also used in the steady state notebook. Asap, after getting rid of it there (#2891). :+1:

PR #2894, opened 2025-07-09
Make `DEModel._sparsesyms` symbols `sympy.Symbol`s

### LINKED ISSUE #1061: Use symbolic comparison when writing header files in ode_export
In ode_export, symbolic comparison should be used instead of string comparison (see #1060): 
https://github.com/ICB-DCM/AMICI/blob/77dd58a3b92d43b1e0895c49a2f835c5c1181c2e/python/amici/ode_export.py#L2170
This is currently not possible, as some model quantities do not pass symbolics, but strings (e.g., dwdx).. One should find out, which routines pass string and make them pass symbolics, then change this comparisno to a symbolic one.

dweindl:
Previously, sparse matrix symbols in `DEModel._sparsesyms` were strings, as opposed to other model "syms" which were `sympy.Symbol`. Make that consistent.

Fixes #1061.

### DIFF
--- a/python/sdist/amici/de_model.py
+++ b/python/sdist/amici/de_model.py
@@ -143,7 +143,7 @@ class DEModel:
         model
 
     :ivar _sparsesyms:
-        carries linear list of all symbolS for sparsified
+        carries linear list of all symbols for sparsified
         variables
 
     :ivar _colptrs:


python/sdist/amici/de_model.py
```diff
@@ -143,7 +143,7 @@ class DEModel:
         model
 
     :ivar _sparsesyms:
-        carries linear list of all symbolic identifiers for sparsified
+        carries linear list of all symbolS for sparsified
```
FFroehlich:
```suggestion
        carries linear list of all symbols for sparsified
```

FFroehlich (approved):

PR #2913, opened 2025-07-18
Handle state-independent, parameter-dependent events without root-finding

### LINKED ISSUE #2185: Handle events with time-based triggers without rootfinding?

dweindl:
Since #2227, events that trigger at an explicit timepoint are handled without root-finding. Events with state-independent but parameter-dependent triggers were still handled with root-finding. With these changes, those event triggers will also be handled without root-finding.

Closes #2185.


dweindl:
This breaks/fixes the [root-after-reinitialization error example](https://amici.readthedocs.io/en/latest/examples/example_errors.html#Cvode-routine-CVode-returned-a-root-after-reinitialization) :cry: .

Any ideas for other small benchmark collection models where this error can easily be triggered?

PR #2921, opened 2025-07-22
No event trigger reinitialization after pre-equilibration and pre-simulation

### LINKED ISSUE #2918: Document/fix reinitialization of Heavisides after pre-equilibration and pre-simulation
Currently, event trigger state is reinitialized after pre-equilibration and pre-simulation.
Applying the initialValue of an SBML trigger is part of model initialization, and therefore, I think, it would be more plausible to do that only once, and continue with the Heaviside state from the previous period.
Either way, the behavior should be documented.

Related to https://github.com/PEtab-dev/PEtab/issues/642

After #2913, #2920.

dweindl:
Meanwhile, events are handled during pre-equilibration and pre-simulation. So far, SBML's `event.trigger.initialValue` was applied at the beginning of each period. However, this should only be applied at the beginning of the very first simulation period. This is fixed and tested here.

Closes #2918.

### DIFF
--- a/python/tests/test_preequilibration.py
+++ b/python/tests/test_preequilibration.py
@@ -761,8 +761,8 @@ def test_preequilibration_events(tempdir):
         bolus2 = 1
         bolus3 = 1
         bolus4 = 1
-        # E1 & E2 will both trigger during pre-equilibration and main
-        #  simulation (Heaviside is reset after pre-equilibration)
+        # E1 & E2 will only trigger during pre-equilibration
+        # (initialValue is only used once at the start of pre-equilibration)
         E1: at some_time >= 0, t0 = false: target1 = target1 + bolus1
         E2: at time >= 0, t0 = false: target2 = target2 + bolus2
         # requires early time point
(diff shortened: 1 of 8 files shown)


python/tests/test_preequilibration.py
```diff
@@ -761,8 +761,8 @@ def test_preequilibration_events(tempdir):
         bolus2 = 1
         bolus3 = 1
         bolus4 = 1
-        # E1 & E2 will both trigger during pre-equilibration and main
-        #  simulation (Heaviside is reset after pre-equilibration)
+        # E1 & E2 will only trigger during pre-equilibration
+        # (Heaviside is not reset after pre-equilibration)
```
FFroehlich: where is the Heaviside function here? I do think we should reset Heaviside functions, i.e. not reapply trigger initialisation but do reinitialise trigger function states.

dweindl: Ah, yeah, no Heaviside function there. I meant `h`.
I don't think I understand your suggestion. You suggest replacing reinitializing `h` by what?

FFroehlich (approved):

python/tests/test_preequilibration.py
FFroehlich: I think it would be good to test that we are properly reinitialising Heaviside variables, if we aren't already doing so.

dweindl: I think that is covered in the tests here.

PR #2927, opened 2025-07-28
Fix check for sign change, fix Heaviside update

### LINKED ISSUE #2926: Incorrect handling of Heavisides if simultaneous event changes trigger state
For the following model, amici produces incorrect results. The event should change the state of the piecewise-derived Heaviside function, but does not.

```
from amici import import_model_module
from amici.antimony_import import antimony2amici

tempdir = "deleteme"
model_name = tempdir
antimony2amici(
    r"""
    some_time = 0
    some_time' = 1

    threshold = 0
    
    x1 := piecewise(2, (some_time >= threshold), 1)
    at some_time >= 0, t0 = false: threshold = 10
    """,
    model_name=model_name,
    output_dir=tempdir,
)

model_module = import_model_module(model_name, tempdir)
model.setTimepoints([0, 5, 10, 11])
solver = model.getSolver()

rdata = amici.runAmiciSimulation(model, solver)
assert rdata.status == amici.AMICI_SUCCESS
assert rdata.by_id("x1").tolist() == [1.0, 1.0, 2.0, 2.0]
``` 

`E       assert [2.0, 2.0, 2.0, 3.0] == [1.0, 1.0, 2.0, 2.0]`

dweindl:
The old check for sign changes did not work if one values was zero.

Heaviside state has to be updated independent of whether the event occurs at the initial timepoint or not.

Fixes #2926.

### DIFF
--- a/src/forwardproblem.cpp
+++ b/src/forwardproblem.cpp
@@ -495,9 +495,7 @@ void EventHandlingSimulator::handle_events(
 
             store_pre_event_info(true);
 
-            if (!initial_event) {
-                model_->updateHeaviside(ws_->roots_found);
-            }
+            model_->updateHeaviside(ws_->roots_found);
         }
     }
     store_post_event_info();
@@ -616,8 +614,8 @@ int EventHandlingSimulator::detect_secondary_events() {
     for (int ie = 0; ie < model_->ne; ie++) {
         // the same event should not trigger itself
         if (ws_->roots_found.at(ie) == 0) {
-            // check whether there was a zero-crossing
-            if (0 > ws_->rval_tmp.at(ie) * ws_->rootvals.at(ie)) {
+            // check whether the value of the Heaviside function changed
+            if (heaviside_differs(ws_->rval_tmp.at(ie), ws_->rootvals.at(ie))) {
                 if (ws_->rval_tmp.at(ie) < ws_->rootvals.at(ie)) {
                     ws_->roots_found.at(ie) = 1;
                     auto const& event = model_->get_event(ie);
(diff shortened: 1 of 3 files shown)


src/forwardproblem.cpp
```diff
@@ -616,8 +614,8 @@ int EventHandlingSimulator::detect_secondary_events() {
     for (int ie = 0; ie < model_->ne; ie++) {
         // the same event should not trigger itself
         if (ws_->roots_found.at(ie) == 0) {
-            // check whether there was a zero-crossing
-            if (0 > ws_->rval_tmp.at(ie) * ws_->rootvals.at(ie)) {
```
FFroehlich, line 620: is this about one of the two being zero? I think you achieve the same by checking `0 >= ws_->rval_tmp.at(ie) * ws_->rootvals.at(ie)` instead of `0 > ws_->rval_tmp.at(ie) * ws_->rootvals.at(ie)`. Current implementation treats `0` as positive, why is that the right thing to do?

dweindl: Yes, it can be done in there. I find a separate function clarifies the intent, though.

In my understanding, `h_i = Heaviside(root[i]) = root[i] >= 0 ? 1 : 0`, so 0 and positive is the same in this context.

For the general function, one can argue about that, I guess. Mathematically, `sgn(0) != sgn(1)`, but in IEEE 754, there is +0 and -0. Agreed that at least a comment should be added there. Maybe I'll just rename it to `heavisides_differ`.

FFroehlich (approved):

PR #2939, opened 2025-08-18
Fix incorrect initialization for non-constant initial assignments

### LINKED ISSUE #2936: Constant parameter with initial assignment is compiled as a dynamic expression

The SBML model below has a constant parameter (k_synthesis), whose initial value is assigned using the initial value of some dynamic expression (v_shedding). I expect the constant parameter to stay constant during the amici simulation. Simulating with COPASI produces the expected behaviour.

In the compiled amici model, the parameter is listed in the `w.cpp` dynamic expressions and changes its value during simulation.

**To Reproduce**
The SBML is written out below, and the amici model was compiled with this Python fragment:
```python
import amici

model_name = "model_minimal"
sbml_file = f"{model_name}.xml"
sbml_importer = amici.SbmlImporter(sbml_file)
model_output_dir = f"amici_{model_name}"
sbml_importer.sbml2amici(model_name=model_name, output_dir=model_output_dir)
```

 - OS and version: Ubuntu 22.04.5 LTS
 - AMICI interface: Python
 - AMICI version: 0.34.0
 - Additional information: Compiler name and version used: `gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0`, Python version 3.12.2
 
**model_minimal.xml**
```
<!-- Created by libAntimony version v2.15.0 with libSBML version 5.20.2. -->
  <model metaid="debug" id="debug" substanceUnits="substance" timeUnits="time_unit" volumeUnits="volume" areaUnits="area" lengthUnits="length">
    <listOfUnitDefinitions>
      <unitDefinition id="length">
        <listOfUnits>
          <unit kind="metre" exponent="1" scale="0" multiplier="1"/>
        </listOfUnits>
      </unitDefinition>
      <unitDefinition id="area">
        <listOfUnits>
          <unit kind="metre" exponent="2" scale="0" multiplier="1"/>
        </listOfUnits>
      </unitDefinition>
      <unitDefinition id="volume">
        <listOfUnits>
          <unit kind="litre" exponent="1" scale="0" multiplier="1"/>
        </listOfUnits>
      </unitDefinition>
      <unitDefinition id="time_unit" name="minutes">
        <listOfUnits>
          <unit kind="second" exponent="1" scale="1" multiplier="6"/>
        </listOfUnits>
      </unitDefinition>
      <unitDefinition id="substance">
        <listOfUnits>
          <unit kind="mole" exponent="1" scale="-9" multiplier="1"/>
        </listOfUnits>
      </unitDefinition>
    </listOfUnitDefinitions>
    <listOfCompartments>
      <compartment id="well" spatialDimensions="3" size="0.0005" constant="true"/>
    </listOfCompartments>
    <listOfSpecies>
      <species id="MR" compartment="well" hasOnlySubstanceUnits="false" boundaryCondition="false" constant="false"/>
    </listOfSpecies>
    <listOfParameters>
      <parameter id="k_synthesis" constant="true"/>
      <parameter id="v_shedding" constant="false"/>
      <parameter id="MR_t0" value="0.001" constant="true"/>
      <parameter id="k_b" value="0.001" constant="true"/>
    </listOfParameters>
    <listOfInitialAssignments>
      <initialAssignment symbol="MR">
          <ci> MR_t0 </ci>
        </math>
      </initialAssignment>
      <initialAssignment symbol="k_synthesis">
          <ci> v_shedding </ci>
        </math>
      </initialAssignment>
    </listOfInitialAssignments>
    <listOfRules>
      <assignmentRule variable="v_shedding">
          <apply>
            <times/>
            <ci> MR </ci>
            <ci> k_b </ci>
            <apply>
              <plus/>
              <cn type="integer"> 1 </cn>
              <csymbol encoding="text" definitionURL="http://www.sbml.org/sbml/symbols/time"> time </csymbol>
            </apply>
          </apply>
        </math>
      </assignmentRule>
    </listOfRules>
    <listOfReactions>
      <reaction id="_J0" reversible="true">
        <listOfProducts>
          <speciesReference species="MR" stoichiometry="1" constant="true"/>
        </listOfProducts>
        <kineticLaw>
            <ci> k_synthesis </ci>
          </math>
        </kineticLaw>
      </reaction>
      <reaction id="_J1" reversible="true">
        <listOfReactants>
          <speciesReference species="MR" stoichiometry="1" constant="true"/>
        </listOfReactants>
        <kineticLaw>
            <ci> v_shedding </ci>
          </math>
        </kineticLaw>
      </reaction>
    </listOfReactions>
  </model>
</sbml>
```

**How to fix**
Do you know how to resolve the problem? No
Can you submit a pull request? Happy to try after understanding the issue.

dweindl:
Fixes a bug where targets of initial assignments are not initialized correctly if the initial assignment expression is implicitly time-dependent. Fixes #2936.

* eliminate state variables from x0 after replacing rate-of expressions
* implement supposed-to-be static expressions that have non-constant entities in their initial condition as state variables instead of expressions.

### DIFF
--- a/python/sdist/amici/de_model.py
+++ b/python/sdist/amici/de_model.py
@@ -454,23 +454,6 @@ class DEModel:
             )
             self._eqs["xdot"] = smart_subs_dict(self.eq("xdot"), subs)
 
-        # replace rateOf-instances in x0 by xdot equation
-        made_substitutions = False
-        for i_state in range(len(self.eq("x0"))):
-            new, replacement = self._eqs["x0"][i_state].replace(
-                rate_of_func, get_rate, map=True
-            )
-            if replacement:
-                self._eqs["x0"][i_state] = new
-                made_substitutions = True
-        if made_substitutions:
-            # replace any newly introduced state variables
-            #  by their x0 expressions
-            subs = toposort_symbols(
-                dict(zip(self.sym("x_rdata"), self.eq("x0"), strict=True))
-            )
-            self._eqs["x0"] = smart_subs_dict(self.eq("x0"), subs)
-
         # replace rateOf-instances in w by xdot equation
         #  here we may need toposort, as xdot may depend on w
         made_substitutions = False
@@ -517,6 +500,26 @@ class DEModel:
             self._expressions = [self._expressions[i] for i in new_order]
             self._syms["w"] = sp.Matrix(topo_expr_syms)
             self._eqs["w"] = sp.Matrix(list(w_sorted.values()))
+
+        # replace rateOf-instances in x0 by xdot equation
+        made_substitutions = False
+        for i_state in range(len(self.eq("x0"))):
+            new, replacement = self._eqs["x0"][i_state].replace(
+                rate_of_func, get_rate, map=True
+            )
+            if replacement:
+                self._eqs["x0"][i_state] = new
+                made_substitutions = True
+        if made_substitutions:
+            # Replace any newly introduced state variables
+            #  by their x0 expressions.
+            # Also replace any newly introduced `w` symbols by their
+            #  expressions (after `w` was toposorted above).
+            subs = toposort_symbols(
+                dict(zip(self.sym("x_rdata"), self.eq("x0"), strict=True))
+            )
+            subs = dict(zip(self._syms["w"], self.eq("w"), strict=True)) | subs
+            self._eqs["x0"] = smart_subs_dict(self.eq("x0"), subs)
 
         for component in chain(
             self.observables(),
--- a/python/sdist/amici/de_model.py
+++ b/python/sdist/amici/de_model.py
@@ -502,15 +502,16 @@ class DEModel:
             self._eqs["w"] = sp.Matrix(list(w_sorted.values()))
 
         # replace rateOf-instances in x0 by xdot equation
-        made_substitutions = False
+        # indices of state variables whose x0 was modified
+        changed_indices = []
         for i_state in range(len(self.eq("x0"))):
             new, replacement = self._eqs["x0"][i_state].replace(
                 rate_of_func, get_rate, map=True
             )
             if replacement:
                 self._eqs["x0"][i_state] = new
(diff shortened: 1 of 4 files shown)


python/sdist/amici/de_model.py
```diff
@@ -455,12 +455,21 @@ def get_rate(symbol: sp.Symbol):
...
         # replace rateOf-instances in x0 by xdot equation
+        made_substitutions = False
         for i_state in range(len(self.eq("x0"))):
             new, replacement = self._eqs["x0"][i_state].replace(
                 rate_of_func, get_rate, map=True
             )
             if replacement:
                 self._eqs["x0"][i_state] = new
+                made_substitutions = True
+        if made_substitutions:
```
FFroehlich: Only apply this to `new` above?

python/sdist/amici/de_model.py
```diff
@@ -509,6 +501,26 @@ def get_rate(symbol: sp.Symbol):
...
+            # Replace any newly introduced state variables
+            #  by their x0 expressions.
+            # Also replace any newly introduced `w` symbols by their
+            #  expressions (after `w` was toposorted above).
+            subs = toposort_symbols(
+                dict(zip(self.sym("x_rdata"), self.eq("x0"), strict=True))
+            )
+            subs = dict(zip(self._syms["w"], self.eq("w"), strict=True)) | subs
```
FFroehlich, line 522: I think this has to be done in a loop as `dict(zip(self._syms["w"], self.eq("w"), strict=True))` subs may introduce new `self.sym("x_rdata")` variables.

dweindl: `w` should not depend on `x_rdata`:

https://github.com/AMICI-dev/AMICI/blob/e38afde6423847698380158778464aa423963347/python/sdist/amici/_codegen/cxx_functions.py#L313-L319

Unless the respective substitutions in `w` are only made later, but I don't think that is the case -  the rateOf processing should happen last.

PR #2943, opened 2025-08-28
Fix handling of initial assignments containing splines

dweindl:
Currently, `x0` must not depend on splines. Therefore, substitute the spline symbol by its piecewise expression when compiling x0 expressions.

Reported by @m-philipps.


m-philipps:
Thanks a lot!

Issue #2849, opened 2025-06-27
beartype type checking in jax broken


FFroehlich:
beartype type checking seems to only work with unjitted code, so currently checks don't do anything

PR #2947, opened 2025-09-02
PEtab SciML test suite - test_net and (some) test_ude cases

BSnelling:
Implements support for petab sciml test suite: https://github.com/sebapersson/petab_sciml_testsuite/tree/main

Includes support for all `test_net` cases and a subset of `test_ude`. Test cases with frozen layers and networks in the observable formulae are not yet implemented.

### DIFF
--- a/python/sdist/amici/jax/petab.py
+++ b/python/sdist/amici/jax/petab.py
@@ -1104,17 +1104,21 @@ class JAXProblem(eqx.Module):
         p_array = jnp.stack(
             [self.load_model_parameters(sc) for sc in conditions]
         )
-        unscaled_parameters = jnp.stack(
-            [
-                jax_unscale(
-                    self.parameters[ip],
-                    self._petab_problem.parameter_df.loc[
-                        p_id, petab.PARAMETER_SCALE
-                    ],
-                )
-                for ip, p_id in enumerate(self.parameter_ids)
-            ]
-        )
+
+        if self.parameters.size:
+            unscaled_parameters = jnp.stack(
+                [
+                    jax_unscale(
+                        self.parameters[ip],
+                        self._petab_problem.parameter_df.loc[
+                            p_id, petab.PARAMETER_SCALE
+                        ],
+                    )
+                    for ip, p_id in enumerate(self.parameter_ids)
+                ]
+            )
+        else:
+            unscaled_parameters = jnp.zeros((*self._ts_masks.shape[:2], 0))
 
         if op_numeric is not None and op_numeric.size:
             op_array = jnp.where(
--- a/python/sdist/amici/jax/petab.py
+++ b/python/sdist/amici/jax/petab.py
@@ -97,8 +97,6 @@ def _get_hybridization_df(petab_problem):
         ]
         hybridization_df = pd.concat(hybridizations)
         return hybridization_df
-    else:
-        return None
 
 
 def _get_hybrid_petab_problem(petab_problem: petab.Problem):
--- a/python/sdist/amici/jax/petab.py
+++ b/python/sdist/amici/jax/petab.py
@@ -16,6 +16,7 @@ import equinox as eqx
 import jaxtyping as jt
 import jax.lax
 import jax.numpy as jnp
+import jax.tree_util as jtu
 import numpy as np
 import pandas as pd
 import petab.v1 as petab
@@ -1414,6 +1415,7 @@ def run_simulations(
         ..., diffrax._custom_types.BoolScalarLike
     ] = diffrax.steady_state_event(),
     max_steps: int = 2**10,
+    is_grad_mode: bool = False,
     ret: ReturnValue | str = ReturnValue.llh,
 ):
     """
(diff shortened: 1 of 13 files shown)


python/sdist/amici/jax/petab.py
```diff
@@ -524,28 +527,51 @@ def _get_nominal_parameter_values(
...
+            if (net := pname.split("_")[0]) in model.nns:
                 to_set = []
                 nn = model_pars[net]
-                try:
-                    value = float(row[petab.NOMINAL_VALUE])
-                except ValueError:
-                    value = par_arrays[row[petab.NOMINAL_VALUE]]
+                scalar = True
+
+                # ?? When value is NaN do we want to go to par_arrays ??
```
FFroehlich: Modern IDEs should support [Codetags](https://peps.python.org/pep-0350/) which can make it easier to track these comments

python/sdist/amici/jax/petab.py
```diff
@@ -588,6 +614,38 @@ def _get_nominal_parameter_values(
...
+                                    [
+                                        (
+                                            k,
+                                            jnp.array(
+                                                nn_input_arrays[net_id][input][
+                                                    k
+                                                ][:],
+                                                dtype=jnp.float64,
+                                            ),
+                                        )  # ?? hardcoded dtype not ideal ?? could infer from env somehow ??
```
FFroehlich: is it really necessary to set `dtype` here? Usually jax infers float precision from https://docs.jax.dev/en/latest/config_options.html. Might be necessary to cast this as numpy array first if conversion from hdf5 is the problem

python/sdist/amici/jax/petab.py
```diff
@@ -709,15 +780,70 @@ def _eval_nn(self, output_par: str):
...
+
+        hybridization_parameter_map = dict(
+            [
+                (petab_id, self._hybridization_df.loc[petab_id, "targetValue"])
+                for petab_id in model_id_map.values()
+                if petab_id in set(self._hybridization_df.index)
+            ]
+        )
+
+        # ?? conditional nightmare ?? refactor ??
```
FFroehlich: 👍

python/sdist/amici/jax/petab.py
```diff
@@ -709,15 +780,70 @@ def _eval_nn(self, output_par: str):
             .to_dict()
         )
 
+        condition_parameter_map = (
```
FFroehlich: `condition_input_map`?

python/sdist/amici/jax/petab.py
```diff
@@ -709,15 +780,70 @@ def _eval_nn(self, output_par: str):
...
+                    (
+                        petab_id,
+                        self._petab_problem.parameter_df.loc[
+                            self._petab_problem.condition_df.loc[
+                                condition_id, petab_id
+                            ],
+                            petab.NOMINAL_VALUE,
+                        ],
+                    )
+                    if "input"
```
FFroehlich: this check seems a bit too unspecific. I think you want to construct a sequence of petab id's that are mapped to `$nnId.inputs{[$inputArgumentIndex]{[$inputIndex]}}`?

python/sdist/amici/jax/petab.py
```diff
@@ -994,6 +1134,8 @@ def run_simulation(
         nps: jt.Float[jt.Array, "nt *nnp"],  # noqa: F821, F722
         mask_reinit: jt.Bool[jt.Array, "nx"],  # noqa: F821, F722
         x_reinit: jt.Float[jt.Array, "nx"],  # noqa: F821, F722
+        init_override: jt.Float[jt.Array, "np"], # ?? what do these annotations mean ??
```
FFroehlich: type hints via https://docs.kidger.site/jaxtyping/, supposed to be checked via https://github.com/beartype/beartype but currently not functioning, see https://github.com/AMICI-dev/AMICI/issues/2849

I think the dimension should be `nx` not `np` here.

tests/sciml/test_sciml.py
```diff
@@ -202,8 +194,16 @@ def test_ude(test):
 
     # llh
     if test in (
+        # ?? cases where nn part of observable formula ??
```
FFroehlich: I think this might have been cases with neural network architecture we do not support

python/sdist/amici/jax/petab.py
```diff
@@ -1590,3 +1615,52 @@ def petab_simulate(
             )
         dfs.append(df_sc)
     return pd.concat(dfs).sort_index()
+
+def apply_grad_filter(problem: JAXProblem,):
```
FFroehlich: This is a great solution! Only thing that I am a bit worried about is that in the current implementation `stop_gradient` is only applied when calling `problem` methods in the context of `run_simulations`, which may lead to confusion when trying to compute gradients outside of that context. My interpretation of the petab problem definition is that setting `estimate=0` means that gradient computation is permanently disabled and we should apply `apply_grad_filter` during `JaxProblem` instantiation.

FFroehlich:
just checking test failures:

- Notebook tests, also fails on the base branch albeit for a different reasons, but this is not related to changes here
- mac os, this looks like an issue with PRs from a fork
- doc tests: also problem in base branch, h5py is missing from doc requirements
- sbml jax: unrelated, also failing in base branch

python/sdist/amici/jax/petab.py
BSnelling: I tried casting to a numpy array but the array persisted as float32s. I've defined the `dtype` based on the current jax config settings which I think is better than hard coding.

python/sdist/amici/jax/petab.py
BSnelling: I've updated this.  The complication is that the values here could be pulled from a nominal value in the parameter table or from a value in the conditions table, depending on whether the id appears in the parameters table. That's my understanding anyway.

tests/sciml/test_sciml.py
BSnelling: Fair enough, they've all been implemented now in any case!

python/sdist/amici/jax/petab.py
```diff
@@ -588,6 +647,40 @@ def _get_nominal_parameter_values(
                         model,
                         model_pars[net_id][layer_id][attribute],
                     )
+
+        # set inputs in the model if provided
+        if len(nn_input_arrays) > 0:
+            for net_id in model_pars:
+                for input in model.nns[net_id].inputs:
+                    input_array = dict(
```
FFroehlich: bit unwieldy could we simplify this?

python/sdist/amici/jax/petab.py
```diff
@@ -1118,6 +1302,38 @@ def run_simulations(
...
+        )
+        init_override = jnp.stack(
+            [
+                jnp.array(
+                    [
+                        self._eval_nn(
+                            self._parameter_mappings[sc].map_sim_var[p], sc
+                        )
+                        if p
+                        in set(self._parameter_mappings[sc].map_sim_var.keys())
```
FFroehlich, line 1314: This isn't quite the check I would expect here (rather something with `condition_df`/`mapping_df`), but fine to keep as is for now as we will anyways have to change the whole parameter mapping setup for v2.

FFroehlich (approved):

Issue #2944, opened 2025-08-28
Replace amici.runAmiciSimulation


dweindl:
In the next major release, we might want to replace `amici.runAmiciSimulation` by something shorter. Just `amici.simulate` or `Model.simulate`. Open for other suggestions.

dweindl:
Maybe this should be planned in combination with #2845 to come up with a more convenient API, in particular for interactive use. See also the discussion in https://github.com/AMICI-dev/AMICI/issues/931.

dweindl:
For a start, because `runAmiciSimulation` and `runAmiciSimulations` will anyways be renamed as part of #887, I'd rename them to `simulate` and `simulate_multiple` (or `simulate_condition` and `simulate_conditions`? or `simulate` and `simulate_conditions`? other suggestions?).

I'd add them to `Model` in Python, because that's more convenient for interactive use, but keep them as free function in C++, for now.

dweindl:
Closed by [#2963](https://github.com/AMICI-dev/AMICI/pull/2963)

PR #2963, opened 2025-10-06
Add `Model.simulate`

dweindl:
~~Replace~~ Complement the old `amici.runAmiciSimulation`, `amici.runAmiciSimulations` ~~by~~ with something more convenient.

Related to #2944.


python/tests/test_sbml_import.py
```diff
@@ -115,10 +115,7 @@ def test_nosensi(tempdir):
 
     model = model_module.getModel()
     model.setTimepoints(np.linspace(0, 60, 61))
-    solver = model.getSolver()
-    solver.setSensitivityOrder(amici.SensitivityOrder.first)
-    solver.setSensitivityMethod(amici.SensitivityMethod.forward)
-    rdata = amici.runAmiciSimulation(model, solver)
+    rdata = model.simulate(sensi_order="first", sensi_method="forward")
```
dilpath, line 121:
```suggestion
    rdata = model.simulate(sensi_order=amici.SensitivityOrder.first, sensi_method=amici.SensitivityMethod.forward)
```

dweindl: Both are supported. I like the former, because it's less verbose, in particular for interactive use.

dweindl:
> Works for me 👍 I would also be fine with removing `Model.simulate_multiple` and instead adapting `Model.simulate` to support `edatas = ExpData | list[ExpData]`.

Thanks. Yeah, I am divided on that one.

dweindl:
> > Works for me 👍 I would also be fine with removing `Model.simulate_multiple` and instead adapting `Model.simulate` to support `edatas = ExpData | list[ExpData]`.
> 
> Thanks. Yeah, I am divided on that one.

I think my concerns are easily addressable by two `typing.overload`s. So, in favor of merging them.

Issue #2727, opened 2025-05-15
Drop MATLAB interface


dweindl:
Following up on #2475, let's get rid of the MATLAB interface which was dragged along, but not really well maintained over the last years.

* [x] port any matlab examples that are still valuable for testing to Python and add them to them CI workflows (#2898, #2822, https://github.com/AMICI-dev/AMICI/issues/854)
* [x] remove matlab code for model import (#2932)
* [x] remove matlab examples (#2932)
* [x] remove pre-generated c++ code for matlab test models (#2931)
* [x] remove matlab interface from the documentation (#2932) 
* [x] ... as well as any other mentions of the matlab interface in the docs (#2961)
* [x] remove all mtoc / matlab-doxygen related stuff (#2932)
* [x] remove matlab GHA workflows (#2931)
* [x] remove matlab c++ interface code (#2932)
* [x] remove all matlab- vs python-code generation switches from the library (#2957)
* [x] remove Python tests for matlab-pregenerated models (#2931)
* [x] remove old C++ spline code and obsolete parts from `symbolic_functions.{h,cpp}` (started in #2932, #2956)
* [x] Update https://raw.githubusercontent.com/AMICI-dev/AMICI/main/doc/gfx/amici_workflow.png (#2962)

Prerequisites for porting all old tests:
* [x] #2102 
* [x] #1539 
~~* [ ] #2905~~ (out of scope)

dweindl:
> * port any matlab examples that are still valuable for testing to Python and add them to them CI workflows 

Note that not all of the current Matlab/C++ tests are testing what they are supposed to test. 
E.g. `TEST(ExampleSteadystate, SensiFwdNewtonPreeq)` does not use Newton's method. Probably because the default steady state method was changes to integration-only, but the test relied on the default. And the test code never checked whether Newton's method was used successfully (or at all).

PR #2964, opened 2025-10-06
refactor: 0-based indexing in `Model::z2event_`

dweindl:
Get rid of 1-based event-indexing in `Model::z2event_` from ancient MATLAB times.

Related to #2727.


FFroehlich (approved):
🥳

PR #2974, opened 2025-10-13
PEtab v2 import

### LINKED ISSUE #2662: Implement PEtab v2 import
We need to revise the PEtab import to handle the upcoming PEtab v2 format.

This implies:
* New parameter mapping
  * instead of "preequilibration" and simulation, there will be a variable number of periods (at first, more than 2 conditions will only be supported through the JAX interface)
  * the mapped values will not be identifier or float, but could also be more complex math expressions
  * the values for reinitialization may depend on the model state and parameters (this also needs to be accounted for when computing sensitivities)
* Auto-upconversion of PEtab v1 problems, to avoid duplicating the import logic  
* ...

dweindl:
Experimental import of SBML- and PySB-based PEtab v2 problems.
Not particularly performant or feature-complete yet - to be extended.

Closes #2662.

:eyes: https://amici--2974.org.readthedocs.build/en/2974/examples/example_petab/petab_v2.html
:eyes: https://amici--2974.org.readthedocs.build/en/2974/generated/amici.petab.petab_importer.html#module-amici.petab.petab_importer

### DIFF
--- a/python/sdist/amici/petab/petab_importer.py
+++ b/python/sdist/amici/petab/petab_importer.py
@@ -65,7 +65,7 @@ __all__ = [
     "RES",
     "SRES",
 ]
-logger = get_logger(__name__, log_level=logging.DEBUG)
+logger = get_logger(__name__, log_level=logging.INFO)
 
 
 #: Default experiment ID to be used for measurements without an experiment ID.
--- a/tests/petab_test_suite/test_petab_v2_suite.py
+++ b/tests/petab_test_suite/test_petab_v2_suite.py
@@ -56,9 +56,7 @@ def _test_case(case, model_type, version, jax):
 
     # compile amici model
     if case.startswith("0006") and not jax:
-        # TODO: petab.flatten_timepoint_specific_output_overrides(problem)
-        # petab.flatten_timepoint_specific_output_overrides(problem)
-        pytest.skip("Timepoint-specific output overrides not yet supported")
+        flatten_timepoint_specific_output_overrides(problem)
 
     model_name = (
         f"petab_{model_type}_test_case_{case}_{version.replace('.', '_')}"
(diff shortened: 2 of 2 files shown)


tests/petab_test_suite/test_petab_v2_suite.py
```diff
@@ -0,0 +1,213 @@
...
+
+    # load
+    case_dir = petabtests.get_case_dir(case, model_type, version)
+    yaml_file = case_dir / petabtests.problem_yaml_name(case)
+    problem = v2.Problem.from_yaml(yaml_file)
+
+    # compile amici model
+    if case.startswith("0006") and not jax:
+        # TODO: petab.flatten_timepoint_specific_output_overrides(problem)
+        # petab.flatten_timepoint_specific_output_overrides(problem)
```
FFroehlich: Other tests suggest this is now supported

python/sdist/amici/petab/petab_importer.py
```diff
@@ -0,0 +1,2282 @@
...
+    ) -> v2.Problem:
+        """Upgrade the problem to petab v2 if necessary.
+        Otherwise, create a deep copy of the problem."""
+        if isinstance(problem, v2.Problem):
+            return copy.deepcopy(problem)
+
+        # TODO: So far, PEtab can only upgrade file-based problems,
+        #  not petab.v1.Problem objects.
+        raise NotImplementedError(
+            "Only `petab.v2.Problem` is currently supported. "
```
FFroehlich: should this error message explicitly mention incompatibility with file based problems to avoid confusion?

python/sdist/amici/petab/petab_importer.py
```diff
@@ -0,0 +1,2282 @@
...
+                        (default, True),
+                    )
+
+                    # Update the target expression
+                    if target_entity is not None and isinstance(
+                        target_entity, pysb.Expression
+                    ):
+                        target_entity.value = expr
+                    else:
+                        # if the target is not an expression, it must be an
```
FFroehlich, line 2107: TODO: would be more efficient to convert this into a multi-model problem, but who knows whether we will ever get around to implementing that.

tests/petab_test_suite/test_petab_v2_suite.py
dweindl: Now implemented.

python/sdist/amici/petab/petab_importer.py
dweindl: Reworded.

python/sdist/amici/petab/petab_importer.py
dweindl: Added as TODO.

Issue #1126, opened 2020-06-10
Serialization of amici::Model and amici::Solver and potentially others from Python


dweindl:
Following up on https://github.com/ICB-DCM/pyPESTO/issues/391, it would be great to have an easy way to serialize AMICI objects from python, as pickling SWIG objects doesn't work out of the box.

Not sure what is the best way. Some options:

* Going via HDF5 files
  Pro: functionality is in place
  Con: requires building AMICI with HDF5 support. additional C++ dependency, probably troublesome for some users.

* Using boost serialization
  Pro: functionality is in place
  Con: requires building AMICI with boost. additional C++ dependency, not at all implemented/tested from python probably troublesome for some users.

* Implementing pickling functions in Python
  Pro: easy to use, no additional dependency
  Con:  Redundancy / maintainability: Needs to be updated on every change of the respective classes, which turned out to not work too well in the existing serialization code.

* Implementing {to,from}{json,yaml,...} in C++:
  Pro: nice to have, usable from C++ and python
  Con: Introduces additional dependencies or custom re-implementation of parsers (strongly opposed). Redundancy / maintainability:  Needs to be updated on every change of the respective classes, which turned out to not work too well in the existing serialization code.

Any ideas how to handle that well?

paulstapor:
So, if I get it correctly: Last option is most cumbersome to implement (or introduces some new dependency), but potentially most helpful.

With hdf5 requirements, there may be problems for people who use anaconda,  as anaconda does weird stuff with its own hdf5-libraries. 
So, I think going via Boost introduces headaches to less people than going via hdf5, right?

FFroehlich:
Regarding directly implementing pickling in python, I think its feasible to implement this by automatically identifying getter/setter methods by text parsing and store the respective values. This approach would ensure maintainability as long as standard naming schemes are adhered to (which could be enforced by only excluding checks for a whitelisted set of methods).

This is already more or less implemented as https://github.com/ICB-DCM/AMICI/blob/ffb5f33001047556a718c504ad17bc81ed3e206d/python/tests/test_swig_interface.py#L17

dweindl:
> implement this by automatically identifying getter/setter methods by text parsing and store the respective values

Yeah, might be the most reasonable short-term solution.

dweindl:
We might want to look at https://github.com/USCiLab/cereal (header-only C++ library) as an alternative to `boost::serialization`. This could be included here to avoid any extra dependencies that have to be satisfied by the user.

PR #2985, opened 2025-10-22
Enable pickling of `ModelPtr`

dweindl:
Enable pickling of `ModelPtr` for use in multiprocessing contexts.

Related to #1126.

### DIFF
--- a/python/tests/test_swig_interface.py
+++ b/python/tests/test_swig_interface.py
@@ -5,12 +5,15 @@ Test getters, setters, etc.
 
 import copy
 import numbers
+import pickle
 from math import nan
 
 import amici
 import numpy as np
 import pytest
 import xarray
+from amici import SteadyStateSensitivityMode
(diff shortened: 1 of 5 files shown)


python/tests/test_swig_interface.py
```diff
@@ -664,3 +667,33 @@ def test_reporting_mode_obs_llh(sbml_example_presimulation_module):
...
                 assert rdata.sllh.size > 0
                 assert not np.isnan(rdata.sllh).any()
+
+
+@skip_on_valgrind
+def test_pickle_model(sbml_example_presimulation_module):
+    model_module = sbml_example_presimulation_module
+    model = model_module.get_model()
+
+    assert (
```
FFroehlich, line 677: we are already testing other settings programmatically in corresponding tests for get_model_settings, right?

dweindl: Correct.

PR #3001, opened 2025-11-01
Add `ModelQuantity.get_sym`, make `ModelQuantity.get_id` return `str`

### LINKED ISSUE #2940: Add ModelQuantity.sym, make ModelQuantity.get_id return str
The current behavior (`def get_id(self) -> sp.Symbol`) is confusing.

dweindl:
Distinguish between `id: str` and `symbol: sp.Symbol`. The old behavior (`def get_id(self) -> sp.Symbol`) was confusing.

Fix inconsistent type for `AlgebraicEquation` symbol/id.

Closes #2940.

### DIFF
--- a/python/sdist/amici/de_model_components.py
+++ b/python/sdist/amici/de_model_components.py
@@ -47,15 +47,15 @@ class ModelQuantity:
 
     def __init__(
         self,
-        identifier: sp.Symbol,
+        symbol: sp.Symbol,
         name: str,
         value: SupportsFloat | numbers.Number | sp.Expr,
     ):
         """
         Create a new ModelQuantity instance.
 
-        :param identifier:
-            unique identifier of the quantity
+        :param symbol:
+            Symbol of the quantity with unique identifier.
 
         :param name:
             individual name of the quantity (does not need to be unique)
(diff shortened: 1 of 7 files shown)


python/sdist/amici/de_model_components.py
```diff
@@ -91,16 +90,29 @@ def __repr__(self) -> str:
...
-    def get_id(self) -> sp.Symbol:
+    def get_id(self) -> str:
         """
         ModelQuantity identifier
 
         :return:
             identifier of the ModelQuantity
         """
-        return self._identifier
+        return (
```
FFroehlich, line 111: This looks more like a `get_name` than a `get_id` to me.

dweindl: The amici IDs correspond to sympy names (`Symbol.name`). The AMICI names can be different and don't have to be unique.  `get_name` is just below.

FFroehlich: Ah I see, that sounds like we can replace most if not all `strip_pysb(c.get_symbol)` and `str(c.get_sym())` with `c.get_id()` then and get rid of `strip_pysb`.

dweindl: That would be great. I was about to ask whether we can get rid of `strip_pysb` somehow. I'd keep that for a separate PR.

PR #3023, opened 2025-11-11
separate out importer code

FFroehlich:


python/sdist/amici/importers/__init__.py
```diff
@@ -0,0 +1,81 @@
+"""Unified model import interface for AMICI.
+
+This subpackage aggregates functionality for importing models from various
+formats into AMICI.
+
+Re-exported symbols (imported conditionally where dependencies are optional):
```
dweindl: This will pull in a large number of packages. I'd prefer not importing anything here by default. Usually, one only needs a single importer. I think `from amici.importers.sbml import *` is quite easy to memorize and not overly verbose.

python/sdist/amici/de_model.py
```diff
@@ -45,7 +45,7 @@
     var_in_function_signature,
 )
 from .exporters.sundials.cxxcodeprinter import csc_matrix
-from .import_utils import (
+from .importers.utils import (
```
dweindl, line 48: Okay for now, but I'd say we should make the core symbolic code not import anything from amici.importers, only vice verse. Can be addressed when moving de_model.py.

python/sdist/amici/exporters/sundials/cxxcodeprinter.py
```diff
@@ -12,7 +12,7 @@
 from sympy.utilities.iterables import numbered_symbols
 from toposort import toposort
 
-from ...import_utils import symbol_with_assumptions
+from amici.importers.utils import symbol_with_assumptions
```
dweindl, line 15: Similar to above, amici.exporters should not import anything from amici.importers, only from the symbolic middle layer.

dweindl (approved):
:+1:

PR #3036, opened 2025-11-14
Refactor rateOf handling

dweindl:
Pull rateOf-handling out of DEModel and keep it along other SBML processing where it belongs.

This is easier to follow and prevents some lingering issues with the old approach due to the xdot / w interdependencies.

This also handles rateOf expressions in some additional, previously unsupported places like event assignments.

I think this also fixes a potential bug in the `w` reordering in `_process_hybridization` that might otherwise trigger in combination with conservations laws, because they might end up at the beginning of `w`.

A minor downside is that in xdot rateOf expressions are replaced by the respective xdot equations where previously xdot symbols were used, but I prefer the reduced complexity. If this becomes a performance problem, this can easily be addressed by shifting rateOf into `w` by adding an extra assignment rule to the model.

### DIFF
--- a/python/sdist/amici/importers/sbml/__init__.py
+++ b/python/sdist/amici/importers/sbml/__init__.py
@@ -5,6 +5,7 @@ This module provides all necessary functionality to import a model specified
 in the `Systems Biology Markup Language (SBML) <https://sbml.org/>`_.
 """
 
+import contextlib
 import copy
 import itertools as itt
 import logging
@@ -14,6 +15,7 @@ import re
 import warnings
 import xml.etree.ElementTree as ET
 from collections.abc import Callable, Iterable, Sequence
+from itertools import chain
 from pathlib import Path
 from typing import (
     Any,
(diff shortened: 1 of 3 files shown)


python/sdist/amici/importers/sbml/__init__.py
```diff
@@ -3014,6 +3023,122 @@ def _transform_dxdt_to_concentration(
...
 
+    def _process_sbml_rate_of(self, de_model: DEModel) -> None:
+        """Substitute any SBML-rateOf constructs in the model equations"""
+        from sbmlmath import rate_of as rate_of_func
+
+        for ae in de_model.algebraic_equations():
+            if rate_ofs := ae.get_val().find(rate_of_func):
+                # Not implemented
+                raise SBMLException(
+                    "rateOf() is not supported in algebraic equations."
```
FFroehlich: include name of ae?, why is this necessary in given `is_ode` check below?

dweindl: > include name of ae?

:+1:

> why is this necessary in given `is_ode` check below?

Currently, no substitution is even attempted for any ae. Therefore, that check will never trigger. Will add a comment.

PR #3053, opened 2025-11-18
Replace min/max disguised as piecewise

dweindl:
During model import, simplify piecewise functions that can be expressed as min()/max().

So far, all piecewise constructs are treated as discontinuous (see also #2049). For min/max, only the derivative is discontinuous, which currently does not receive any special treatment in amici (see also https://amici.readthedocs.io/en/latest/implementation_discontinuities.html).

So this change prevents root tracking for some unnecessary piecewises. Addresses some aspects of https://github.com/AMICI-dev/AMICI/issues/2049.

### DIFF
--- a/python/sdist/amici/importers/sbml/__init__.py
+++ b/python/sdist/amici/importers/sbml/__init__.py
@@ -66,6 +66,7 @@ from amici.importers.utils import (
 from amici.logging import get_logger, log_execution_time, set_log_level
 from amici.sympy_utils import (
     _monkeypatch_sympy,
+    _piecewise_to_minmax,
     smart_is_zero_matrix,
     smart_multiply,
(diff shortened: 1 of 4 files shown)


python/sdist/amici/importers/sbml/__init__.py
```diff
@@ -2882,6 +2883,7 @@ def subs_locals(expr: sp.Basic) -> sp.Basic:
         # piecewise to heavisides
         if piecewise_to_heaviside:
             try:
+                expr = expr.replace(sp.Piecewise, _piecewise_to_minmax)
```
FFroehlich, line 2886: also add small test with actual model to make sure that we are not reverting this substitution someplace else?

PR #3077, opened 2025-11-25
Catch time overflow during steady-state simulation

### LINKED ISSUE #3076: Incorrect check for finiteness of x in fxdot
https://github.com/AMICI-dev/AMICI/blob/6a85422ad233fc268fb71272bba39f5b82db387a/src/solver_cvodes.cpp#L1181-L1188

This check will abort simulation at the specified time *unless* `x` is non-finite (see return values of `check_finite`). Broken from the [start](https://github.com/AMICI-dev/AMICI/commit/6f7b51fabc96d6c6bc058943b2f26f89f26b899f).

dweindl:
* Remove the old and faulty check for finiteness of `x` in `Model::fxdot`. Closes #3076.
* The root cause of those NaNs in `x` was not CVODES. The problem is uncaught integer overflow during steady-state simulations and asking CVODES to integrate until `t=inf`. Catch that.
* Add a dedicated error code
* Add debug-level expectations to all integration functions of `Solver`

### DIFF
--- a/src/solver_cvodes.cpp
+++ b/src/solver_cvodes.cpp
@@ -1178,14 +1178,6 @@ static int fxdot(realtype t, N_Vector x, N_Vector xdot, void* user_data) {
         return AMICI_MAX_TIME_EXCEEDED;
     }
 
-    if (t > 1e200
-        && !model->check_finite(gsl::make_span(x), ModelQuantity::xdot, t)) {
-        /* when t is large (typically ~1e300), CVODES may pass all NaN x
-           to fxdot from which we typically cannot recover. To save time
-           on normal execution, we do not always want to check finiteness
-           of x, but only do so when t is large and we expect problems. */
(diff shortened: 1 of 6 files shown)


src/solver_cvodes.cpp
```diff
@@ -1178,14 +1178,6 @@ static int fxdot(realtype t, N_Vector x, N_Vector xdot, void* user_data) {
         return AMICI_MAX_TIME_EXCEEDED;
     }
 
-    if (t > 1e200
```
FFroehlich, line 1181: Ah, it would actually be nice to stop simulation early (which the code so far did not achieve) and rather than simulating nonsense until maxsteps is reached.

dweindl: This isn't an issue anymore. Previously, model state became infinite only when asking CVODES to integrate until t=inf. This is now prevented at a higher level (`EventHandlingSimulator::run_steady_state`).

PR #3080, opened 2025-12-01
Implement event assignments in JAX

BSnelling:


python/sdist/amici/importers/sbml/__init__.py
```diff
@@ -1904,6 +1906,23 @@ def _process_events(self) -> None:
                 "priority": self._sympify(event.getPriority()),
             }
 
+            if self.jax:
+                if not initial_value:
```
FFroehlich: bit more nuanced than than, let's just not support events with initial values for now.

python/sdist/amici/importers/sbml/__init__.py
```diff
@@ -1904,6 +1906,23 @@ def _process_events(self) -> None:
...
+                    raise NotImplementedError(
+                        "The JAX backend does not support events with False initialValue."
+                    )
+                # Add a negative event for JAX models to handle
+                neg_event_id = event_id + "_negative"
+                neg_event_sym = sp.Symbol(neg_event_id)
+                self.symbols[SymbolId.EVENT][neg_event_sym] = {
```
FFroehlich: I think we should be able to drop this bit now if we additionally restrict number of trigger times to 1.

python/sdist/amici/jax/_simulation.py
```diff
@@ -457,11 +473,24 @@ def _handle_event(
     )
     roots_dir = jnp.sign(droot_dt)  # direction of the root condition function
 
-    h_next = h + jnp.where(
+    overall_dir = jnp.sign(jnp.sum(jnp.where(roots_found, h * roots_dir, 0.0)))
```
FFroehlich: this doesn't sound right. I don't think we want to compute an overall directions, every Heaviside variable needs to be updated independently based on the direction of zero crossing of its root function.

python/sdist/amici/importers/sbml/__init__.py
BSnelling: Do you mean we can drop the whole negative event? Or just the event priority part?

python/sdist/amici/importers/sbml/__init__.py
BSnelling: What should the condition be? `initial_value` is a required attribute for the trigger so it is always either `True` or `False`, not `None`, which means we'll have to support them.

python/sdist/amici/jax/_simulation.py
BSnelling: I've updated the logic here. Now, where roots were found the value of h gets flipped (0 -> 1, 1 -> 0) but otherwise remains the same. The test cases we're running all pass with this logic, but I'm not confident whether it'll hold when we introduce more event related features (priorities, delays etc).

python/sdist/amici/importers/sbml/__init__.py
FFroehlich: I was thinking of the whole negative events, but let's just keep it in for now.

python/sdist/amici/importers/sbml/__init__.py
FFroehlich: Ah. So we would have to check whether the initial value is consistent with the initial values of the trigger function. Sounds like it's easier to just support having `initial_value`. That essentially boils down to storing the values somewhere as part of the generated code and then replace the current code in https://github.com/AMICI-dev/AMICI/blob/a8dfd6e0963f32c1ce4647a652e93a271b14a053/python/sdist/amici/jax/model.py#L365

BSnelling: I believe I've implemented this now (see this commit: https://github.com/AMICI-dev/AMICI/pull/3080/commits/56e92676f4bb94803e5de11d998125648536f7b4).

python/sdist/amici/jax/_simulation.py
```diff
@@ -472,4 +490,34 @@ def _handle_event(
...
+    mask = jnp.array(
+        [
+            (roots_found & (roots_dir > 0.0) & (h == 0.0))
+            for _ in range(y0_next.shape[0])
+        ]
+    ).T
+    delx = delta_x(y0_next, p, tcl)
+    if y0_next.size:
+        delx = delx.reshape(delx.size // y0_next.shape[0], y0_next.shape[0],)
+    y0_up = jnp.where(mask, delx, 0.0)
```
FFroehlich, line 485:
```suggestion
    mask = roots_found & (roots_dir > 0.0) & (h == 0.0)
    delx = delta_x(y0_next, p, tcl)
    if y0_next.size:
        delx = delx.reshape(delx.size // y0_next.shape[0], y0_next.shape[0],)
    y0_up = jnp.where(mask[None, :], delx, 0.0)
```

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2665,6 +2672,35 @@ def has_event_assignments(self) -> bool:
...
+        return any(event.get_priority() is not None for event in self._events)
+    
+    def has_implicit_event_assignments(self) -> bool:
+        """
+        Checks whether the model has event assignments with implicit triggers
+
+        :return:
+            boolean indicating if event assignments with implicit triggers are present
+        """
+        allowed_syms = set(self.sym("p"))
```
FFroehlich: I think we want to temporarily drop `p` from the allowed syms here.

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2665,6 +2672,35 @@ def has_event_assignments(self) -> bool:
...
+        return any(event.updates_state and not event.has_explicit_trigger_times(allowed_syms) for event in self._events)
+    
+    def has_only_time_dependent_event_assignments(self) -> bool:
```
FFroehlich: I think we can drop this check (covered by check above after removing `p` from allowed syms.

BSnelling: Done!

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -1594,6 +1596,11 @@ def _compute_equation(self, name: str) -> None:
 
             self._eqs[name] = event_eqs
 
+        elif name == "x_old":
+            self._eqs[name] = sp.Matrix(
+                [state.get_x_rdata() for state in self.states()]
+            )
```
dweindl: Can we get rid of this one? So far, it gets its job done, but this won't be compatible with handling `useValuesFromTriggerTime=true` events. You'll have to store the state vector from the time the event triggered in the event queue and pass it to `_delta_x`.
If this is not easily possible at the moment, maybe use `x_rdata` directly in the jax `_delta_x`, to keep this workaround confined to that?

python/sdist/amici/importers/sbml/__init__.py
```diff
@@ -1880,6 +1882,19 @@ def _process_events(self) -> None:
                 "priority": self._sympify(event.getPriority()),
             }
 
+            if self.jax:
+                # Add a negative event for JAX models to handle
```
dweindl, line 1899: This seems to be somewhat redundant with https://github.com/AMICI-dev/AMICI/blob/a724161087fc5b73fa27b0728839c638c2e01c2f/python/sdist/amici/_symbolic/de_model.py#L2448-L2452
Not completely sure. Would be nice if we could unify that for jax/sundials. Fine to add a TODO comment for now.

python/sdist/amici/jax/model.py
```diff
@@ -308,6 +294,20 @@ def _root_cond_fn(
         """
         ...
 
+    @abstractmethod
+    def _delta_x(
```
dweindl, line 298: Fine as is, but it's a bit funny that $\Delta x$ updates $y$. This might be something to harmonize at some point. Using state $y$ would also be more in line with SUNDIALS.

python/sdist/amici/jax/ode_export.py
```diff
@@ -333,3 +349,13 @@ def set_name(self, model_name: str) -> None:
             )
 
         self.model_name = model_name
+
+def _parse_trigger_root(root: sp.Expr) -> str:
```
```suggestion
def _print_trigger_root(root: sp.Expr) -> str:
```
?

python/sdist/amici/_symbolic/de_model.py
BSnelling: Makes sense, I removed this change and isolated the workaround to the JAX code for now.

python/sdist/amici/importers/sbml/__init__.py
BSnelling: I've added a TODO note about this. We need to work out a way to trace the direction of the condition function for diffrax to monitor, through the diffrax solve. I didn't get far with it in my initial attempts.

python/sdist/amici/jax/model.py
BSnelling: IIUC this is a matter of different naming conventions in different parts of the code base so shouldn't be too hard to harmonise, but I'd want to check with @FFroehlich to be sure of that.

FFroehlich: open to changing that, but if we do so, let's do it in a separate PR.

PR #3104, opened 2025-12-12
`PetabSimulationResult` for `PetabSimulator.simulate`

### LINKED ISSUE #3094: Make `simulate_petab`, `PetabSimulator.simulate` return dataclass instead of dict
Change return type of `simulate_petab`, `PetabSimulator.simulate` etc to dataclass instead of dict.
I think this will be more convenient: compatible with type annotations, static analysis, and auto-completion.

dweindl:
Introduce dataclass `PetabSimulationResult` for `PetabSimulator.simulate` results. I think this will be more convenient: compatible with type annotations, static analysis, and auto-completion.

Closes #3094.

### DIFF
--- a/python/sdist/amici/sim/sundials/petab/_v2.py
+++ b/python/sdist/amici/sim/sundials/petab/_v2.py
@@ -6,7 +6,7 @@ import logging
 import numbers
 from collections import Counter
 from collections.abc import Sequence
-from typing import Any
+from dataclasses import dataclass, field
 
 import numpy as np
 import sympy as sp
@@ -16,21 +16,13 @@ from petab.v2 import ExperimentPeriod
 import amici
 from amici.logging import get_logger
 from amici.sim.sundials import SensitivityOrder
-from amici.sim.sundials.petab.v1._simulations import (
-    EDATAS,
-    LLH,
-    RDATAS,
-    RES,
-    S2LLH,
-    SLLH,
-    SRES,
-)
 
 logger = get_logger(__name__, log_level=logging.INFO)
 
 __all__ = [
     "PetabSimulator",
     "ExperimentManager",
+    "PetabSimulationResult",
 ]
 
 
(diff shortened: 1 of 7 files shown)


python/sdist/amici/sim/sundials/petab/_v2.py
```diff
@@ -508,6 +500,55 @@ def _get_placeholder_mapping(
...
+class PetabSimulationResult:
+    """
+    Container for results of a PEtab simulation.
+
+    Holds the per-experiment AMICI data objects and aggregated metrics
+    produced by :class:`PetabSimulator.simulate`.
+    """
+
+    #: List of :class:`amici.sim.sundials.ExpData` instances, one per
+    #: simulated experiment. These objects may be modified by subsequent
```
FFroehlich, line 513: Why would these be modified? I was wondering whether it would make sense to have dataclass(frozen=True)

dweindl: This depends on how the `ExpData`s are handled. For performance/memory reasons, it would be preferable to create them once, and store them in the `ExperimentManager`, and then only update the parameters when necessary.
We can then either always copy them when creating the `PetabSimulationResult` (which means we could also just not re-use them in the first place), or we only store the reference here, meaning that after the next simulation, the parameters may have changed.
Ideally, we'd split up `ExpData` into a parameter-dependent and parameter-independent part to circumvent this problem. That would however be a bigger thing (#3106).

The `frozen=True` decision is independent and would have only little effect because all current members are mutable. No objections to applying that, though.

FFroehlich: > This depends on how the `ExpData`s are handled. For performance/memory reasons, it would be preferable to create them once, and store them in the `ExperimentManager`, and then only update the parameters when necessary. We can then either always copy them when creating the `PetabSimulationResult` (which means we could also just not re-use them in the first place), or we only store the reference here, meaning that after the next simulation, the parameters may have changed. Ideally, we'd split up `ExpData` into a parameter-dependent and parameter-independent part to circumvent this problem. That would however be a bigger thing.
> 
> The `frozen=True` decision is independent and would have only little effect because all current members are mutable. No objections to applying that, though.

Oh, I see. Not great, but makes sense.

dweindl: Agreed.

Issue #3107, opened 2025-12-12
Include model entity IDs in ExpData / ReturnData


dweindl:
Currently, working with `ExpData` and `ReturnData` is less convenient than it could be, because only always needs the `Model` to get the parameter/observable/... IDs. It would be good to include the relevant IDs in each of those objects. 
We currently have an ugly hack for attaching parameters to `ReturnData` in Python for "full" reporting mode, but this should be changed.

The IDs are currently stored in free `std::array`s. It shouldn't be too hard to attach them to `ExpData` and `ReturnData` in a memory-efficient way. As they should be immutable anyway, there is no need for copying.

PR #3110, opened 2025-12-15
Refactor names and ids

dweindl:
Refactor name and ID handling as preparation for #3107.

* Store IDs/names in const string_view array and only pass around spans to avoid always constructing new vectors
* Store solver state IDs/names arrays instead of creating them dynamically, to be compatible with the previous point
* Update tests accordingly
* Add the respective typemaps
* Regenerate test models


FFroehlich (approved):
👍

PR #3115, opened 2026-01-09
Add support in JAX backend for PEtab v2

BSnelling:


.github/workflows/test_petab_sciml.yml
```diff
@@ -1,14 +1,14 @@
...
+# on:
+#   push:
+#     branches:
+#       - main
+#       - 'release*'
+#   pull_request:
+#     branches:
+#       - main
+#   merge_group:
+#   workflow_dispatch:
```
BSnelling, line 11: I'm deactivating this test suite temporarily as I plan to update the existing support for petab-sciml to petabv2 in another PR.

.github/workflows/test_petab_test_suite.yml
```diff
@@ -172,7 +172,7 @@ jobs:
           git clone https://github.com/PEtab-dev/petab_test_suite \
             && source ./venv/bin/activate \
             && cd petab_test_suite \
-            && git checkout c12b9dc4e4c5585b1b83a1d6e89fd22447c46d03 \
+            && git checkout 9542847fb99bcbdffc236e2ef45ba90580a210fa \
```
BSnelling, line 175: This commit is the tip of `main` (latest commit is from December 16th).

.github/workflows/test_petab_test_suite.yml
```diff
@@ -186,7 +186,7 @@ jobs:
         run: |
           source ./venv/bin/activate \
             && python3 -m pip uninstall -y petab \
-            && python3 -m pip install git+https://github.com/petab-dev/libpetab-python.git@8dc6c1c4b801fba5acc35fcd25308a659d01050e \
+            && python3 -m pip install git+https://github.com/petab-dev/libpetab-python.git@d57d9fed8d8d5f8592e76d0b15676e05397c3b4b \
```
BSnelling, line 189: This commit is the tip of `main` as of Jan 21st.  I updated it so as to include the implementation of sampling for prior distributions.

doc/examples/example_jax_petab/ExampleJaxPEtab.ipynb
```diff
@@ -594,86 +622,91 @@
...
-    "    solver=solver,\n",
-    "    problem_parameters=problem_parameters,\n",
-    "    scaled_parameters=True,\n",
-    "    scaled_gradients=True,\n",
-    ")"
-   ],
-   "id": "3f2ab1acb3ba818f"
+    "# %%timeit\n",
+    "# amici_model.simulate(petab_problem.get_x_nominal_dict())\n",
+    "# Results in an error regarding the FIM\n",
```
BSnelling: This last cell threw an error that the "FIM was not computed". I couldn't find a setting or config to calculate the FIM in `amici_model.simulate`.

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2676,14 +2684,48 @@ def has_priority_events(self) -> bool:
...
     def has_implicit_event_assignments(self) -> bool:
         """
         Checks whether the model has event assignments with implicit triggers
+        (i.e. triggers that are not time based).
 
         :return:
             boolean indicating if event assignments with implicit triggers are present
         """
-        return any(event.updates_state and not event.has_explicit_trigger_times({}) for event in self._events)
+        return any(event.updates_state and event._implicit_symbols() for event in self._events)
```
BSnelling: Context for this @FFroehlich: I updated this check so that the majority of the petabv2 test cases get run, but the SBML test cases which have implicit event triggers are still skipped, because I haven't yet added support for those.

python/sdist/amici/jax/petab.py
```diff
@@ -1578,67 +1783,73 @@ def run_simulations(
...
-        "simulation_conditions": simulation_conditions,
     }
 
-    if preequilibration_conditions:
-        preeqs, preresults = problem.run_preequilibrations(
-            preequilibration_conditions,
+    has_preeq = any(exp.periods[0].time < 0.0 for exp in experiments)
+
+    if has_preeq:
+        preeqs, preresults, h_preeqs = problem.run_preequilibrations(
```
BSnelling, line 1740: `h_preeqs` provides the state of the event triggers before the dynamic part of the simulation starts. It enables the `initialValue` SBML functionality to be supported.

tests/petab_test_suite/test_petab_v2_suite.py
```diff
@@ -148,6 +198,8 @@ def _test_case(case, model_type, version, jax):
     else:
         if (case, model_type, version) in (
             ("0016", "sbml", "v2.0.0"),
+            ("0024", "sbml", "v2.0.0"),
+            ("0025", "sbml", "v2.0.0"),
```
BSnelling, line 202: Locally these two cases were failing with ```RuntimeError: CVODE routine CVodeSensSStolerances failed with error code -40.```  I don't think it's related to my src changes, but the fact that I upgraded the version of the petab test suite, which means these test cases are running where they weren't before.

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2676,14 +2684,48 @@ def has_priority_events(self) -> bool:
...
+        """
+        Checks whether the model has events that can be triggered simultaneously
+        by checking for duplicate event triggering expressions that depend on time.
+
+        :return:
+            boolean indicating if simultaneous events are present
+        """
+        t_exprs = []
+        for event in self._events:
+            if "negative" in event._name:
```
FFroehlich: prone to false negatives, could we turn this into a property?

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2676,14 +2684,48 @@ def has_priority_events(self) -> bool:
...
+        t_exprs = list(itertools.chain(*t_exprs))
+
+        return len(t_exprs) != len(set(t_exprs))
+
+    
+    def _args_containing_t(self, expr):
+        t = sp.Symbol("t", real=True)
+        hits = []
+        for node in sp.preorder_traversal(expr):
+            if isinstance(node, sp.Min):
```
FFroehlich: what about `sp.Max`?

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2676,14 +2684,48 @@ def has_priority_events(self) -> bool:
...
+            t_args = self._args_containing_t(trigger)
+            t_exprs.append(t_args)
+
+        t_exprs = list(itertools.chain(*t_exprs))
+
+        return len(t_exprs) != len(set(t_exprs))
+
+    
+    def _args_containing_t(self, expr):
+        t = sp.Symbol("t", real=True)
```
FFroehlich: all states have an implicit time dependency, wouldn't this produce false negatives?

I think there we define something like `amci_time_symbol` somewhere.

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2676,14 +2684,48 @@ def has_priority_events(self) -> bool:
         """
         return any(event.get_priority() is not None for event in self._events)
     
+    def has_simultaneous_events(self) -> bool:
+        """
+        Checks whether the model has events that can be triggered simultaneously
```
FFroehlich: I don't think we can reliably test this pre-runtime. say you have triggers `p1-t` and `p2-t`. You end up with simultaneous events if and only if p1==p2, but you will only know p1 and p2 at runtime.

python/sdist/amici/importers/petab/v1/sbml_import.py
```diff
@@ -304,7 +306,10 @@ def import_model_sbml(
 
     if validate:
         logger.info("Validating PEtab problem ...")
-        petab.lint_problem(petab_problem)
+        if isinstance(petab_problem, petabv2.Problem):
```
FFroehlich: This is the v1 importer, I would suggest throw an error here for v2 problems, do the v1/v2 branching elsewhere and keep other v2 code out of this function.

python/sdist/amici/jax/_simulation.py
```diff
@@ -108,55 +109,76 @@ def cond_fn(carry):
...
-            jnp.logical_or(
-                diffrax.is_successful(sol.result),
-                diffrax.is_event(sol.result),
-            ),
-            sol.ys[-1],
-            jnp.inf * jnp.ones_like(sol.ys[-1]),
+        t_start, y0, h, _, _ = carry
+
+        # skip the solve if already at steady state
+        # TODO: remove once https://github.com/patrick-kidger/diffrax/issues/720 is resolved?
```
FFroehlich: 😢

FFroehlich (changes requested):

python/sdist/amici/importers/petab/v1/sbml_import.py
BSnelling: I've removed this, turns out I wasn't using this importer anymore anyway.

python/sdist/amici/_symbolic/de_model.py
BSnelling: Good point. I've managed to implement sequential event assignments, which lets us support simultaneous events and remove this whole method.  Note that there is a limitation to this, event priorities are still not supported. The assignments are done based on the order the events are stored in.

python/sdist/amici/_symbolic/de_model.py
BSnelling: Removed with simultaneous events check

python/sdist/amici/_symbolic/de_model.py
BSnelling: Removed with simultaneous events check

python/sdist/amici/_symbolic/de_model.py
BSnelling: Removed with simultaneous events check

python/sdist/amici/jax/petab.py
```diff
@@ -1015,36 +1090,103 @@ def _map_model_parameter_value(
...
-                else mapping.scale_map_sim_var[pname]
-                for pname in self.model.parameter_ids
+                petabv1.LIN
+                for _ in self.model.parameter_ids
             ]
         )
+
         return self._unscale(p, pscale)
+    
+    def _map_experiment_model_parameter_value(
```
BSnelling, line 1059: FYI @FFroehlich from our discussion, there is parameter mapping logic here too.

FFroehlich (approved):

dweindl:
Sorry for the delay, I'll try to review tomorrow or the day after.

doc/examples/example_jax_petab/ExampleJaxPEtab.ipynb
```diff
@@ -42,13 +43,22 @@
     ")\n",
     "\n",
     "# Load the PEtab problem from the YAML file\n",
-    "petab_problem = petab.Problem.from_yaml(yaml_url)\n",
+    "# petab_problem = petab.Problem.from_yaml(yaml_url)\n",
```
dweindl: Remove if no PEtab v1 is supported anymore? Also the dead `simulate_petab` code further down.

doc/examples/example_jax_petab/ExampleJaxPEtab.ipynb
```diff
@@ -42,13 +43,22 @@
...
+    "\n",
+    "# Load the PEtab problem from the YAML file as a PEtab v2 problem\n",
+    "# (the JAX backend only supports PEtab v2)\n",
+    "petab_problem = petab1to2.petab1to2(yaml_url)\n",
```
dweindl:
```python
from petab.v2 import Problem

petab_problem = Problem.from_yaml(yaml_url)
```
petab.v2.Problem will automatically upconvert

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -1141,6 +1141,14 @@ def _generate_symbol(self, name: str) -> None:
...
             )
             return
+        elif name == "allh":
+            self._syms[name] = sp.Matrix(
+                [
+                    sym
+                    for sym, _ in zip(self.sym("h"), self._events)
+                ]
+            )
+            return
```
dweindl: Do we need that? It's basically just `DEModel.sym("h")`, isn't it?

python/sdist/amici/_symbolic/de_model_components.py
```diff
@@ -875,6 +883,18 @@ def get_trigger_times(self) -> set[sp.Expr]:
...
+        """Get implicit symbols in the event trigger function.
+        That is, all symbols except time and petab indicator variables.
+        """
+        symbols = [str(s) for s in list(self.get_val().free_symbols)]
+        implicit_symbols = []
+        for s in symbols:
+            if (s.startswith("_petab_") and "indicator" in s) or s == "t":
+                continue
+            implicit_symbols.append(s)
+        return len(implicit_symbols) > 0
```
dweindl: This sub-package should stay PEtab-independent. I think a better way to handle that would be passing all indicator variables as `fixed_parameters` to `SbmlImporter._build_ode_model` via `SbmlImporter.sbml2jax` and then use the same logic as for the non-jax event handling.

python/sdist/amici/_symbolic/de_model.py
dweindl: Related to the comment on `_implicit_symbols` below, I'd prefer keeping the old logic here and passing the indicator symbols to `has_explicit_trigger_times`.

python/sdist/amici/jax/_simulation.py
dweindl: Apparently fixed in https://github.com/patrick-kidger/diffrax/releases/tag/v0.7.1

python/sdist/amici/jax/jax.template.py
```diff
@@ -30,7 +30,7 @@ def _xdot(self, t, x, args):
         TPL_X_SYMS = x
         TPL_P_SYMS = p
         TPL_TCL_SYMS = tcl
-        TPL_IH_SYMS = h
+        TPL_ALLH_SYMS = h
```
dweindl: `TPL_ALL_H_SYMS`?

python/sdist/amici/jax/petab.py
```diff
@@ -60,30 +65,43 @@ def jax_unscale(
...
+        self.__dict__.update(petab_problem.__dict__)
+        self.hybridization_df = _get_hybridization_df(petab_problem)
+
+class HybridV2Problem(petabv2.Problem):
     hybridization_df: pd.DataFrame
+    extensions_config: dict
 
-    def __init__(self, petab_problem: petab.Problem):
+    def __init__(self, petab_problem: petabv2.Problem):
+        if not hasattr(petab_problem, "extensions_config"):
```
dweindl, line 95: `extensions_config` doesn't exist in `petab.v2.Problem`. There's `petab.v2.Problem.config.extensions`.
However, I just noticed that it's specified as `extensions: list[ExtensionConfig] | dict` there --- this needs to be fixed in libpetab (https://github.com/PEtab-dev/libpetab-python/issues/474). So fine as is for now, but don't be surprised that this condition is always true...

python/sdist/amici/jax/ode_export.py
```diff
@@ -217,6 +218,7 @@ def _generate_jax_code(self) -> None:
             "x",
             "tcl",
             "ih",
+            "allh",
```
```suggestion
            "all_h",
```
To avoid confusion with llh and sllh in amici?

python/sdist/amici/jax/petab.py
```diff
@@ -143,13 +163,21 @@ def __init__(self, model: JAXModel, petab_problem: petab.Problem):
...
+            self.simulation_conditions = scs.simulationConditionId.to_list()
+        else:
+            scs = petab_problem.get_simulation_conditions_from_measurement_df()
+            self.simulation_conditions = tuple(tuple(sc) for sc in scs.values)
         self._petab_problem = _get_hybrid_petab_problem(petab_problem)
         self.parameters, self.model = (
             self._initialize_model_with_nominal_values(model)
         )
-        self._parameter_mappings = self._get_parameter_mappings(scs)
+        if isinstance(petab_problem, petabv1.Problem):
```
dweindl: Merge with the `if isinstance` above?

python/sdist/amici/jax/petab.py
```diff
@@ -143,13 +163,21 @@ def __init__(self, model: JAXModel, petab_problem: petab.Problem):
         :param petab_problem:
             PEtab problem to simulate.
         """
-        scs = petab_problem.get_simulation_conditions_from_measurement_df()
-        self.simulation_conditions = tuple(tuple(sc) for sc in scs.values)
+        if isinstance(petab_problem, petabv2.Problem):
```
dweindl: Is the plan to support both, petab v1 and v2, or is this just until petab-sciml is updated?

python/sdist/amici/jax/petab.py
```diff
@@ -1015,36 +1090,103 @@ def _map_model_parameter_value(
...
+        Get values for the given parameter `pname` from the relevant petab tables.
+
+        :param pname: PEtab parameter id
+        :param p_index: Index of the parameter in the model's parameter list
+        :param experiment: PEtab experiment
+        :param is_preeq: Whether to get preequilibration or simulation parameter value
+        :return: Value of the parameter
+        """
+        condition_ids = []
+        for p in experiment.periods:
```
dweindl:
```suggestion
        for p in experiment.sorted_periods:
```

`experiment.periods` isn't guaranteed to be sorted, but it's currently assumed, right?

python/sdist/amici/jax/petab.py
```diff
@@ -1015,36 +1090,103 @@ def _map_model_parameter_value(
...
+            if is_preeq:
+                if p.time >= 0.0:
+                    continue
+                else:
+                    condition_ids = p.condition_ids
+                    break
+            else:
+                if p.time < 0.0:
```
dweindl: This is to find the first non-pre-equilibration condition, right? If so, this fill fail nor negative t0.
There is `petab.v2.core.ExperimentPeriod.is_preequilibration` which might make these checks more explicit.

dweindl (commented):
Cool, thanks. Here some initial comments. Didn't manage to go over everything yet.

doc/examples/example_jax_petab/ExampleJaxPEtab.ipynb
```diff
@@ -594,86 +608,65 @@
...
-    "    petab_problem,\n",
-    "    amici_model=amici_model,\n",
-    "    solver=solver,\n",
-    "    problem_parameters=problem_parameters,\n",
-    "    scaled_parameters=True,\n",
-    "    scaled_gradients=True,\n",
-    ")"
-   ],
-   "id": "3f2ab1acb3ba818f"
+    "amici_model.simulate(petab_problem.get_x_nominal_dict())"
```
BSnelling, line 668: When the notebook runs there is an error from this cell that the FIM was not computed: https://github.com/AMICI-dev/AMICI/actions/runs/21985309084/job/63517900881?pr=3115

@dweindl can you advise on that?

python/sdist/amici/_symbolic/de_model_components.py
```diff
@@ -855,11 +863,19 @@ def has_explicit_trigger_times(
...
             for t in self._t_root
         )
+    
+    def has_implicit_triggers(
+        self, allowed_symbols: set[sp.Symbol] | None = None
+    ) -> bool:
+        """Check whether the event has implicit triggers.
+        """
+        t = self.get_val()
+        return not t.free_symbols.issubset(allowed_symbols)
```
BSnelling, line 878: I removed the PEtab related references from this function and am passing the indicator variables as `fixed_parameters`. I wanted to include this logic in the `has_explicit_trigger_times` function but in all the JAX examples, `self._t_root` is not set.  Should `_t_root` be set during the import of petab events?

python/sdist/amici/jax/petab.py
```diff
@@ -143,13 +163,18 @@ def __init__(self, model: JAXModel, petab_problem: petab.Problem):
         :param petab_problem:
             PEtab problem to simulate.
         """
-        scs = petab_problem.get_simulation_conditions_from_measurement_df()
-        self.simulation_conditions = tuple(tuple(sc) for sc in scs.values)
+        if isinstance(petab_problem, petabv1.Problem):
+            raise TypeError(
+               "JAXProblem does not support PEtab v1 problems. Upgrade the problem to PEtab v2."
```
BSnelling, line 168: The plan is just to support v2, so I've added this error if a v1 problem is encountered.

BSnelling (commented):
Thank you for the review @dweindl! I have implemented most points, but I have a couple of outstanding questions, if you could advise on those.

doc/examples/example_jax_petab/ExampleJaxPEtab.ipynb
dweindl: My fault. I hope https://github.com/AMICI-dev/AMICI/pull/3125 fixes that.

python/sdist/amici/_symbolic/de_model_components.py
dweindl: Oh, okay. I think this might be a general issue, not just with JAX. My guess is that sympy isn't able to solve those petab event trigger piecewise expressions for $t$ (here https://github.com/AMICI-dev/AMICI/blob/20b07e9a8c19bdb7bf94a562f61b3d08dc0299d0/python/sdist/amici/_symbolic/de_model_components.py#L768-L772). That should be feasible, but I'm not sure how easy it is to implement. I won't be able to look into that for at least another week. Okay to keep your previous approach for now then. Please open an issue to address that later.

python/sdist/amici/sim/jax/__init__.py
```diff
@@ -1 +1,127 @@
...
+        if (
+            petabv1.OBSERVABLE_PARAMETERS
+            in problem._petab_problem.measurement_df
+        ):                
+            df_sc[petabv2.C.OBSERVABLE_PARAMETERS] = (
+                problem._petab_problem.measurement_df.query(
+                    f"{petabv2.C.EXPERIMENT_ID} == '{experiment_id}'"
+                )[petabv2.C.OBSERVABLE_PARAMETERS]
+            )
+        if petabv1.NOISE_PARAMETERS in problem._petab_problem.measurement_df:
```
dweindl:
```suggestion
                petabv2.OBSERVABLE_ID: obs,
                petabv2.C.EXPERIMENT_ID: [experiment_id] * len(t),
                petabv2.TIME: t[problem._ts_masks[ic, :]],
                petabv2.SIMULATION: y[ic, problem._ts_masks[ic, :]],
            },
            index=problem._petab_measurement_indices[ic, :],
        )
        if (
            petabv2.OBSERVABLE_PARAMETERS
            in problem._petab_problem.measurement_df
        ):                
            df_sc[petabv2.C.OBSERVABLE_PARAMETERS] = (
                problem._petab_problem.measurement_df.query(
                    f"{petabv2.C.EXPERIMENT_ID} == '{experiment_id}'"
                )[petabv2.C.OBSERVABLE_PARAMETERS]
            )
        if petabv2.NOISE_PARAMETERS in problem._petab_problem.measurement_df:
```

No effect, but since the function is specifically for petab v2, let's use the constants from there.

python/sdist/amici/jax/petab.py
dweindl: Great. Can you please replace all the `petab.v1` constants by those in `petab.v2` to make it clear that we don't want to deal with v1 here?
It seems there is still quite a bit of v1 code left here. Probably all occurrences of `simulationConditionId`/`SIMULATION_CONDITION_ID`. Is that still needed for petab-sciml? If not please remove.

python/sdist/amici/jax/petab.py
```diff
@@ -305,38 +330,63 @@ def _get_measurements(
...
+                petabv1.OBSERVABLE_TRANSFORMATION
                 in self._petab_problem.observable_df
             ):
                 iy_trafos = np.array(
                     [
                         SCALE_TO_INT[
                             self._petab_problem.observable_df.loc[
-                                oid, petab.OBSERVABLE_TRANSFORMATION
+                                oid, petabv1.OBSERVABLE_TRANSFORMATION
                             ]
```
dweindl: As a general comment: With `petab.v2.Problem`, all `.*_df` properties construct a new `DataFrame` on each call. Try to avoid accessing those properties inside loops. 

In many cases it might be more convenient to directly use the underlying objects instead of the DataFrames (e.g., `Problem.observables` instead of `Problem.observable_df`).

dweindl (commented):
Thanks, @BSnelling. Here a few more comments. Okay for me to merge once all tests pass and to address the remaining issues together with the petab-sciml update if that's more convenient.

python/sdist/amici/_symbolic/de_model_components.py
BSnelling: Indeed, sympy is failing to solve these expressions. Opened https://github.com/AMICI-dev/AMICI/issues/3126

python/sdist/amici/jax/petab.py
BSnelling: I have updated these. There are some instances left which I will handle separately in the PEtab SciML update. I expect there will also be more v1 code I can safely delete as well.

python/sdist/amici/jax/petab.py
BSnelling: Thanks for pointing this out! I have updated a few instances and will look out for others in the sciml code.

tests/petab_test_suite/test_petab_v2_suite.py
dweindl: Addressed in #3124

Issue #3041, opened 2025-11-15
Move simulation code to subpackage


dweindl:
Move simulation code to subpackages under amici.sim, amici.simulation, amici.simulations or similar,  Don't import the amici extension by default.

For example, everything related to the swig interface can go to amici.simulation.sundials, petab+sundials functionality to amici.simulation.sundials.petab, amici.simulation.sundials.petab.v1.
Similar for amici.simulation.jax, amici.simulation.jax.petab.

- [x] Wait for #3039 
- [x] Move code related to sundials-based simulation to `amici.sim.sundials`
- [x] Move code related to simulating jax-based models to `amici.sim.jax` (after update of JAX-PEtab code) (#3135)
- [x] move simulation related code from `amici.importers.petab` to `amici.sim.sundials` (started in #3086)
  - [x] petab v2 (#3093)
  - [x] parameter mapping (#3136)
  - [x] amici.debug (#3091)

PR #3135, opened 2026-02-25
Move amici.jax to amici.exporters.jax and amici.sim.jax

dweindl:
Related to #3041.


python/sdist/amici/exporters/jax/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+This module provides an interface to generate and use AMICI models with JAX.
+Please note that this module is experimental, the API may substantially change
+in the future. Use at your own risk and do not expect backward compatibility.
+"""
+
+from warnings import warn
+
+from .nn import Flatten, cat, generate_equinox, tanhshrink
+
+warn(
```
FFroehlich: I think we can remove this warning at this point

python/sdist/amici/sim/jax/__init__.py
```diff
@@ -1,138 +1,32 @@
...
+    petab_simulate,
+    run_simulations,
+)
+
+warn(
+    "The JAX module is experimental and the API may change in the future.",
+    ImportWarning,
+    stacklevel=2,
+)
+
```
FFroehlich:
```suggestion
```

python/sdist/amici/exporters/jax/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    "The JAX module is experimental and the API may change in the future.",
+    ImportWarning,
+    stacklevel=2,
+)
```
FFroehlich:
```suggestion
```

FFroehlich (approved):
👍  thanks!

python/sdist/amici/exporters/jax/__init__.py
dweindl: Great. Will do.

PR #3156, opened 2026-03-09
doc: Fix formatting, update names, fix swig-ignores

dweindl:
Fix formatting, update object names, fix swig-ignores


FFroehlich (approved):
👍

PR #3165, opened 2026-03-23
Update PEtab SciML support to PEtab v2

BSnelling:


tests/sciml/test_sciml.py
```diff
@@ -178,36 +182,68 @@ def test_net(test):
...
+
     with open(test_dir / "petab" / "problem.yaml") as f:
         petab_yaml = safe_load(f)
     with open(test_dir / "solutions.yaml") as f:
         solutions = safe_load(f)
 
     with change_directory(test_dir / "petab"):
-        from petab.v2 import Problem
+        # HACK!! Again!! Around "array" in parameters table
+        petab_problem = _v2_sciml_problem_helper(petab_yaml, test_dir / "petab")
```
BSnelling, line 196: Temporary solution to importing the problem, until there is a SciML version of the petab linter available.

tests/sciml/test_sciml.py
```diff
@@ -290,6 +320,146 @@ def test_ude(test):
...
+    leaves = jax.tree_util.tree_leaves(arrays)
+
+    total = jnp.array(0.0)
+    for leaf in leaves:
+        if leaf is not None:
+            total = total + _uniform_logpdf(leaf, low, high)
+    return total
+
+
+def _model_logprior(model, layer1_bias_std=1.0, layer1_weight_std=1.0) -> jnp.ndarray:
```
BSnelling, line 452: Is there a better way to calculate these priors?  This feels like a lot of helper functions to add to a test file.

doc/examples/example_jax_petab/ExampleJaxPEtab.ipynb
```diff
@@ -228,7 +228,7 @@
     "import jax\n",
     "\n",
     "# Generate random noise to update the parameters\n",
-    "noise = (\n",
+    "noise = jax.numpy.exp(\n",
```
FFroehlich, line 231: Leftover temporary edits?

python/sdist/amici/exporters/jax/ode_export.py
```diff
@@ -306,6 +310,42 @@ def _generate_jax_code(self) -> None:
             tpl_data,
         )
 
+    def _generate_array_inputs_init(self) -> str:
```
FFroehlich, line 313: Something I hadn’t really thought about before. This relies on the HDF5 file staying in the same place, effectively making the module non-portable. Alternative would be to copy the HDF5 file to the module directory, but that may also be a bad idea, depending on the size of the file.

python/sdist/amici/importers/petab/_petab_importer.py
```diff
@@ -561,6 +582,157 @@ def _get_observation_model(
...
+            or self.petab_problem.model.type_id != MODEL_TYPE_SBML
+        ):
+            raise NotImplementedError(
+                "petab_sciml extension is currently only supported for sbml models"
+            )
+
+        from petab_sciml.standard import NNModelStandard
+
+        config = self.petab_problem.config.extensions["sciml"]
+        # TODO: only accept YAML format for now
```
FFroehlich, line 614: ^ agreed, fine for now but should be easy to implement

python/sdist/amici/sim/jax/petab.py
```diff
@@ -837,42 +847,16 @@ def _initialize_model_with_nominal_values(
...
-        if isinstance(self._petab_problem, HybridV2Problem):
-            param_map = {
-                p.id: p.nominal_value for p in self._petab_problem.parameters
-            }
-            parameter_array = jnp.array(
-                [float(param_map[pval]) for pval in self.parameter_ids]
-            )
-        else:
-            parameter_array = self._create_scaled_parameter_array()
+        # if isinstance(self._petab_problem, HybridV2Problem):
```
FFroehlich: ?

python/sdist/amici/sim/jax/petab.py
```diff
@@ -2079,7 +2066,7 @@ def _conditions_to_experiment_map(
 
 def _try_float(value):
     try:
-        return float(value)
+        return jnp.asarray(float(value), dtype=jnp.float64)
```
FFroehlich: https://docs.jax.dev/en/latest/default_dtypes.html same goes for int32 values above

FFroehlich (approved):
great! only some minor comments

.github/workflows/test_petab_sciml.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         run: |
           source ./venv/bin/activate \
             && python3 -m pip uninstall -y petab \
-            && python3 -m pip install git+https://github.com/petab-dev/libpetab-python.git@sciml
+            && python3 -m pip install git+https://github.com/petab-dev/libpetab-python.git@44c8062ce1b87a74a0ba1bd2551de0cdc2a13ff1
```
dweindl:
```suggestion
            && python3 -m pip install git+https://github.com/petab-dev/libpetab-python.git@main
```
I think we can go for `main` again

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2499,26 +2499,67 @@ def _process_hybridization(self, hybridization: dict) -> None:
...
-                comp
-                for comp in self._components
-                if str(comp.get_sym()) in net["input_vars"]
-            ]
-            # sort inputs by order in input_vars
-            inputs = sorted(
-                inputs,
-                key=lambda comp: net["input_vars"].index(str(comp.get_sym())),
-            )
+            comp_by_sym = {str(comp.get_sym()): comp for comp in self._components}
```
dweindl:
```suggestion
            comp_by_sym = {comp.get_id(): comp for comp in self._components}
```

I think this will be more compatible with pysb models, but not completely sure if this is still an issue.

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2499,26 +2499,67 @@ def _process_hybridization(self, hybridization: dict) -> None:
...
+                        continue
+
+                    if {str(s) for s in expr.free_symbols} - set(comp_by_sym):
+                        unresolved_vars.append(input_var)
+                        continue
+
+                    expr_sym = sp.Symbol(
+                        f"_nn_{net_id}_input_{len(inputs)}", real=True
+                    )
+                    new_expr_comp = Expression(
```
dweindl, line 2528: Use `amici.importers.utils.symbol_with_assumptions` instead of plain `Symbol` for maintainability.

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2499,26 +2499,67 @@ def _process_hybridization(self, hybridization: dict) -> None:
...
+
             if len(inputs) != len(net["input_vars"]):
-                found_vars = {str(comp.get_sym()) for comp in inputs}
-                missing_vars = set(net["input_vars"]) - found_vars
-                raise ValueError(
-                    f"Could not find all input variables for neural network {net_id}. "
-                    f"Missing variables: {sorted(missing_vars)}"
-                )
+                missing_vars = set(unresolved_vars)
+                if missing_vars == set(["array"]):
```
dweindl:
```suggestion
                if missing_vars == {'array'}:
```

python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2609,9 +2651,18 @@ def _process_hybridization(self, hybridization: dict) -> None:
...
-                    *[input.get_sym() for input in inputs], parts["ind"]
+                nn_call = sp.Function(net_id)(
+                    *[input.get_sym() for input in inputs], parts["index"]
                 )
+                formula = parts["formula"]
+                petab_id = parts["petab_id"]
+                if formula == petab_id:
+                    out_val = nn_call
+                else:
+                    out_val = sp.sympify(
```
dweindl: Not sure about the exact semantics here, but mind the potential difference between PEtab expression strings and sympy expression strings. `petab.v2.math.sympify.sympify_petab` might be more appropriate here. However, we don't support `locals` there yet.

python/sdist/amici/importers/petab/_petab_importer.py
```diff
@@ -216,6 +217,21 @@ def _preprocess_sbml(self):
...
+                change.target_id
+                for condition in self.petab_problem.conditions
+                for change in condition.changes
+            }
+            mapping_petab_ids = {
+                mapping.petab_id for mapping in self.petab_problem.mappings
+            }
+            if condition_targets.intersection(mapping_petab_ids):
+                raise NotImplementedError(
+                    "The JAX backend does not currently support PEtab problems where network"
```
dweindl:
```suggestion
                    "The JAX backend does not currently support PEtab problems where network "
```

python/sdist/amici/importers/petab/_petab_importer.py
```diff
@@ -561,6 +582,157 @@ def _get_observation_model(
...
+
+        from petab_sciml.standard import NNModelStandard
+
+        config = self.petab_problem.config.extensions["sciml"]
+        # TODO: only accept YAML format for now
+        hybridizations = [
+            pd.read_csv(hf, sep="\t")
+            for hf in config["hybridization_files"]
+        ]
+        hybridization_table = pd.concat(hybridizations)
```
dweindl, line 618: Fine as is for now, but let's add that to libpetab-python.

python/sdist/amici/importers/petab/_petab_importer.py
```diff
@@ -561,6 +582,157 @@ def _get_observation_model(
...
+                        for cid in p.condition_ids
+                    ],
+                }
+
+        non_estimated_param_ids = {
+            p.id
+            for pt in self.petab_problem.parameter_tables
+            for p in pt.elements
+            if not p.estimate
+        }
```
dweindl: Should be equivalent to `set(problem.x_free_ids)`, but either is fine.

python/sdist/amici/sim/jax/petab.py
```diff
@@ -148,9 +149,13 @@ class JAXProblem(eqx.Module):
     _np_indices: np.ndarray
     _petab_measurement_indices: np.ndarray
     _petab_problem: petabv1.Problem | HybridProblem | petabv2.Problem
```
```suggestion
    _petab_problem: HybridProblem | petabv2.Problem
```

If v1.Problem is not supported, let's remove it here to avoid static type checking issues.

python/sdist/amici/sim/jax/petab.py
```diff
@@ -837,42 +847,16 @@ def _initialize_model_with_nominal_values(
...
-            }
-            parameter_array = jnp.array(
-                [float(param_map[pval]) for pval in self.parameter_ids]
-            )
-        else:
-            parameter_array = self._create_scaled_parameter_array()
+        # if isinstance(self._petab_problem, HybridV2Problem):
+        param_map = {
+            p.id: p.nominal_value for p in self._petab_problem.parameters
+        }
```
dweindl: alternatively `v2.Problem.get_x_nominal_dict()`

python/sdist/amici/sim/jax/petab.py
```diff
@@ -957,107 +965,64 @@ def _is_net_input(model_id):
...
-        hybridization_parameter_map = {
-            petab_id: self._petab_problem.hybridization_df.loc[
-                petab_id, "targetValue"
-            ]
-            for petab_id in model_id_map.values()
-            if petab_id in set(self._petab_problem.hybridization_df.index)
+        parameters_map = {
+            p.id: p.nominal_value
+            for pt in self._petab_problem.parameter_tables
+            for p in pt.elements
```
dweindl: alternatively `v2.Problem.get_x_nominal_dict()`

python/tests/test_sciml.py
```diff
@@ -399,12 +397,8 @@ def mock_de_model(self):
...
         # Add some differential states
         model._differential_states = [
-            DifferentialState(
-                sp.Symbol("x1"), "state1", sp.Float(0.0), sp.Float(0.1)
-            ),
-            DifferentialState(
-                sp.Symbol("x2"), "state2", sp.Float(0.0), sp.Float(0.2)
-            ),
+            DifferentialState(sp.Symbol("x1"), "state1", sp.Float(0.0), sp.Float(0.1)),
+            DifferentialState(sp.Symbol("x2"), "state2", sp.Float(0.0), sp.Float(0.2)),
```
dweindl: Probably not yet a problem for the jax interface, but better use `amici.importers.utils.symbol_with_assumptions`. Instead of `Symbol`.

tests/sciml/test_sciml.py
dweindl: We could add some jax version to `petab.v2.priors.Prior`, but in the medium term, I am planning to move all this prior (or general probability distribution) handling to a separate library to be used by amici/petab/pypesto.
I'd leave it as is for now.

dweindl (approved):
Thanks, great. Please install/run (`pre-commit run --all-files`) the [pre-commit hooks](https://amici.readthedocs.io/en/latest/development.html#python) before merging.

doc/examples/example_jax_petab/ExampleJaxPEtab.ipynb
BSnelling: This is intentional actually. The petab v1 version of this problem had log10 scaled parameters, so we were effectively add log scaled normally distributed noise to log scaled parameters. I figured the equivalent for the v2 version, where parameters are not scaled, was to exponentiate the noise and then multiply the parameters by it. Does that make sense?

python/sdist/amici/exporters/jax/ode_export.py
BSnelling: Agreed, this is a potential failure point. How important is portability of the module? Do we expect users to want to export the model separately from where they have imported the PEtab problem?

python/sdist/amici/importers/petab/_petab_importer.py
BSnelling: Do you mean support for loading the hybridization table should go in libpetab-python? I didn't think that would belong in libpetab-python, because hybridization is part of an extension.

dweindl: I guess, you are right because petab-sciml is not an "official" extension yet. But once it is, validation should ideally be covered there, which includes reading those files.

python/sdist/amici/_symbolic/de_model.py
BSnelling: I can use `sympify_petab` without locals.  I kept the import just inside the conditional for now to avoid having the add `petab` as a dependency for whole module.

PR #3175, opened 2026-06-01
Implement sciml initialisation test cases

BSnelling:
Adding support and test cases for initialisation of neural networks: https://github.com/PEtab-dev/petab_sciml_testsuite/tree/main/test_cases/initialization

### DIFF
--- a/python/sdist/amici/sim/jax/petab.py
+++ b/python/sdist/amici/sim/jax/petab.py
@@ -708,8 +708,22 @@ class JAXProblem(eqx.Module):
         :param par_arrays:
             Parameter arrays loaded from files
         """
+        mapping_df = self._petab_problem.mapping_df
+
+        def _lookup_mid(pname: str) -> str | None:
+            if mapping_df is not None and pname in mapping_df.index:
+                return mapping_df.loc[pname, petabv2.C.MODEL_ENTITY_ID]
+            return None
+
         for table in self._petab_problem.parameter_tables:
-            for parameter in table.elements:
+            # Array-indexed params (e.g. net.parameters[layer]) must be processed
+            # after scalar params to avoid overwriting a full-layer assignment.
+            sorted_params = sorted(
+                table.elements,
+                key=lambda p: "parameters[" in str(_lookup_mid(p.id) or ""),
+            )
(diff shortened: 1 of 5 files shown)


python/sdist/amici/sim/jax/petab.py
```diff
@@ -708,8 +708,22 @@ def _extract_nominal_values_from_petab(
         :param par_arrays:
             Parameter arrays loaded from files
         """
+        mapping_df = self._petab_problem.mapping_df
+
+        def _lookup_mid(pname: str) -> str | None:
```
FFroehlich: is there no such function in the petab base library?

python/sdist/amici/sim/jax/petab.py
```diff
@@ -708,8 +708,22 @@ def _extract_nominal_values_from_petab(
...
+            if mapping_df is not None and pname in mapping_df.index:
+                return mapping_df.loc[pname, petabv2.C.MODEL_ENTITY_ID]
+            return None
+
         for table in self._petab_problem.parameter_tables:
-            for parameter in table.elements:
+            # Array-indexed params (e.g. net.parameters[layer]) must be processed
+            # after scalar params to avoid overwriting a full-layer assignment.
+            sorted_params = sorted(
```
FFroehlich: this does not ensure that `net.parameter[layer].weight[5]` takes precedence over `net.parameter[layer].weight` and `net.parameter[layer]`. I think counting the number of `.` and `[` in `str(_lookup_mid(p.id) or "")` should work. Or is that not supported by spec/implementation?

FFroehlich (approved):

python/sdist/amici/sim/jax/petab.py
BSnelling: Not really. There is `MappingTable.get` but I'd have to iterate through the array of mapping tables in the petab problem so it's convenient to use the`mapping_df` which concatenates everything in this case.

python/sdist/amici/sim/jax/petab.py
BSnelling: Good point, added!

PR #3176, opened 2026-06-05
Implement missing layer types 

BSnelling:
Adds custom eqx.Module classes for PyTorch layers:
- BatchNorm
- InstanceNorm
- AlphaDropout
- Bilinear

Still unsupported because JAX won't support state tracking:
- MaxPool with dilation
- Dropout with inplace


tests/sciml/test_sciml.py
```diff
@@ -234,17 +221,34 @@ def test_ml_model_import(test):
...
+                )
+                for _ in range(iters):
+                    res += net.forward(
+                        input, key=jr.PRNGKey(np.random.randint(0, 2**31))
+                    )
+                np.testing.assert_allclose(
+                    res / iters,
+                    output,
+                    atol=2e-2,
+                    rtol=2e-2,
```
BSnelling, line 238: Relaxing these tolerances as per Petab.jl: https://github.com/sebapersson/PEtab.jl/blob/69bc774ca978ab1ff44c3ef29abbd4a085bdca99/test/petab_sciml_testsuite/helper.jl#L139

tests/sciml/test_sciml.py
```diff
@@ -234,17 +221,34 @@ def test_ml_model_import(test):
...
-            net = eqx.nn.inference_mode(net)
 
-            if test == "net_004_alt":
-                return  # skipping, no support for non-cross-correlation in equinox
+            if test in ("019", "020", "021", "022"):
+                iters = solutions.get("dropout", 40000)
+                res = net.forward(
+                    input, key=jr.PRNGKey(np.random.randint(0, 2**31))
+                )
+                for _ in range(iters):
```
BSnelling, line 230: This iteration does slow the test down a fair bit but, since the dropouts are probabilistic, this is the recommended way to test.

FFroehlich (approved):

PR #3177, opened 2026-06-07
Fix noise formula substitution in flattening of PEtab problem

m-philipps:
- Fix a bug where flattening a PEtab problem would not replace the old observable id in the noise formula.
- Update the error message upon import of a non-flattened PEtab problem, as discussed [here](https://github.com/PEtab-dev/libpetab-python/issues/481)


dweindl (approved):
👍

PR #3179, opened 2026-06-08
Add equinox module import support for petab sciml problems

BSnelling:


.github/workflows/test_python_cplusplus.yml
```diff
@@ -213,7 +213,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [ "3.11" ]
+        python-version: [ "3.12" ]
```
BSnelling, line 216: The reasoning for this is that petab-sciml v0.1 will most likely specify python >=3.12 due to numpy requirements: https://github.com/PEtab-dev/petab_sciml/pull/83  

Is there a specific requirement to continue 3.11 support in AMICI? From another dependency for example?

dweindl: > Is there a specific requirement to continue 3.11 support in AMICI? From another dependency for example?

No, there isn't. It's time to update amici's python requirement, I'll do that on a separate PR.

BSnelling: Thanks @dweindl!

python/sdist/amici/exporters/jax/nn.py
```diff
@@ -391,6 +392,12 @@ def generate_equinox(
     # TODO: move to top level import and replace forward type definitions
     from petab_sciml import Layer
 
+    filename.parent.mkdir(parents=True, exist_ok=True)
+
+    if isinstance(nn_model, Path):
+        copyfile(nn_model, filename)
+        return
+
```
FFroehlich, line 404: add Path to type annotation and add processing when nn_model is string? Do we want to refuse overwriting unless something like `force=True` is specified?

python/tests/test_sciml.py
```diff
@@ -789,3 +818,40 @@ def test_valid_hybridization_no_error(self, mock_de_model):
...
+
+            assert isinstance(hybridization["net1"]["model"], Path)
+
+            generate_equinox(
+                hybridization["net1"]["model"],
+                pi.output_dir / "net1.py",
+                hybridization["net1"]["frozen_layers"],
+            )
+
+            assert os.path.isfile(pi.output_dir / "net1.py")
```
FFroehlich, line 866: would be nice to add some regression tests against model outputs here to also make sure the model can be called

FFroehlich (changes requested):

python/sdist/amici/exporters/jax/nn.py
BSnelling: Added string handling and annotation. 

On the overwriting, I think it is sufficient that the output directory is specified as a required arg.  If we added a `force` keyword we'd run into a "pass-through keyword" trap.  We have the `force_import` keyword in `PetabImporter.import_module`, which calls `SbmlImporter` and `ODEExporter` and finally this function `generate_equinox`. The kwarg would just get funnelled through all those classes.

FFroehlich (approved):

PR #3185, opened 2026-06-16
Add a JAX regression test workflow to track changes over time

BSnelling:
The workflow runs several representative test cases using JAX implementations and records performance metrics:
- llh
- number of steps
- wall time

The workflow is scheduled to run weekly and the metrics are uploaded in json format as a workflow artefact. The intention is that these metrics will highlight if we see performance regressions in AMICI due to either src changes here or dependency updates. 

Example metrics json file: https://github.com/AMICI-dev/AMICI/actions/runs/27679856537/artifacts/7691644836


.github/workflows/test_jax_performance.yml
```diff
@@ -0,0 +1,60 @@
+name: JAX Performance Regression
+
+on:
+  schedule:
```
FFroehlich, line 7: Lets also run this on PRs otherwise this can easily break e.g. when jax api version changes

FFroehlich (approved):
👍  I would be very interested in seeing results for bigger models and across jax versions, but let’s keep that for separate PRs

PR #3188, opened 2026-06-18
Update references

dweindl:


FFroehlich:
add https://arxiv.org/abs/2608.05777 and https://www.biorxiv.org/content/10.64898/2026.07.30.741759v1.abstract

FFroehlich (approved):

PR #3196, opened 2026-07-04
Fix JAX PEtab backend: discontinuities, PEtab v1 import, event assignments

FFroehlich:
## Summary

A cluster of related fixes to the JAX simulation/PEtab backend, found while investigating how explicit (time-triggered) discontinuities are handled:

- **Precise discontinuity stepping**: clip the diffrax step-size controller onto known (explicit) discontinuity times (`ClipStepSizeController(controller, jump_ts=...)`) instead of relying solely on root-finding. Verified 10 orders of magnitude accuracy improvement and zero rejected steps on a synthetic explicit-discontinuity model.
- **Explicit heavisides as tracked variables**: time-triggered Heaviside placeholders are now generated as tracked variables (like implicit/state-dependent ones) instead of being inlined as symbolic `jnp.select` functions of time. Keeps the vector field smooth within each integration segment.
- **PEtab v1 → v2 conversion**: `PetabImporter` now upgrades an in-memory `petab.v1.Problem` to v2 (via `to_files_generic` + `petab.v2.Problem.from_yaml`, which auto-upgrades) instead of raising `NotImplementedError`. This lets the JAX backend (which requires v2) consume v1 PEtab problems, e.g. from the PEtab benchmark collection.
- **Noise/observable placeholder bug**: JAX gave wrong sigmas (and thus wrong chi2/log-likelihood) for any model using PEtab's standard `noiseParameter{n}_{observableId}` / `observableParameter{n}_{observableId}` naming convention. The anchored classification regex only matched the bare `noiseParameter{n}` form, so these placeholders leaked into ordinary model parameters instead of being resolved via the per-measurement override arrays — every measurement ended up using the first observable's noise value. Fixed by porting the substitution already used by the legacy PEtab v1 JAX importer (`v1/_sbml_import.py`) into the newer `PetabImporter`, JAX-gated, operating only on an ephemeral importer-internal copy (never touching the PEtab problem itself or the SUNDIALS path). Verified against the PEtab benchmark model Boehm_JProteomeRes2014: chi2/llh now match SUNDIALS exactly (was off by ~7.4 in llh).
- **Event assignments silently dropped**: `_apply_event_assignments` processed roots in fixed windows of two, assuming every event contributes a Heaviside +/− root pair. A plain SBML `<event>` with an assignment (no associated Heaviside) contributes a single root, so whenever a model's total root count was odd, the last root (in declaration order) was silently dropped — no error, no warning, no state update. Fixed by iterating over every root individually. Investigated and confirmed this doesn't regress the paired-Heaviside mechanism, including when Heaviside pairs and plain events are mixed in the same model (explicit events always sort before Heaviside pairs internally, so a dropped tail root there is always a Heaviside root with a zero delta).

## Test plan

- [x] `python/tests/test_jax.py`: existing discontinuity tests pass; added `test_explicit_discontinuity` and `test_event_assignments_odd_root_count` (verified the latter fails without the fix and passes with it)
- [x] `python/tests/petab_/test_petab_v2.py`: added `test_petab_importer_upgrades_v1_problem` and `test_jax_matches_sundials_with_per_observable_noise_parameters` (JAX vs SUNDIALS llh agreement)
- [x] Verified SUNDIALS (v1 and v2-via-conversion) llh unaffected by the JAX-gated noise-parameter fix, on Boehm_JProteomeRes2014 and Fujita_SciSignal2010

## Notes for reviewers

Two related pre-existing bugs were found during this investigation but are **out of scope** for this PR (tracked separately):
- `_known_discs(self, p)` crashes with `NameError` when an explicit trigger time depends on a model expression (`w`) rather than pure parameters (e.g. `Fujita_SciSignal2010`'s EGF pulse timing) — predates this change (PR #3165).


FFroehlich:
known_discs fix is here https://github.com/AMICI-dev/AMICI/pull/3195

python/tests/test_jax.py
```diff
@@ -465,3 +465,171 @@ def test_time_dependent_discontinuity_equilibration(tmp_path):
...
+            diffrax.DirectAdjoint(),
+            diffrax.ODETerm(model._xdot),
+            model._root_cond_fns(),
+            model._root_cond_fn,
+            model._delta_x,
+            model._known_discs(p),
+            model.observable_ids,
+        )
+
+        expected = jnp.clip(ts - 5.0, min=0.0)
```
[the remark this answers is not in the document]
FFroehlich: nonsense, see https://docs.jax.dev/en/latest/_autosummary/jax.numpy.clip.html

python/tests/test_jax.py
```diff
@@ -465,3 +465,171 @@ def test_time_dependent_discontinuity_equilibration(tmp_path):
...
+        assert_allclose(ys.squeeze(), expected, atol=1e-6, rtol=1e-6)
+        # clipping onto the known discontinuity avoids rejected steps
+        assert stats["num_rejected_steps"] == 0
+
+    except NotImplementedError as err:
+        if "The JAX backend does not support" in str(err):
```
dweindl: It's supposed to be supported - do we need these checks?

dweindl (approved):

python/sdist/amici/importers/petab/_petab_importer.py
```diff
@@ -277,12 +330,17 @@ def _upgrade_or_copy_if_needed(
...
-            "'petab_problem' must be a `petab.v2.Problem`. "
-            "`petab.v1.Problem` is not directly supported, but "
-            "file-based PEtab v1 problems can be upgraded via "
-            "`petab.v2.Problem.from_yaml(petab_v1_yaml_file)`."
-        )
+        # Upgrade an in-memory PEtab v1 problem to PEtab v2: serialize it to a
+        # temporary PEtab v1 problem on disk and let petab auto-upgrade it
+        # (``petab.v2.Problem.from_yaml`` upgrades v1 YAML files via
+        # ``petab1to2``). The resulting problem holds the model and tables in
+        # memory, so the temporary directory can be removed afterwards.
```
BSnelling: I think this comment (or a more concise version of it) should be moved into the docstring of this function.

python/sdist/amici/sim/jax/_simulation.py
```diff
@@ -229,6 +237,14 @@ def solve(
     :return:
         solution+heaviside variables at time points ts and statistics
     """
+    # force the solver to step exactly onto known (explicit) discontinuities so
+    # the discontinuous vector field is integrated accurately instead of relying
+    # solely on root finding
```
BSnelling: Remove the comment? The code is self explanatory.

python/sdist/amici/importers/petab/_petab_importer.py
```diff
@@ -55,6 +55,57 @@
     v2.C.NOISE_PARAMETERS,
 ]
 
+#: Matches PEtab's observable-suffixed observable/noise placeholder
+#: parameters, e.g. ``noiseParameter1_pSTAT5A_rel``. Group 1 is the
+#: generic, index-based placeholder name (``noiseParameter1``).
+_JAX_PLACEHOLDER_PARAMETER_RE = re.compile(
+    r"((?:observable|noise)Parameter\d+)_\w+"
+)
```
BSnelling: Is this regex pattern (`observableParameter${n}_${observableId}`) required by the PEtab v2 specification? Some test cases have placeholders which don't follow that syntax (e.g. https://github.com/PEtab-dev/petab_test_suite/blob/main/petabtests/cases/v2.0.0/sbml/0021/_observables.tsv).

python/tests/test_jax.py
FFroehlich: Good catch — that guard has since been removed (it predated this reply; the JAX backend does now support this). The whole try/except skip pattern in this file was dropped in a later commit, so this concern no longer applies.

python/sdist/amici/importers/petab/_petab_importer.py
FFroehlich: Done — moved into the method's docstring.

python/sdist/amici/sim/jax/_simulation.py
FFroehlich: Removed (both occurrences, in `eq` and `solve`).

python/sdist/amici/importers/petab/_petab_importer.py
FFroehlich: Good catch, no — that convention isn't required. Confirmed against `petab.v2.lint.CheckOverridesMatchPlaceholders`: it matches placeholders to measurement overrides purely by count and position (`len(observable.noise_placeholders)` vs `len(measurement.noise_parameters)`), not by name. Your test case (0021) uses `obs_a_noise_scale` and would have silently broken with the old regex-based approach.

Fixed in 41ec966: dropped the regex entirely and now use `Observable.observable_placeholders`/`noise_placeholders` directly (substituting each declared placeholder, by position, with the corresponding generic `observableParameter{i}`/`noiseParameter{i}` symbol). Added a regression test (`test_jax_matches_sundials_with_arbitrarily_named_noise_placeholders`) using exactly this naming pattern, verified JAX/SUNDIALS agreement.

(The legacy PEtab v1 importer's regex at `v1/_sbml_import.py` is unaffected/still correct — v1 doesn't have placeholder-declaration columns, so the suffixed convention is the actual v1 spec requirement there, not a heuristic.)

FFroehlich:
still a couple of benchmark failures, let's adress them in a separate PR

BSnelling (approved):

PR #3197, opened 2026-07-04
Avoid JAX recompilation from fresh steady_state_event closures

FFroehlich:
## Summary
- `diffrax.steady_state_event()` returns a new closure on every call, and Python compares closures by identity, not value. Since it's passed as a static argument into the `eqx.filter_jit`-compiled `JAXModel.simulate_condition` / `JAXModel.preequilibrate_condition`, constructing a fresh instance with identical settings on every call (e.g. once per iteration of an optimization loop) silently defeats the JIT cache and forces a full recompilation, even though only numeric parameters changed.
- Adds `SteadyStateEvent`, an `equinox.Module` wrapping the same `rtol`/`atol` settings with value-based `__eq__`/`__hash__`, and uses it as the default in `run_simulations`/`petab_simulate` so equivalent instances share the compiled executable regardless of how many times they're constructed.
- Adds a regression test using `equinox.debug.assert_max_traces`/`get_num_traces` that verifies `simulate_condition` and `preequilibrate_condition` are traced exactly once across repeated calls with only numeric differences, even when solver/controller/root-finder/event objects are freshly reconstructed each call.

## Test plan
- [x] `python/tests/test_jax.py::test_steady_state_event_no_recompile_across_conditions` passes with the fix
- [x] Same test fails against the pre-fix code (import error / would show >1 trace with a bare `diffrax.steady_state_event()`)
- [x] Full non-PySB-dependent subset of `python/tests/test_jax.py` passes

### DIFF
--- a/python/tests/test_jax.py
+++ b/python/tests/test_jax.py
@@ -337,6 +337,131 @@ def test_serialisation(lotka_volterra):  # noqa: F811
 
 
 @skip_on_valgrind
+def test_steady_state_event_no_recompile_across_conditions(
+    tmp_path, monkeypatch
+):
+    """Simulating different conditions/parameters must not force JAX to
+    recompile, as long as only numeric inputs (e.g. parameters) change.
+
+    Regression test: ``diffrax.steady_state_event()`` returns a fresh
+    closure on every call, which Python compares by identity rather than
+    value. Since it is passed as a static argument to the
+    ``eqx.filter_jit``-compiled ``JAXModel.simulate_condition`` and
+    ``JAXModel.preequilibrate_condition``, constructing a fresh instance on
+    every call (e.g. once per condition, or once per optimizer iteration)
+    used to defeat the JIT cache and force a full recompilation every time,
+    even though only numeric parameters differed.
+    :class:`amici.sim.jax.petab.SteadyStateEvent` fixes this by giving the
+    event value-based equality/hashing, so this test also reconstructs the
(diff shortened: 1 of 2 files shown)


dweindl (approved):

python/tests/test_jax.py
```diff
@@ -336,6 +336,134 @@ def test_serialisation(lotka_volterra):  # noqa: F811
...
+    value. Since it is passed as a static argument to the
+    ``eqx.filter_jit``-compiled ``JAXModel.simulate_condition`` and
+    ``JAXModel.preequilibrate_condition``, constructing a fresh instance on
+    every call (e.g. once per condition, or once per optimizer iteration)
+    used to defeat the JIT cache and force a full recompilation every time,
+    even though only numeric parameters differed.
+    :class:`amici.sim.jax.petab.SteadyStateEvent` fixes this by giving the
+    event value-based equality/hashing, so this test also reconstructs the
+    solver/controller/root finder from scratch on every call to make sure
+    those (which already hash by value) don't regress either.
```
BSnelling: I'd say the explanation here is probably redundant. The reasoning is explained clearly in the class.

BSnelling (approved):
LGTM!

To fix the sciml workflow I think you will have to pin the version of libpetab to an earlier one, until this PR is done: https://github.com/AMICI-dev/AMICI/pull/3194.

PR #3204, opened 2026-07-08
Fix JAXProblem.save()/load() round-trip against current petab.v2 API

FFroehlich:
## Summary

- `JAXProblem.save()` (`python/sdist/amici/sim/jax/petab.py`) called `petab.v2.Problem.to_files()` with a keyword-argument signature (`prefix_path=`, `model_file=`, `condition_file=`, ...) that no longer exists — the current `to_files(self, base_path)` relies on each table/model already having `rel_path`/`base_path` set (normally done by `Problem.from_yaml`). This raised `TypeError: Problem.to_files() got an unexpected keyword argument 'prefix_path'`. Fixed by setting `config.filepath = "problem.yaml"` (so the written YAML has the name `JAXProblem.load()` expects) and calling `to_files(base_path=directory)`. `load()` needed no change.
- Getting the regression test far enough to hit that bug required fixing a prerequisite issue in `import_petab_problem(..., jax=True)` (`python/sdist/amici/importers/petab/v1/_petab_import.py`): it passed the original PEtab v1 problem straight into `JAXProblem()`, which rejects v1 problems outright. Now upgrades v1→v2 via `to_files_generic`/`from_yaml` first, mirroring the conversion already used elsewhere in the codebase.
- Removed the now-unneeded `try/except (TypeError, NotImplementedError): pytest.skip(...)` guard from `test_serialisation` in `python/tests/test_jax.py` so it actually asserts the save/load round trip instead of silently skipping. Left the equivalent guard on `test_preequilibration_failure` alone — that covers a separate, already-known bug in the preequilibration-condition heuristic, out of scope here.

## Test plan

- [x] `pytest python/tests/test_jax.py::test_serialisation` passes (round-trips perturbed parameters through `save()`/`load()` and asserts `assert_allclose`).
- [x] Confirmed the two other pre-existing failures in `test_jax.py` (`test_conversion`, `test_dimerization`) are unrelated environment issues (missing BioNetGen binary), and `test_preequilibration_failure` fails on a separate, already-tracked bug — neither is a regression from this change.

### DIFF
--- a/python/sdist/amici/importers/petab/v1/_petab_import.py
+++ b/python/sdist/amici/importers/petab/v1/_petab_import.py
@@ -255,6 +255,10 @@ def import_petab_problem(
     )
 
     if jax:
+        import tempfile
+
+        import petab.v2 as petabv2
+
         from amici.sim.jax import JAXProblem
 
         model = model_module.Model()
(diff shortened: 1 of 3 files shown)


BSnelling (approved):

FFroehlich:
CI failures here are expected due to v1->v2 upgrade and adressed in #3196

dweindl (approved):

PR #3205, opened 2026-07-08
Fix condition-name-based preequilibration detection in JAX PEtab backend

FFroehlich:
## Summary

`JAXProblem.run_preequilibrations` (and several sibling call sites in `python/sdist/amici/sim/jax/petab.py`) identified preequilibration periods/conditions with a substring check (`"preequilibration" in condition_id`). This only worked for AMICI's own synthetic condition names (e.g. `_petab_preequilibration_on`/`_petab_preequilibration_off`) and silently produced an empty or misaligned condition list for PEtab problems that use arbitrary, real condition names for preequilibration — e.g. `"no_interaction"` in the `lotka_volterra` test fixture. This crashed downstream with `ValueError: Need at least one array to stack.` or a `vmap` shape mismatch.

Per the [PEtab v2 spec](https://petab.readthedocs.io/en/latest/v2/documentation_data_format.html), only `time == -inf` signifies preequilibration (matching `petab.v2.ExperimentPeriod.is_preequilibration`); ordinary periods can have any numeric `time`, including negative finite values, so a `time < 0.0` check (an earlier draft of this fix) isn't precise enough either.

- Replaced every name/inequality-based check with `period.is_preequilibration` (or the `TIME_PREEQUILIBRATION` constant where operating on a raw DataFrame column in `get_simulation_conditions_v2`).
- Made the derived condition lists (`preequilibration_conditions`, `dynamic_conditions`) one entry per experiment, aligned with `experiments`/`p_array`, instead of a deduplicated global set — required since `_prepare_experiments` later does `zip(conditions, p_array)`.
- Fixed two additional bugs found while getting the fixture end-to-end, which blocked this code path from producing correct results once reachable:
  - `JAXProblem._get_parameter_mappings` built its per-condition `targets_map` with a malformed (flattened, non-nested) dict comprehension that discarded all but the last parameter override per condition.
  - `import_petab_problem(..., jax=True)` for PEtab v1 problems constructed `JAXProblem` directly from the v1 problem, which `JAXProblem` rejects; now upgrades to v2 first, mirroring the conversion already used by `PetabImporter`.
- Restored `test_preequilibration_failure` in `test_jax.py` without the `try/except (TypeError, NotImplementedError): pytest.skip(...)` guard that had been masking the crash — it now passes for real.

## Test plan

- `pytest python/tests/test_jax.py::test_preequilibration_failure` passes.
- `pytest python/tests/test_jax.py` — same pre-existing/unrelated failures as before this change (missing BioNetGen binary for `test_conversion`/`test_dimerization`; a separate, already-tracked `petab.v2 Problem.to_files()` API mismatch in `JAXProblem.save()` for `test_serialisation`).
- Manually verified against the `lotka_volterra` fixture through both repro routes: direct `amici.importers.petab.v1.import_petab_problem(..., jax=True)`, and PEtab v1→v2→`PetabImporter`.

### DIFF
--- a/python/sdist/amici/importers/petab/v1/_petab_import.py
+++ b/python/sdist/amici/importers/petab/v1/_petab_import.py
@@ -255,6 +255,10 @@ def import_petab_problem(
     )
 
     if jax:
+        import tempfile
+
+        import petab.v2 as petabv2
+
         from amici.sim.jax import JAXProblem
 
         model = model_module.Model()
@@ -263,9 +267,17 @@ def import_petab_problem(
             f"Successfully loaded jax model {model_name} from {output_dir}."
         )
 
+        # JAXProblem requires a PEtab v2 problem; upgrade the v1 problem by
+        # serializing it to a temporary PEtab v1 problem on disk and letting
+        # petab auto-upgrade it (``petab.v2.Problem.from_yaml`` upgrades v1
+        # YAML files via ``petab1to2``).
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            yaml_path = petab_problem.to_files_generic(prefix_path=tmp_dir)
+            petab_problem_v2 = petabv2.Problem.from_yaml(yaml_path)
+
         # Create and return JAXProblem
         logger.info(f"Successfully created JAXProblem for {model_name}.")
-        return JAXProblem(model, petab_problem)
+        return JAXProblem(model, petab_problem_v2)
 
     model = model_module.get_model()
     check_model(amici_model=model, petab_problem=petab_problem)
--- a/python/sdist/amici/sim/jax/petab.py
+++ b/python/sdist/amici/sim/jax/petab.py
@@ -285,12 +285,6 @@ class JAXProblem(eqx.Module):
                     )
 
         for _, simulation_condition in simulation_conditions.iterrows():
-            if (
-                "preequilibration"
-                in simulation_condition[petabv2.C.CONDITION_ID]
-            ):
-                continue
-
             query = " & ".join(
                 [
                     f"{k} == '{v}'" if isinstance(v, str) else f"{k} == {v}"
@@ -527,9 +521,9 @@ class JAXProblem(eqx.Module):
                 ch.target_id: jnp.asarray(
                     ch.target_value, dtype=self.model.parameters.dtype
                 )
+                for ch in c.changes
             }
             for c in self._petab_problem.conditions
-            for ch in c.changes
         }
 
         hybrid_map = {}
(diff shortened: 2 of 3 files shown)


python/sdist/amici/sim/jax/petab.py
```diff
@@ -95,6 +95,36 @@ def jax_unscale(
...
+        The first condition id of the matching period.
+    :raises ValueError:
+        If ``exp`` has no matching period with a non-empty
+        ``condition_ids``.
+    """
+    for period in exp.periods:
+        if period.is_preequilibration != is_preequilibration:
+            continue
+        if period.condition_ids:
+            return period.condition_ids[0]
```
dweindl: The assumption is that there is at most one condition ID per period, right? Add an assertion?

python/sdist/amici/sim/jax/petab.py
```diff
@@ -1987,12 +2005,14 @@ def get_simulation_conditions_v2(petab_problem) -> pd.DataFrame:
...
-    exps = {}
-    for exp_id in experiment_df[petabv2.C.EXPERIMENT_ID].unique():
-        exps[exp_id] = experiment_df[
-            experiment_df[petabv2.C.EXPERIMENT_ID] == exp_id
-        ][petabv2.C.CONDITION_ID].unique()
 
+    # drop preequilibration periods, identified by `time == -inf` (per the
+    # PEtab v2 spec, this is the only value that signifies preequilibration)
+    # rather than by condition name, since preequilibration conditions may be
+    # named arbitrarily.
```
```suggestion
    # drop preequilibration periods
```
bla

python/sdist/amici/importers/petab/v1/_petab_import.py
```diff
@@ -263,9 +267,17 @@ def import_petab_problem(
             f"Successfully loaded jax model {model_name} from {output_dir}."
         )
 
+        # JAXProblem requires a PEtab v2 problem; upgrade the v1 problem by
+        # serializing it to a temporary PEtab v1 problem on disk and letting
+        # petab auto-upgrade it (``petab.v2.Problem.from_yaml`` upgrades v1
+        # YAML files via ``petab1to2``).
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            yaml_path = petab_problem.to_files_generic(prefix_path=tmp_dir)
+            petab_problem_v2 = petabv2.Problem.from_yaml(yaml_path)
```
dweindl, line 276: If that works well, we can also implement that in libpetab directly. I had some reason for not doing that initially, but I can't recall.

dweindl (approved):

python/sdist/amici/sim/jax/petab.py
FFroehlich: good catch, fixed to tuple of conditions

python/sdist/amici/importers/petab/v1/_petab_import.py
FFroehlich: so far all conversions have been pretty smooth.

PR #3210, opened 2026-07-10
Fix JAX benchmark failures at the root cause (unblock Benchmark Collection JAX)

FFroehlich:
Fixes the red **Benchmark Collection JAX** required check by resolving the underlying failures at their root cause, rather than skipping. (Supersedes the skip-only #3209; incorporates the event fix from #3211.)

## Background

The check went red on `main` after #3204 added a PEtab v1→v2 auto-upgrade inside `import_petab_problem(..., jax=True)`, which removed the `"JAXProblem does not support PEtab v1 problems"` error that [`test_petab_benchmark_jax.py`](tests/benchmark_models/test_petab_benchmark_jax.py) caught in order to `pytest.skip`. That un-skipped the JAX benchmark tests and surfaced **24 pre-existing failures**.

## Root causes fixed

**1. Measurement/override parsing** — `amici/sim/jax/petab.py` (`_get_measurements`)
- Fixed (non-estimated) `observableParameters`/`noiseParameters` overrides were never substituted: the upgraded v2 `parameter_df` stores `estimate` as the string `"false"`/`"true"` (truthy either way), so the `not parameter_df.loc[...]` guard always failed and the parameter name leaked into `astype(float)` (`could not convert string to float`). Now resolved via the parsed parameter objects (real bool `estimate`, numeric `nominal_value`).
- Empty override cells are treated as absent (padded to the neutral `1.0`) instead of failing the float cast.
- `iy_trafos` is built **per measurement row** instead of per observable; the length-`n_obs` array was later split dyn/post-eq as if per-measurement → `index can't contain negative values`, plus a latent log-normal `res`/`chi2` mis-assignment.

**2. Parameter scaling in the benchmark test** — `test_petab_benchmark_jax.py`
- The PEtab v2 JAX backend operates on **linear** parameters, but the gradient check injected **scaled** (log10) values, so JAX evaluated at the wrong physical point (stiff models → NaN, others → wrong LLH/SLLH). The point is now unscaled before injection, and the JAX (linear) gradient is converted to the estimation scale via the chain rule for the SLLH comparison.
- Benign v1→v2 upgrade warnings (`log10-normal`→`log-normal`; dropped initialisation priors) that pytest turned into errors are now filtered.

**3. Preequilibration + parameter-dependent events** — `amici/sim/jax/model.py` (`_handle_t0_event`, via #3211)
- The dynamic phase reused the preequilibration Heaviside/event state verbatim, which is wrong when an event trigger depends on a parameter that differs between the preeq and dynamic periods (a stimulus whose onset time is inactive during preeq). The event state at the dynamic `t0` is now re-evaluated from the triggers under the dynamic-period parameters (seeded from the preeq state so already-active events are not re-fired).

**4. petab pinning** — `.github/workflows/test_benchmark_collection_models.yml`
- Pin petab for the benchmark-JAX job (matching the v1 testsuite job); the failure set was non-deterministic against `petab@main`.

## Documented skips (not correctness failures)

- **log10-normal noise (6):** genuine PEtab v2 limitation — v2 has no log10-normal, the upgrade rewrites it to log-normal (a different likelihood; the noise is not rescaled by ln(10)), so it can't match the log10-normal SUNDIALS reference. Mirrors the existing SUNDIALS-v2 test.
- **Weber_BMC2015 (1) — runtime only:** fixed and correct, but uses `max_steps=4e7` under reverse-mode AD (empirically >8 min; dominates the job). Its preeq+event path is covered by the fast Brannmark/Isensee, which stay enabled.

## Result

Full JAX benchmark: **19 passed / 11 skipped / 0 failed**. Of the 24 originally-failing models: **17 now pass**, 6 log10-normal skip, 1 (Weber) runtime skip. All stiff models pass with no per-model tolerance changes.

### DIFF
--- a/.github/workflows/test_benchmark_collection_models.yml
+++ b/.github/workflows/test_benchmark_collection_models.yml
@@ -128,7 +128,7 @@ jobs:
 
     - name: Install test dependencies
       run: |
-        python3 -m pip uninstall -y petab && python3 -m pip install git+https://github.com/petab-dev/libpetab-python.git@main \
+        python3 -m pip uninstall -y petab && python3 -m pip install git+https://github.com/petab-dev/libpetab-python.git@1b8599dd1eb9bda74853255b4cc4baf75b7e4d63 \
         &&  python3 -m pip install -U sympy \
         &&  python3 -m pip install git+https://github.com/ICB-DCM/fiddy.git@amici100
 
--- a/python/sdist/amici/sim/jax/model.py
+++ b/python/sdist/amici/sim/jax/model.py
@@ -975,12 +975,19 @@ class JAXModel(eqx.Module):
         h_preeq: jt.Bool[jt.Array, "ne"],
         stats: dict,
     ):
-        y0 = y0_next.copy()
         rf0 = self.event_initial_values - 0.5
 
         if h_preeq.shape[0]:
-            # return immediately because preequilibration is equivalent to handling t0 event?
-            return y0, t0_next, h_preeq, stats
+            # Dynamic phase following preequilibration: carry the event state
+            # out of preequilibration, but re-evaluate the triggers at t0 under
+            # the dynamic-period parameters, which may differ from the
+            # preequilibration ones (e.g. a stimulus whose onset time is a
+            # condition-specific parameter that is inactive during
+            # preequilibration). Events already active after preequilibration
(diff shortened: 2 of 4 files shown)


FFroehlich:
@claude review once

BSnelling (approved):

PR #3212, opened 2026-07-11
perf(jax): cut JAX PEtab benchmark compile time via CSE + ImplicitAdjoint

FFroehlich:
## Summary

The JAX PEtab benchmark models flagged as long-running (Fiedler, Giordano, Zheng, Lucarelli, Weber) are **JIT/XLA-compile bound, not integration bound**: measured ODE integration is 0.2–0.4 s per model, while gradient *compilation* ranged from seconds to minutes. This PR removes that compile cost at the root cause.

## Root cause & main fix — CSE in the JAX code printer

The JAX code printer emitted every equation without common-subexpression elimination. Blocks with large repeated subterms — notably the analytic steady-state initial values in `_x0` — were written out many times (Fiedler's `_x0` repeated an ~11-term `sqrt(...)` polynomial 10×). JAX's autodiff then differentiates every copy, producing a huge reverse-mode graph whose XLA optimization is superlinearly slow (XLA even printed a "very slow compile" alarm).

Adding `sp.cse` to `AmiciJaxCodePrinter._get_sym_lines`, with the extracted temporaries interleaved with the (sequential) target assignments in dependency order via `toposort` (mirroring the SUNDIALS code printer), keeps the differentiated graph compact.

### Gradient-compile time — llh/sllh byte-identical before vs after

| Model | before | after |
|---|---|---|
| Fiedler_BMCSystBiol2016 | ~281 s | **~4 s** |
| Giordano_Nature2020 (110 piecewise) | ~51 s | **~14 s** |
| Zheng_PNAS2012 (preeq) | ~31 s | **~10 s** |
| Weber_BMC2015 | skipped (hours) | **~50 s → re-enabled** |

## Also in this PR

- **`ImplicitAdjoint` for the event-free steady-state solve** — differentiates the equilibration via the implicit function theorem (one linear solve) instead of backpropagating through the whole trajectory. Stacks with CSE on preequilibration models.
- **Fix `simulation_experiments` subsetting** — `run_simulations(problem, simulation_experiments=<subset>)` previously crashed with a vmap size mismatch because the pre-built per-experiment arrays were not subset alongside `p_array`. `petab_simulate` result-building is fixed for subsets too.
- **Remove the redundant forward-only warm-up** in `test_petab_benchmark_jax.py` — it compiled a second, unrelated executable for every gradient-checked model (different static `max_steps`/controller; forward-only vs. forward+backward) and never primed the timed call.
- **Unskip `Weber_BMC2015`** — was skipped for runtime; feasible now (~50 s compile).

## Verification

Regenerated and re-ran Fiedler, Giordano, Zheng, Brannmark, Isensee, and Weber with the CSE printer: forward `llh` matches the pre-change values exactly, gradients (`|sllh|`) are unchanged, and CSE is correct through SBML rationals, `Piecewise → jnp.select`, preequilibration, and event/root code (`toposort` is required — a naive "temporaries first" emission mis-orders `w` entries that later equations depend on). CI runs the full SUNDIALS-reference gradient checks.

> Note: this branch is currently 2 commits ahead of `main` — it also carries two recent JAX-backend commits (#3211 and a benchmark root-cause fix) that are not yet on `main`.

### DIFF
--- a/python/sdist/amici/sim/jax/petab.py
+++ b/python/sdist/amici/sim/jax/petab.py
@@ -1347,6 +1347,18 @@ class JAXProblem(eqx.Module):
         """
         return eqx.tree_at(lambda p: p.parameters, self, p)
 
+    def _experiment_indices(self, experiments) -> np.ndarray:
+        """Positions of ``experiments`` in the full experiment ordering.
+
+        All ``self._*`` per-experiment arrays are built in ``__init__`` keyed by
+        ``self._petab_problem.experiments`` order, so simulating a subset
+        (``simulation_experiments``) requires indexing them by these positions.
+        """
+        positions = {
+            exp.id: i for i, exp in enumerate(self._petab_problem.experiments)
+        }
+        return np.array([positions[exp.id] for exp in experiments], dtype=int)
+
     def _prepare_experiments(
         self,
         experiments: list[petabv2.Experiment],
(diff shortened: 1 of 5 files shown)


python/sdist/amici/sim/jax/petab.py
```diff
@@ -289,6 +289,18 @@ def _get_measurements(
         measurements = dict()
         petab_indices = dict()
 
+        # Nominal (linear) values of fixed (non-estimated) parameters, used to
+        # resolve observable/noise parameter overrides that reference them.
+        # ``estimate`` is taken from the parsed parameter objects, where it is a
+        # real bool; in the raw ``parameter_df`` it is the string
+        # ``"false"``/``"true"`` and therefore truthy either way.
```
```suggestion
```

bla

dweindl (approved):
The speedup sounds good.

BSnelling (approved):

Issue #3214, opened 2026-07-25
x0 code generation constant-folds an initialAssignment-defined initial condition, losing parameter dependence (jakstat_adjoint / init_STAT)


wshlavacek:
### Summary

For the `jakstat_adjoint` test model, generated `x0.cpp` constant-folds a species initial condition that the SBML defines by an `initialAssignment` referencing a model parameter. The parameter dependence is lost, so both the analytic forward sensitivity and AMICI's own finite difference report zero for that parameter.

AMICI 1.0.1; reproduced with the model built by `amici.testing.models.import_model_jakstat()`.

### What the SBML says

```xml
<species id="STAT" compartment="default_compartment" .../>
<parameter id="init_STAT" constant="true"/>
...
<initialAssignment symbol="STAT"><ci> init_STAT </ci></initialAssignment>
```

So `STAT(0) = init_STAT` and `d STAT(0) / d init_STAT = 1`.

### What is generated

`x0.cpp`:

```cpp
void x0_model_jakstat_adjoint_py(realtype *x0, const realtype t, const realtype *p, const realtype *k){
    x0[0] = 1.0;
}
```

The literal `1.0` is `init_STAT`'s nominal value (`10^0`), folded in at code generation. `p` is unused for this state.

### Consequences

`init_STAT` is still reported by `model.get_free_parameter_ids()`, so it looks like a differentiable parameter, but:

- **Analytic forward sensitivity** `d STAT / d init_STAT` is identically zero at every output time.
- **Finite difference is also zero** — perturbing the parameter does not move `x0`, so the model is internally consistent but no longer matches the SBML it came from:

  ```
  parameter   nominal    x0     x0 (perturbed +1e-5)   max|dx/dp|
  p1          3.98107    1.0    1.0                    9.25e-2     <- control, behaves correctly
  init_STAT   1.0        1.0    1.0                    0.0
  ```

  (`p1` confirms the check itself is sound: its finite difference reproduces the analytic sensitivity peak to six figures.)

I suspect the trigger is that `init_STAT` *itself* carries an `initialAssignment` (`init_STAT = 10^0`, as antimony emits every parameter assignment), and something upstream of code generation concludes it is a fixed derived quantity and substitutes its value — but it stays in the free-parameter list, so nothing signals the loss.

### Expected

[...]

PR #3216, opened 2026-07-28
Fix SBML import constant-folding parameters out of initial conditions

FFroehlich:
## Summary
Fixed a bug in SBML import where parameters with `initialAssignment` elements were being constant-folded out of initial conditions, causing their sensitivities to be silently reported as zero even though they remained in the free-parameter list.

## Key Changes
- **Modified `_make_initial()` in `sbml/__init__.py`**: Added logic to skip substitution of initial assignments for parameters that AMICI keeps as symbolic quantities (free/fixed parameters or expressions). This prevents constant-folding of parameter dependencies into initial conditions.
  - Checks if a variable is a parameter that should remain symbolic rather than being folded
  - Handles both direct parameters and those reached through chains of initial assignments
  - Excludes parameters that become states (rate-rule targets or converted to species)

- **Added regression tests in `test_sbml_import.py`**:
  - `test_initial_assignment_parameter_not_constant_folded()`: Tests the case where a parameter with a compound-expression initial assignment (e.g., `X0 = 6 * 1`) defines a species' initial condition
  - `test_chained_initial_assignment_parameter_not_constant_folded()`: Tests the transitive case where a parameter's initial assignment references another parameter with its own initial assignment

## Implementation Details
The fix recognizes that during parameter classification in `_process_parameters`, certain parameters are intentionally kept as symbolic quantities rather than being folded to numeric values. When processing initial conditions, the code now skips substituting initial assignments for such parameters, preserving the symbolic dependence chain. This ensures that sensitivity analysis correctly reports non-zero derivatives with respect to these parameters.

The solution handles both direct cases (a parameter with a numeric initial assignment) and transitive cases (chains of parameter initial assignments), addressing issue #3214.

https://claude.ai/code/session_012xf5RqvErY7po9N9DNfpKQ

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ See also our [versioning policy](https://amici.readthedocs.io/en/latest/versioni
   (e.g., when using network file systems with not synchronized clocks).
   If you encounter this warning, and you are sure that the model was not
   modified since the last import, you can safely ignore it.
+* Fixed SBML import constant-folding a parameter out of the initial conditions
+  when that parameter both carried an `initialAssignment` with a numeric
+  right-hand side and defined a species' initial value. The parameter remained
+  in the free-parameter list, so its (initial-condition) sensitivity was
+  silently reported as zero (e.g. `init_STAT` in `jakstat_adjoint`) (#3214).
 
 ### v1.0.1 (2026-03-13)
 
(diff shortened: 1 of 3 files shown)


dweindl (approved):
Thanks, looks good to me.

PR #3218, opened 2026-08-04
Implement residual sensitivity aggregation for PEtab v2

FFroehlich:
## Summary

This PR implements aggregation of residual sensitivities for PEtab v2 simulations and improves handling of edge cases like non-Gaussian noise models and experiments without measurements.

## Key Changes

* **Implement residual sensitivity aggregation**: Added `_aggregate_sres()` method to compute sensitivities of concatenated residuals w.r.t. estimated PEtab parameters. The `PetabSimulationResult.sres` field now contains the aggregated residual sensitivities instead of `None`.

* **Extract parameter mapping logic**: Created `_get_plist_to_problem_par_ix()` helper method to map simulation result parameter indices to problem parameter indices. This handles both regular parameters and output parameter placeholders, and is now reused by both `_aggregate_s2llh()` and `_aggregate_sres()`.

* **Improve edge case handling**:
  - Experiments without measurements (no timepoints) are now correctly skipped when aggregating residuals and their sensitivities
  - Non-Gaussian noise models now gracefully return `None` for unavailable results instead of raising errors
  - Residual-only reporting mode (`RDataReporting.residuals`) is now properly supported
  - Added `_has_timepoints()` helper to distinguish between experiments without measurements and missing computed results

* **Enhanced documentation**: Updated docstrings for `PetabSimulationResult.sres` and `res()` property to clarify the structure and ordering of aggregated results.

## Implementation Details

* Residual sensitivities are aggregated by concatenating sensitivities from individual experiments in order, matching the row order of concatenated residuals
* Multiple model parameters mapping to the same problem parameter (e.g., output parameter placeholders) are correctly handled using `np.add.at()` to sum their contributions
* The implementation validates that all experiments have consistent availability of sensitivities before aggregation
* Least-squares identities (`FIM == sres.T @ sres` and `sllh == -res @ sres`) are verified to hold for parameter-independent noise models in the test suite

https://claude.ai/code/session_01F52fCqHwmLeAm1TDNUYCCc


python/sdist/amici/sim/sundials/petab/_v2.py
```diff
@@ -545,13 +558,29 @@ def res(self) -> np.ndarray | None:
         """
         Concatenated residuals.
 
+        The residuals of the individual experiments are concatenated in the
```
dweindl, line 562: Unrelated, but shall we change `.res` to `@property` for consistency with `.llh`? Not sure why I didn't do that.

python/sdist/amici/sim/sundials/petab/_v2.py
```diff
@@ -545,13 +558,29 @@ def res(self) -> np.ndarray | None:
...
+            :class:`numpy.ndarray`, or ``None`` if they are not available,
+            i.e., if any experiment failed to simulate or if residuals were
+            not computed at all. If no experiment has any measurement, an
+            empty array is returned.
         """
-        if any(rdata.res is None for rdata in self.rdatas):
-            return None
+        res = []
+        for rdata in self.rdatas:
+            # AMICI does not invalidate the residuals of failed simulations,
```
dweindl: Was there a reason for that, or is that something to fix?

dweindl (approved):
Thanks

python/tests/petab_/test_petab_v2.py
```diff
@@ -326,6 +326,202 @@ def test_petab_simulator_deepcopy_and_pickle():
...
+        SensitivityMethod,
+        SensitivityOrder,
+    )
+
+    problem = _residual_test_problem()
+    ps = PetabImporter(
+        problem,
+        module_name="test_petab_v2_residuals",
+        verbose=False,
+    ).create_simulator(force_import=False)
```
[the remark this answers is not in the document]
FFroehlich: Fair point — fixed in 6c3791a. The model module is now generated once in a module-scoped fixture with `force_import=True`, and each test builds its own simulator (and thus its own model and solver instance) from that importer. So there is no dependence on state left behind by earlier runs, and the model is still only compiled once for all three tests.

---
_Generated by [Claude Code](https://claude.ai/code)_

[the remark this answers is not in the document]
dweindl: It's still not fully independent - for that it should use a temporary model directory. This will still break with pytest-xdist or with concurrent pytest calls.

python/sdist/amici/sim/sundials/petab/_v2.py
FFroehlich: Done in f0caca2 — `res` is a property now, matching `llh`, and `res`/`sres` are symmetric.

Note that this is a (small) breaking change: `res()` shipped as a method in v1.0.0, so `result.res()` now raises `TypeError`. I added a breaking-changes bullet to the changelog for v1.0.2 and updated the one mention in the PEtab v2 example notebook.

---
_Generated by [Claude Code](https://claude.ai/code)_

python/sdist/amici/sim/sundials/petab/_v2.py
FFroehlich: Something to fix — it looks like an oversight rather than intent, so I fixed it in f0caca2.

`ReturnData::invalidate` NaN'ed `x`, `y`, `w`, `sx`, `sy` and (via `invalidate_llh`/`invalidate_sllh`) `llh`, `chi2`, `sllh`, `s2llh`, but left `res`, `sres` and `FIM` untouched. Those are initialized to 0.0 and only written for timepoints that were actually reached, so after a failure the residuals of the remaining timepoints stayed at 0.0 — a perfect fit — and the `FIM` held a partial sum. Reproduced on a PEtab v2 problem with `max_steps=1`: `llh` was NaN and `sllh`/`s2llh`/`sres` were `None`, but the aggregated residuals came back as `[-2.2, 1.14, 0, 0, 0, 0]`.

`invalidate` now NaNs `res`/`sres` from the failed timepoint on (both the residual and, if `sigma_res`, the error-residual block) and the whole `FIM`, since it is accumulated over all timepoints and thus has no valid partial result — same reasoning as for `llh`/`sllh`. New test in `test_rdata.py::test_invalidate_residuals_on_failure`, which also pins down that the timepoints before the failure stay valid.

I kept the status check in `PetabSimulationResult.res` regardless: returning a vector that is half NaN is less useful than `None`, and it keeps the aggregated results consistent with `sllh`/`s2llh`/`sres`, which are all `None` when any experiment failed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3219, opened 2026-08-04
Keep constant parameter chains as expressions; refresh pregenerated test models

FFroehlich:
## What

Refreshes the pregenerated test models in `models/` and the expected results in `tests/cpp/expected_results_py.h5`, and fixes two bugs that regenerating surfaced.

## Why the artifacts were stale

`models/` holds checked-in generated model code that the C++ tests build against, and `expected_results_py.h5` was generated from it. The regeneration target is not wired into the build:

```cmake
# add_dependencies(external_model_${MODEL} import_test_models)
```

So both artifacts stayed self-consistent with each other while the importer they were derived from moved on, and CI stayed green. Running the current importer against the committed expectations fails four cases:

```
FAILED model_calvetti-nosensi
FAILED model_jakstat_adjoint-sensiadjoint
FAILED model_jakstat_adjoint-sensiforward
FAILED model_jakstat_adjoint-sensiforwardlogparam
```

## Fixes

**Constant parameter chains no longer become states.** `_process_parameters` only admitted a parameter as an expression if its initial assignment resolved to free/fixed parameters alone. Once #3216 stopped constant-folding `p2`, `C2ss = V2ss / (p2 - 0.5 * R2ss)` had an unresolvable symbol and fell through to the species branch, which is meant for genuinely time-dependent initial assignments. `calvetti` grew two constant states (`nx_rdata` 6 → 8) as a result. Expressions may depend on other expressions — they are toposorted in `_process_rules` — so the candidate set is now resolved to a fixed point before any parameter is turned into a state.

**`res`, `sres` and `FIM` of failed simulations are invalidated.** Residuals of timepoints that were not reached stayed at their initial value of `0.0`, so a failed simulation looked like a perfect fit, and the `FIM` held a partial sum. Cherry-picked from `claude/petab-v2-residual-aggregation-vslsid` (`src/rdata.cpp` and its test only — the PEtab v2 `res`-property change stays on that branch), since the regenerated expectations depend on it.

## Expectation changes

| Change | Cause |
| --- | --- |
| `jakstat_adjoint` `x0[0]` 1.0 → 1.25893, non-zero `sx0` | #3216's own fix — `init_STAT` was folded out and its sensitivity silently reported as zero |
| `steadystate` failure cases: `res`/`FIM` → NaN | the `ReturnData::invalidate` fix above |
| `calvetti` `w` columns permuted (`C2ss`,`C3ss` 1,2 → 7,8) | `C2ss`/`C3ss` now reference `p2`/`p3` symbolically, so they sort later; `x` and `y` are bit-identical |
| `robertson` SPBCG `sx`/`sy` drift 1.8e-4 | iterative-solver noise — worth watching as a flake candidate |
| 139 solver-diagnostics counters | `numerrtestfails`, `order`, `numrhsevals` — noise |

## Verification

```
36 passed, 4 skipped   test_pregenerated_models.py + test_rdata.py
34 passed              test_sbml_import.py
```

The 4 skips are the `sensi2*` cases needing `_o2` models — unchanged from before.

## For reviewers

`models/` and `expected_results_py.h5` must stay in one commit — each half is only consistent with the other, so splitting them turns CI red. Worth reviewing the `calvetti` and `jakstat_adjoint` model diffs specifically; the rest of `models/` is the swig mtime-warning template from #3164 plus commit hashes.


dweindl (approved):
👍

PR #3220, opened 2026-08-05
Skip redundant HDF5 detection during model CMake configure

dweindl:
## Summary
- `AmiciConfig.cmake` currently calls `find_dependency(HDF5 COMPONENTS C HL CXX REQUIRED)`, which re-runs `FindHDF5.cmake`'s compiler-wrapper interrogation (spawning `h5cc`/`h5c++` subprocesses) on *every single* model configure, even though the result never changes for a given AMICI installation.
- On non-Config-mode HDF5 packaging (e.g. Debian/Ubuntu system HDF5), this consistently costs several hundred ms per model configure, on top of unavoidable CMake ABI/toolchain detection.
- This resolves the HDF5 library paths once, when AMICI core itself is built, and bakes them into `AmiciConfig.cmake` as plain imported targets (`hdf5::hdf5`, `hdf5::hdf5_cpp`, `hdf5::hdf5_hl`, `hdf5::hdf5_hl_cpp`) instead of re-deriving them for every model. Falls back to the full `find_dependency()` search if the cached paths no longer exist on disk (e.g. HDF5 was moved/reinstalled since AMICI was built).
- Goal: speed up importing/building many models in one go (test suite, benchmark collections, PEtab batch imports), where this per-model overhead adds up across dozens of independently-configured build directories — including under pytest-xdist, where models can't share a single build directory across parallel workers.

## Test plan
- [x] Rebuilt AMICI core from source (`pip install -e python/sdist --no-build-isolation`) and confirmed the regenerated `AmiciConfig.cmake` has the baked-in HDF5 paths substituted correctly
- [x] Measured 3 back-to-back CMake configures of fresh, never-before-seen model directories: ~1.0s each, down from ~3.4s cold before this change
- [x] Full model build → install → simulate with the rebuilt core: correct simulation results
- [x] Ran a subset of `python/tests/test_sbml_import.py` (model-import/build tests) against the rebuilt core: all pass

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,18 @@ elseif(AMICI_TRY_ENABLE_HDF5)
   find_package(HDF5 COMPONENTS C HL CXX)
 endif()
 
+if(HDF5_FOUND)
+  # Extract the plain library paths for hdf5/hdf5_cpp/hdf5_hl/hdf5_hl_cpp
+  # (dropping the transitive system libs like crypto/curl/z that
+  # HDF5_C_LIBRARIES/HDF5_CXX_LIBRARIES also list) so that AmiciConfig.cmake
+  # can bake them in as fixed imported targets, instead of every downstream
+  # model re-running FindHDF5.cmake's compiler-wrapper interrogation.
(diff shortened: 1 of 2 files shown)


FFroehlich (approved):
👍

PR #3226, opened 2026-08-07
Fix flaky Weber_BMC2015 gradient check

FFroehlich:
`test_benchmark_gradient[Weber_BMC2015-*-unscaled]` has been flip-flopping between passing and failing on `Benchmark Collection CPP (3.14, false)` -- two runs of the same tree (differing only in CHANGELOG.md, with identical dependency versions) produced one pass and one failure.

The forward and adjoint AMICI gradients agree to nine significant digits, so the gradients are fine; the finite differences are not. The failing direction is `a32`, where fiddy reports 860.83 against AMICI's 872.68 (1.36%, just over `rtol_check=1e-2`).

Root cause: fiddy's `Consistency` checker averages the estimates from every step size whose forward/backward/central differences agree with each other, and Weber has no step size window that suits all directions at once. The noise parameters (e.g. `std_yPKDpN25`) have such a large second derivative that forward and backward differences only come within 5% of the central one for relative steps <= 5e-5. At those steps the differences for parameters with a small nominal value (`a32` ~ 1e-4) are integrator-noise dominated -- the central difference is off by 0.8% at 1e-4 and 4.2% at 1e-5, while being off by only 2e-6 at 2e-1. Those noisy estimates are internally consistent (the bias comes from a shifted solver trajectory, not from per-evaluation jitter), so `Consistency` accepts and averages them in. Whether a given step size clears the internal agreement check depends on last-ulp details of the runner, which is what makes the outcome vary between runs.

Use central differences only for this problem, over the step size window in which every direction is converged. Measured worst-case relative deviation from the AMICI gradient over {forward, adjoint} x {scaled, unscaled} x {AMICI_EXTRACT_CSE=true, false} drops from 5.5e-3 to 1.9e-3, against an unchanged `rtol_check` of 1e-2.

`method_ids` becomes a `GradientCheckSettings` field, defaulting to the previous forward/backward/central triple, so all other problems are unaffected.


dweindl:
I am trying to address this directly in fiddy (https://github.com/ICB-DCM/fiddy/pull/77).

Waiting for https://github.com/AMICI-dev/AMICI/pull/3223 to collect more data, to confirm the failure mode as I cannot reproduce it locally.

PR #3228, opened 2026-08-07
Fix noise array reshaping when no parameters are estimated

BSnelling:
In a JAXProblem with no estimated scalar parameters and with numerical noise, the noise parameter array was getting mistakenly reshaped by broadcasting in the `jnp.where` call.  This fix skips the `jnp.where` call if `unscaled_parameters` in empty.


FFroehlich (approved):
👍

PR #3230, opened 2026-08-10
Resolve original condition ids in reinit

BSnelling:
Updating the function `_condition_reinit_target_value` to resolve conditions ids in the same way as `_eval_nn` does, while preserving gradient tracing. 

The context in which I noticed this bug was a JAXProblem built from a PetabImporter instance. The underlying petab problem contained conditions which modified a model state variable. The PetabImporter changes the names of conditions and condition targets as part of import which is why `_eval_nn` needs to resolve current condition ids to originals, and why `_condition_reinit_target_value` should do the same.

### DIFF
--- a/python/sdist/amici/sim/jax/petab.py
+++ b/python/sdist/amici/sim/jax/petab.py
@@ -1019,10 +1019,27 @@ class JAXProblem(eqx.Module):
             [jax_unscale(pval, scale) for pval, scale in zip(p, scales)]
         )
 
-    def _resolve_original_condition_id(self, condition_id: str) -> str:
-        """Map a converted condition ID back to its original unconverted ID."""
-        if self._unconverted_problem is None:
-            return condition_id
+    def _find_original_period(
+        self, condition_ids: tuple[str, ...]
+    ) -> (
+        tuple[
+            petabv2.Experiment,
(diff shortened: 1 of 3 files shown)


FFroehlich (approved):

FFroehlich:
@dweindl does this look good to you?

python/sdist/amici/sim/jax/petab.py
dweindl: Generally, the ordering of experiments is not supposed to be significant in PEtab and shouldn't be relied on. I think, here it's okay given that both problems are private. Maybe add an explicit comment here or to the respective attribute docstrings.

dweindl (approved):
Thanks, looks good to me.

I was just wondering whether it might not be easier to just store the mapping during import instead of reverse engineering it later, but fine as is.

python/sdist/amici/sim/jax/petab.py
BSnelling: Added to docstring.

BSnelling:
> I was just wondering whether it might not be easier to just store the mapping during import instead of reverse engineering it later

I tried this out but found it didn't simplify the code much (or speed it up). I think it's slightly easier to follow if all the resolution is done together.

PR #3232, opened 2026-08-17
Speed up parallel model import

FFroehlich:
Parallel model import (`AMICI_IMPORT_NPROCS` > 1) created a brand-new
`spawn` worker pool for every single parallelized operation. Since model
import performs dozens of them, and every worker of every pool had to
import sympy and amici from scratch, the process startup overhead often
exceeded the actual work -- for all but the largest models, parallel
import was *slower* than serial import.

* Create the worker pool lazily and reuse it for all operations, instead
  of creating and tearing one down per operation.
* Use the `forkserver` start method instead of `spawn`. Its workers are
  forked from a small, single-threaded server process that imports sympy
  and amici only once, which makes pool creation ~5x cheaper without
  reintroducing the deadlock potential of forking the (potentially
  multi-threaded) main process. `spawn` is kept only where `forkserver`
  is unavailable, i.e. on Windows.
* Process small matrices serially, where the inter-process communication
  overhead outweighs any speed-up.
* Compute the jacobian sparsity pattern from the symbols of each row
  instead of calling `Basic.has` for every (row, variable) pair, which
  re-traverses the full expression tree every time. This also speeds up
  serial import.
* Validate `AMICI_IMPORT_NPROCS` instead of failing with an opaque error
  further down, and detect unpicklable functions in `_parallel_applyfunc`
  before dispatching them. The previous `PicklingError` handler never
  triggered for the most common case (a lambda), which raises
  `AttributeError`.

Micro-benchmark (4 cores, 12 `smart_jacobian` calls, ~1.5 s to import
amici): 25.0 s -> 2.8 s. For a workload large enough for parallelization
to pay off, the speed-up over serial import improves from 2.3x to 4.2x
on 4 cores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,14 @@ See also our [versioning policy](https://amici.readthedocs.io/en/latest/versioni
 * PEtab SciML: implemented additional PyTorch-style layer types
   (`BatchNorm`, `InstanceNorm`, `AlphaDropout`, `Bilinear`) (#3176).
 * PEtab SciML: updated support to PEtab v2 (#3165).
+* Sped up parallel model import (`AMICI_IMPORT_NPROCS` > 1) considerably.
+  Previously, a new worker pool was created for every parallelized operation,
+  and every worker of every pool had to import sympy and amici from scratch.
+  Now, a single pool is created lazily and reused, and its workers are forked
+  from a `forkserver` process that performs these imports only once. For small
+  operations, where the inter-process communication overhead outweighs any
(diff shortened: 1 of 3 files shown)


FFroehlich:
Performance check:

Before:
> Time for task petab_import is 731.48s, which is within the reference time of 740.00s.

After:
> Time for task petab_import is 378.26s, which is within the reference time of 740.00s.

dweindl (approved):
Cool, thanks.

dweindl:
FYI: For the performance test runs (2 procs), this reduced model import from ~12min to ~5-8min. :+1:

PR #3234, opened 2026-08-18
Fix wrong branch for piecewise nested in another piecewise's condition

### LINKED ISSUE #3233: A piecewise inside another piecewise's condition returns the wrong branch
When one piecewise expression appears inside the condition of another, AMICI takes the wrong branch of the outer one and returns zero. The inner value is reported correctly, so only the outer expression is affected.

The script below defines four parameters by assignment rules and integrates a single state variable.

```python
import tempfile
import amici
import amici.sim.sundials as ss
import numpy as np

MML = 'xmlns="http://www.w3.org/1998/Math/MathML"'

SBML = f"""<?xml version="1.0" encoding="UTF-8"?>
<sbml xmlns="http://www.sbml.org/sbml/level3/version1/core" level="3" version="1">
  <model id="nested_piecewise">
    <listOfCompartments><compartment id="c" size="1" constant="true"/></listOfCompartments>
    <listOfParameters>
      <parameter id="p1" constant="false"/>
      <parameter id="p2" constant="false"/>
      <parameter id="p3" constant="false"/>
      <parameter id="p4" constant="false"/>
      <parameter id="x" value="0" constant="false"/>
    </listOfParameters>
    <listOfRules>
      <assignmentRule variable="p1"><math {MML}><cn> 0.001 </cn></math></assignmentRule>

      <assignmentRule variable="p2"><math {MML}><piecewise>
        <piece><cn> 0 </cn><apply><lt/><ci> p1 </ci><cn> 0 </cn></apply></piece>
        <otherwise><ci> p1 </ci></otherwise>
      </piecewise></math></assignmentRule>

      <assignmentRule variable="p3"><math {MML}>
        <apply><plus/><cn> -0.0001 </cn><ci> p2 </ci></apply>
      </math></assignmentRule>

      <assignmentRule variable="p4"><math {MML}><piecewise>
        <piece><cn> 0 </cn><apply><lt/><ci> p3 </ci><cn> 0 </cn></apply></piece>
        <otherwise><ci> p3 </ci></otherwise>
      </piecewise></math></assignmentRule>

      <rateRule variable="x"><math {MML}><ci> p4 </ci></math></rateRule>
    </listOfRules>
  </model>
</sbml>
"""

outdir = tempfile.mkdtemp()
amici.SbmlImporter(SBML, from_file=False).sbml2amici("nested_piecewise", outdir)
model = amici.import_model_module("nested_piecewise", outdir).get_model()
model.set_timepoints(np.array([0.0, 1.0]))
rdata = ss.run_simulation(model, model.create_solver())

ids = list(model.get_expression_ids())
w = np.asarray(rdata.w).reshape(len(rdata.ts), -1)
for name in ids:
    print(f"{name} = {w[0, ids.index(name)]}")
print(f"x(1) = {np.asarray(rdata.x).reshape(len(rdata.ts), -1)[-1, 0]}")
```

It prints the following.

```
p1 = 0.001
p2 = 0.001
p3 = 0.0009
p4 = 0.0
x(1) = 0.0
```

The correct value of `p4` is 0.0009. AMICI reports `p3` as 0.0009, which is positive, so the test `p3 < 0` is false and `p4` should take the otherwise branch.

Two small changes each make the answer correct, and together they locate the problem. If the first piece of `p2` is changed from 0 to 5, `p4` becomes correct. That number cannot change the value of `p2`, because the test `p1 < 0` is false either way and AMICI reports `p2` as 0.001 in both cases. The only thing it changes is what the outer test sees. Separately, if `p1` is replaced by the literal 0.001 inside `p2`, `p4` also becomes correct, so the inner value has to come from a parameter defined by a rule.

This came from the BioModels entry `MODEL0910846879`. That model has one state variable and no reactions, so it can be solved by hand. The exact answer at time 100 is 9.35813562e-4, and AMICI returns 3.499040825e-5, which matches the exact answer for the case where the inner expression is held at zero to ten digits.

The versions used were AMICI 1.0.1, sympy 1.14.0, libSBML 5.21.0, and Python 3.12.12 on macOS 26.6 running on Apple silicon.

FFroehlich:
`Model::reinit_events` computed the Heaviside variables `h` in a single
pass: it evaluated `froot` once and then derived `h` from the root
function values. Root functions are evaluated from `w`, though, and `w`
may itself depend on `h` -- which is exactly what happens when a
piecewise expression occurs inside the condition of another piecewise
expression.

In that situation the outer root function was evaluated using the
Heaviside variables of the inner expression as they were *before* the
same pass updated them (i.e., all zeros at `t0`), so the outer Heaviside
variable was initialized from a stale root value and the wrong branch was
taken for the rest of the simulation.

For the model from #3233 (`w = [p1, p2 = h0*p1, p3 = p2 - 1e-4,
p4 = h2*p3]`, `root = [p1, -p1, p3, -p3]`), `root[2]` was evaluated as
`-1e-4` with `h0` still `0`, so `h2` became `0` and `p4` stayed `0`
although `p3` evaluated to `9e-4` afterwards.

`h` is now iterated to a fixed point (bounded by `ne + 1` iterations, so
cyclic dependencies cannot loop forever), and `roots_found` is derived
from the converged values. As a side effect, `reinit_explicit_roots` now
also sees a consistent `w`. Models whose root functions don't depend on
`h` -- the common case -- just do one additional `froot` evaluation
confirming that nothing changed.

Closes #3233

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_016ZKvTDruCaD3vHVXXXjz8o

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,11 +31,13 @@ See also our [versioning policy](https://amici.readthedocs.io/en/latest/versioni
 
 **Fixes**
 
-* Fixed the sundials backend selecting the wrong branch of a piecewise
+* Fixed both simulation backends selecting the wrong branch of a piecewise
   expression whose condition depends on the value of another piecewise
   expression. The Heaviside variables were initialized in a single pass, so
   the outer condition was evaluated using the not-yet-updated value of the
-  inner expression. They are now iterated to a fixed point (#3233).
+  inner expression. They are now iterated to a fixed point, which also makes
+  the result independent of the (previously differing) value the Heaviside
+  variables happened to be seeded with in either backend (#3233).
 * Demote the module import mtime check during sundials model import to a
   warning, as it can lead to false positives in some environments
   (e.g., when using network file systems with not synchronized clocks).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ See also our [versioning policy](https://amici.readthedocs.io/en/latest/versioni
 
 **Fixes**
 
+* Fixed the sundials backend selecting the wrong branch of a piecewise
+  expression whose condition depends on the value of another piecewise
+  expression. The Heaviside variables were initialized in a single pass, so
+  the outer condition was evaluated using the not-yet-updated value of the
+  inner expression. They are now iterated to a fixed point (#3233).
 * Demote the module import mtime check during sundials model import to a
   warning, as it can lead to false positives in some environments
   (e.g., when using network file systems with not synchronized clocks).
--- a/python/sdist/amici/sim/jax/model.py
+++ b/python/sdist/amici/sim/jax/model.py
@@ -990,7 +990,35 @@ class JAXModel(eqx.Module):
             rf0 = jnp.where(h > 0.5, 0.5, -0.5)
         else:
             h = jnp.where(h_mask, jnp.heaviside(rf0, 0.0), jnp.ones_like(rf0))
-        args = (p, tcl, h)
+
+        # The root functions may themselves depend on the Heaviside variables,
+        # e.g., for a piecewise expression occurring inside the condition of
+        # another piecewise expression. Evaluating them once at the seeded `h`
+        # would use a stale value for the inner expression and select the wrong
+        # branch of the outer one, so refine `h` to a fixed point before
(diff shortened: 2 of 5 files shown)


dweindl (approved):
👍

PR #3235, opened 2026-09-02
Warn about observation-model observables absent from the PySB model

### LINKED ISSUE #918: Sanity check on input for pysb import
- Throw warning if specified observable is not contained in model expressions

Jorge-Polanco-Roque:
Closes #918

## Problem

Observables specified in the `observation_model` (via `MeasurementChannel`) that do not correspond to any observable in the PySB model were **silently ignored**. `_process_pysb_observables` only iterates over `pysb_model.observables`, so a user-supplied observable ID that is absent from the model (e.g. a typo) never takes effect and no feedback is given.

## Change

Before processing, check each `observation_model` key against the set of PySB observable names and emit a `warnings.warn` for any that is not present, so the mismatch is surfaced instead of lost. `warnings.warn` is used (consistent with the SBML importer) so the message is visible independently of the logger verbosity.

## Testing

Added `test_pysb_unknown_observation_model_observable_warns`: an observation model referencing an observable absent from the PySB model triggers a `UserWarning`. The check runs before equation generation, so the test needs no model compilation.

### DIFF
--- a/python/sdist/amici/importers/pysb/__init__.py
+++ b/python/sdist/amici/importers/pysb/__init__.py
@@ -11,6 +11,7 @@ import itertools
 import logging
 import os
 import sys
+import warnings
 from collections.abc import Callable, Iterable
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


FFroehlich (dismissed):
Thank you!

dweindl:
The currently entangled PySB expression/observable handling is indeed something to be cleaned up and made more robust, but it's not that simple and will require some more thought.

Jorge-Polanco-Roque:
Agreed the expression/observable entanglement needs a real cleanup, and I didn't want to take that on here. This PR is deliberately narrow: it just surfaces observation-model observables that are absent from the model (the #918 typo case) with a warning instead of dropping them silently, and doesn't touch the processing logic. I'm happy to keep it scoped as an incremental fix and leave the bigger refactor for a separate effort, but if you'd rather wait for the redesign I'm fine closing this too. Your call.

dweindl:
Thanks for your effort. However, I think this one is going to be easier for me to pick up directly than to review and iterate on. There's a fair bit of context around the MeasurementChannel handling in combination and without PEtab that's hard to convey in review comments. I'm going to close the PR for now and work on it myself when I get to it.

PR #3241, opened 2026-09-04
Add CI check that committed test models are up to date

### LINKED ISSUE #3050: Check for changes in generated model code for test models
Add a workflow to check whether the code generated for the test models in `models/` is up-to-date. Ignore differences in `get_amici_commit()`.

dweindl:
## Summary

- Adds `scripts/run-check-test-models.sh`, which regenerates all test models under `models/` via `amici.testing.models.import_test_models()` and diffs the result against the checked-in version, ignoring differences in the embedded `get_amici_commit()` hash. Any other difference — including files the generator no longer produces, or new files it now produces — fails the check.
- Wires this into `.github/workflows/test_python_cplusplus.yml` as a new step in the `ubuntu-cpp-python-tests` job, gated to `github.event_name == 'merge_group'` (regeneration is slow, so it only runs as part of the merge-queue gate, not on every push/PR).
- The script refuses to run against a dirty `models/` tree, since its cleanup step force-cleans untracked files there.

Closes #3050

## Test plan

- [ ] Merge this PR through the merge queue and confirm the new "Check that test models are up to date" step runs and passes.
- [ ] Sanity-check the failure path: temporarily edit generated code under `models/` (or bump a code-generation-affecting constant) and confirm the step fails with a useful diff.

---
*This PR was prepared with AI assistance (Claude Code).*


models/model_calvetti_py/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+# REMOVE ME
```
dweindl: the merge queue should refuse merging while this is in here.

FFroehlich (approved):

PR #2093, opened 2023-05-16
Add gradient checks for SBML test suite

dilpath:

Issue #3250, opened 2026-09-06
Forward sensitivities for AlgebraicRule models violate the rule's own constraint


dweindl:
## Title
Forward sensitivities for AlgebraicRule models violate the rule's own constraint

### Summary

For models with an `AlgebraicRule`, AMICI's forward sensitivities (`sx`) do
not satisfy the constraint the rule itself imposes. Differentiating the
algebraic constraint gives an exact, model-independent identity that the
reported sensitivities should satisfy at every timepoint; they don't.

Found via finite-difference sensitivity checks newly added to AMICI's SBML
semantic test suite (`tests/sbml/testSBMLSuite.py`). Unlike a typical FD
mismatch, this doesn't require a finite-difference comparison at all — it's
checkable directly from AMICI's own reported `sx`, against the algebraic
rule's own derivative.

### Example (SBML semantic test suite case 00039)

```
addSpecies[ S1, initialAmount -> 0.5 ];  addSpecies[ S2, initialAmount -> 0.5 ];
addParameter[ k1, value -> 1.0 ];        addParameter[ k2, value -> 1 ];
addRule[ type -> AlgebraicRule, math -> S1 + S2 - k1 ];
addReaction[ products -> {S1}, modifiers -> S2, kineticLaw -> k2*S2 ];
```

The constraint `S1 + S2 = k1` implies, for every `t`:
- `d(S1)/dk1 + d(S2)/dk1 = 1`
- `d(S1)/dk2 + d(S2)/dk2 = 0`

AMICI's raw `sx` at the first non-trivial timepoint:

```
sx[t=1] = [[0.1479,  0.8521],     # dS1/dk1, dS1/dk2
           [0.0682, -0.0682]]     # dS2/dk1, dS2/dk2

d(S1)/dk1 + d(S2)/dk1 = 0.1479 + 0.0682 = 0.216   (should be 1.0)
d(S1)/dk2 + d(S2)/dk2 = 0.8521 - 0.0682 = 0.784   (should be 0.0)
```

Both identities are violated. This isn't a magnitude/tolerance issue — the
values are internally inconsistent with the constraint AMICI itself is
supposed to be enforcing, independent of any reference value.

### Suspected area

DAE index-reduction / algebraic-variable elimination used to compute
sensitivities for `AlgebraicRule` models — same subsystem as
#2101 ("Sensitivities for AlgebraicRule models"). The two parameters here
[...]

Issue #3249, opened 2026-09-06
Forward sensitivities silently wrong (reported as 0) when a boundary-condition species' initial value depends on a parameter


dweindl:
### Summary

For a species with `boundaryCondition="true"` whose initial value is set via
`InitialAssignment` from a free parameter, AMICI correctly computes the
sensitivity of the species' own (constant) value, but fails to propagate
that sensitivity into other species whose reaction rates depend on it. The
result is a forward-sensitivity (`sx`) that is silently reported as exactly
zero when it should be nonzero and growing over time.

Found via finite-difference sensitivity checks newly added to AMICI's SBML
semantic test suite (`tests/sbml/testSBMLSuite.py`); confirmed against raw
`rdata.sx` independent of that check, and independent of PEtab/fiddy
entirely. Affects at least 12 cases in the official SBML semantic test
suite (case IDs 00498–00509).

### Reproducer

Two files, no dependencies beyond AMICI itself (`amici==1.1.0`, commit
`9456396bf`, plain SBML import — no PEtab).

`model.xml`:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<sbml xmlns="http://www.sbml.org/sbml/level3/version2/core" level="3" version="2">
  <model id="boundary_species_sensitivity_bug" timeUnits="second">
    <listOfCompartments>
      <compartment id="C" spatialDimensions="3" size="1" constant="true"/>
    </listOfCompartments>
    <listOfSpecies>
      <species id="S1" compartment="C" boundaryCondition="true" hasOnlySubstanceUnits="true" constant="false"/>
      <species id="S2" compartment="C" initialAmount="1.5" hasOnlySubstanceUnits="true" boundaryCondition="false" constant="false"/>
    </listOfSpecies>
    <listOfParameters>
      <parameter id="k1" value="0.75" constant="true"/>
      <parameter id="k2" value="50" constant="true"/>
    </listOfParameters>
    <listOfInitialAssignments>
      <initialAssignment symbol="S1">
        <math xmlns="http://www.w3.org/1998/Math/MathML">
          <apply><times/><ci>k1</ci><ci>S2</ci></apply>
        </math>
      </initialAssignment>
    </listOfInitialAssignments>
    <listOfReactions>
      <reaction id="R1" reversible="false">
        <listOfReactants>
          <speciesReference species="S1" constant="true"/>
        </listOfReactants>
        <listOfProducts>
          <speciesReference species="S2" constant="true"/>
[...]

PR #3253, opened 2026-09-07
CI: sensitivity checks on SBML test suite

### LINKED ISSUE #1236: Perform gradient checks for SBML test suite
Motivated by https://github.com/AMICI-dev/AMICI/issues/1230

Should be done for both forward and adjoint sensitivities (#18).

dweindl:
Add fiddy-based finite differences sensitivity checks to the SBML semantic test suite for every case with free parameters. fiddy's check_jacobian self-calibrates step sizes/tolerances, so no per-model tuning is needed.

This surfaced pre-existing sensitivity bugs tracked as #3249, #3250 and #3257. Those cases are skipped here for now.

Closes #1236
Closes #2093

### DIFF
--- a/tests/sbml/testSBMLSuite.py
+++ b/tests/sbml/testSBMLSuite.py
@@ -7,7 +7,7 @@ Usage:
     pytest tests.sbml.testSBMLSuite -n CORES --dist=loadgroup --cases=SELECTION
         CORES can be an integer or `auto` for all available cores.
         `--dist=loadgroup` is required whenever `-n` is used: each case's
-        simulation and sensitivity checks share one compiled model
+        simulation and sensitivity checks share one compiled model module
         (`compiled_case` fixture) and must run on the same xdist worker.
         SELECTION can be e.g.: `1`, `1,3`, `-3,4,6-7`, or `100-` to select
         specific test cases. If `--cases` is omitted, all cases are run.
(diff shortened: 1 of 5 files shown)


tests/sbml/testSBMLSuite.py
```diff
@@ -123,43 +134,324 @@ def test_sbml_testsuite_case(test_id, compiled_case, result_path):
...
 
     # record results
     write_result_file(simulated, test_id, result_path)
 
 
-def test_sbml_testsuite_case_sensitivity(test_id, compiled_case):
-    """Check the model's sensitivities for the selected cases in
-    `_SENSITIVITY_CHECK_CASES`, via finite differences and against JAX
-    autodiff.
+def _ast_has_piecewise(node: libsbml.ASTNode | None) -> bool:
```
FFroehlich, line 144: File has grown quite a bit, does it make sense to move some of the helpers (and some nested function definitions?) to a separate file?

tests/sbml/testSBMLSuite.py
```diff
@@ -123,43 +134,324 @@ def test_sbml_testsuite_case(test_id, compiled_case, result_path):
...
+            derivative_variables=["llh"],
+        )
+        fsa_expected = fsa_derivative(point)
+        if not fsa_expected:
+            return None
+
+        # ASA
+        asa_solver = model.create_solver()
+        apply_settings(settings, asa_solver, model, test_id)
+        asa_solver.set_absolute_tolerance(
```
FFroehlich, line 391: Code is repeating flavours of this and its hard to grasp whats being done helper that accepts 2 solver instances and list of keys like “atol”, “atolB”? Could be integrated with _run functions

FFroehlich (approved):

tests/sbml/testSBMLSuite.py
dweindl: You are right. These helpers here specifically will be removed shortly as part of #3258. I thought it's probably better to not touch additional files for these temporary things.

tests/sbml/testSBMLSuite.py
dweindl: I see your point. I'd prefer keeping it like that for the moment until I have a couple of other open issues fixed. I think that will get rid of much of the skipping and allow reducing complexity there. I'll do a cleanup pass afterwards (#3262).

PR #3255, opened 2026-09-07
Restore original entity id in JAX-backend outward-facing id lists

dweindl:
The reserved-symbol rename introduced in 9b17bdd6c (#3239) renames any model entity whose id collides with one of amici's fixed array-parameter names (`x`, `p`, `k`, `h`, `w`, `y`) and restores the original id wherever it's reported outward -- but only the SUNDIALS/C++ exporter was updated to do so. The JAX exporter's state/parameter/observable/expression id lists kept reporting the internally-mangled `amici_*` name instead, breaking the SBML Semantic Test Suite JAX CI job for every test case with an entity literally named `x`, `p`, `k`, or `y` (e.g. cases 01184, 01209, 01234, 01240, 01461, ...).

Fixes both id-list builders in `exporters/jax/ode_export.py` to consult the same `reserved_symbol_original_ids` map the C++ exporter already uses, and adds JAX-specific regression tests mirroring the existing C++ ones.

🤖 Generated with Claude Code

previously failed runs: https://github.com/AMICI-dev/AMICI/actions/workflows/test_sbml_semantic_test_suite_jax.yml

### DIFF
--- a/python/tests/test_reserved_symbols.py
+++ b/python/tests/test_reserved_symbols.py
@@ -5,7 +5,10 @@ and for the time-symbol."""
 from pathlib import Path
 
 import numpy as np
-from amici.importers.antimony import antimony2amici
+import pytest
+from amici import import_model_module
+from amici.importers.antimony import antimony2amici, antimony2sbml
+from amici.importers.utils import RESERVED_SYMBOLS
 from amici.sim.sundials import AMICI_SUCCESS
 from amici.testing import skip_on_valgrind
 
@@ -39,12 +42,7 @@ def _without_comments(source: str) -> str:
 # pre-defined floating-point-NaN constant, rejected at parse time regardless
 # of context -- `NULL`/`EOF`/`INFINITY` already cover this class of case.)
 RESERVED_SPECIES_IDS = [
-    "x",
-    "p",
-    "k",
-    "h",
-    "w",
-    "y",
+    *RESERVED_SYMBOLS,
     "NULL",
     "int",
     "class",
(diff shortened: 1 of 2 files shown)


python/tests/test_reserved_symbols.py
```diff
@@ -54,6 +56,14 @@ def _without_comments(source: str) -> str:
     "my__species",
 ]
 
+# Subset of RESERVED_SPECIES_IDS relevant to the JAX backend: only the
+# array-parameter names (x, p, k, h, w, y) get renamed for JAX-specific
+# reasons (see module docstring of exporters/jax/ode_export.py). The
```
FFroehlich: I cannot find such a docstring and it's not immediately obvious to me what those JAX-specific reasons are.

python/tests/test_reserved_symbols.py
```diff
@@ -54,6 +56,14 @@ def _without_comments(source: str) -> str:
...
+# C++-only reserved words in RESERVED_SPECIES_IDS (`NULL`, `int`, `class`,
+# `EOF`, `INFINITY`, ...) are not applicable here -- the JAX backend
+# generates Python source.
+JAX_RESERVED_ARRAY_PARAM_IDS = ["x", "p", "k", "h", "w", "y"]
```
FFroehlich: I don't think such a list should be defined in the tests ...

dweindl: Okay, using `amici.importers.utils.RESERVED_SYMBOLS` now.

FFroehlich (approved):

Issue #3251, opened 2026-09-06
Improve sensitivity checks in test suite


dweindl:
With the revised fiddy design (https://github.com/ICB-DCM/fiddy/pull/80, https://github.com/AMICI-dev/AMICI/pull/3248) we are about to get robust finite difference checks. Once this landed, consider replacing the current amici-specific FD code by fiddy. This will give us higher confidence FDs and will probably uncover some additional issues.
Also, since this hopefully won't require any manual hyperparameter tuning anymore, we should checks sensitivities at every simulation in the test suite. Sensitivity checks for an already imported model are cheap.
There are also a number of places where sensitivity checks are explicitly skipped, completely or for individual parameters or sensitivity modes. Revisit those.

PR #3258, opened 2026-09-08
Fix adjoint sensitivities for events that are both state-triggered and state-updating

### LINKED ISSUE #18: adjoint sensitivies for models with events

### LINKED ISSUE #3257: Adjoint sensitivities wrong for events that are both state-triggered and state-updating
🤖 *Drafted with AI assistance (Claude).*

`sllh` computed via adjoint sensitivity analysis (ASA) diverges from forward sensitivity analysis (FSA) for any parameter that affects the dynamics of a species used in an event's trigger condition, when that same event also updates state. Confirmed via direct FSA-vs-ASA comparison on 40+ SBML semantic test suite cases.

`fdeltaxB`/`fdeltaqB` (generated from `deltaxB`/`deltaqB` in `de_model.py`) embed the trigger-time-sensitivity terms `dtaudx`/`dtaudp`. These are built from the generic `x`/`w` symbols, evaluated at whatever state the function is called with — in practice the **post-event** state (`backwardproblem.cpp`'s `handle_event_b()` calls these with `disc.x_post`). That's the wrong side of the discontinuity: the mathematically equivalent forward-mode quantity (`stau`) is deliberately computed **pre-event**, before the state bolus is applied. For an event whose root doesn't depend on `x` (or that doesn't update state), pre/post make no difference, which is why most cases are unaffected.

Fix implemented, preparing PR.

Related to #18.

dweindl:
deltaxB/deltaqB embed dtaudx/dtaudp (trigger-time-sensitivity terms), but these were evaluated at whatever state the generated function happens to be called with -- in practice the post-event state. That's the wrong side of the discontinuity: the equivalent forward-mode quantity (stau) is deliberately computed pre-event. Fix: eliminate w via its own algebraic definition, then substitute x -> x_old.

Verified via direct FSA-vs-ASA comparison (bypassing FD entirely) on a 72-case corpus of affected SBML semantic test suite models: divergence for the affected parameter previously ranged from ~0.001% up to ~1% relative (worst observed: FSA=0.16456 vs ASA=0.16303); after the fix, it drops to ~1e-13-3e-11 relative (floating-point noise) with zero regressions.

Corpus (SBML semantic test suite case IDs), for future reference/test cases:

- Explicit SBML events (state-triggered and state-updating): 00026, 00041, 00074, 00619, 00620, 00621, 00634, 00635, 00636, 00646, 00647, 00648, 00652, 00653, 00654, 00679, 00680, 00681, 00689, 00700, 00707, 00723, 00736, 00743, 00744, 00745, 00746, 00747, 00749, 00750, 00751, 00752, 00753, 00754, 00755, 00756, 00789, 00790, 00791, 00845, 00846, 00847, 00883, 00884, 00945, 00946, 00947, 00948, 01045, 01046, 01047, 01071, 01072, 01073, 01094, 01106
- Piecewise reaction-rate laws (no explicit <event>, but AMICI hits the same code path via an internal root): 00191, 00192, 00193, 00194, 00198, 00199, 00200, 00201, 00270, 00272, 00273, 00274, 00276, 00277, 00278, 00279

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

Fixes #3257. Still some way to go to fix #18.
Tests are implemented as part of #3251 and #3253.


tests/sbml/testSBMLSuite.py
```diff
@@ -182,6 +153,15 @@ def _model_has_event_jump_risk(sbml_model: libsbml.Model) -> bool:
     "01104": True,
     "01107": True,
     "01148": True,
+    # Adjoint sensitivities & events. To revisit later.
```
dweindl, line 156: Check if we can unskip anything in #3263

FFroehlich (approved):

PR #3266, opened 2026-09-10
Handle roots after reinitialization

dweindl:
CVODES/IDAS sometimes re-detects an event that was just handled after reinitialization and AMICI treated this as simulation failure.

Now we track which root indices were just processed at each `reinit()` and ignore a re-detection of exactly those indices in the following step; report any other index as a genuine (e.g. simultaneous) event instead of aborting.

The "stuck in an event" guard is now a bounded counter instead of an immediate throw, since genuinely simultaneous events could legitimately be discovered one at a time at the same instant.

Related to #2861.

---


dweindl:
It broke my root-after-reinitialization-error example :disappointed:

FFroehlich (approved):

src/solver_cvodes.cpp
```diff
@@ -504,9 +504,30 @@ void CVodeSolver::reset_state(void* ami_mem, const_N_Vector y0) const {
...
+        return false;
+
+    // If a root was found after reinitialization, this may
+    // be a spurious re-detection of the just-handled crossing
+    // rather than a genuine event. Ignore it if it is exactly the set of
+    // roots that were just handled; otherwise, report it.
+    auto cv_mem = static_cast<CVodeMem>(solver_memory_.get());
+    std::vector<int> rootsfound(cv_mem->cv_nrtfn, 0);
+    get_root_info(rootsfound.data());
+    for (int ie = 0; ie < cv_mem->cv_nrtfn; ++ie) {
```
FFroehlich, line 521: Isn't this missing some form of `ws_->sol.t == ws_->tlastroot` check and invalidation of roots_ignored_after_reinit_?

dweindl: I don't think that's needed.
If there is another root, roots_ignored_after_reinit_ will get replaced. 
If there is no other root, then the value of roots_ignored_after_reinit_ doesn't matter.
I don't think we can have a situation where we have two consecutive legit roots for the same event, but I might be missing something.

dweindl: I'll see if I can break it or add an reset and check that it doesn't interfere with persisting trigger state across preeq/presim/...

dweindl: I think you are right at least for the (not too common) case that we are trying to use the same solver instance to solve different models. Then, for example, the first model's index 0 event could prevent handling the second model's index 0 event.
I'll add a reset.

FFroehlich: how does this prevent stopping the same event from triggering twice in a row?

dweindl: Wait. There was a directionality check missing. I will update.

dweindl: Okay, I think I finally understand now that this whole issue is first of all a conceptual question or policy decision rather than an implementation question. For numerical reasons, we can never be sure whether another root should or should not trigger an event during this first post-reinit step. So we can A) follow a never-guess policy and fail, as done so far, at the cost of failing on many benign cases; or B) we ignore any repeated firing of the same event within the solvers' initial step size after its last firing, which means that we could incorrectly miss some events, but succeeds in cases that failed so far; or C) execute the event, possibly until we run out of solver steps, and with an, in practice, probably much higher chance of obtaining incorrect results than B.
In terms of usability, I think, I would prefer B with a warning that an event trigger was ignored and let the user decide based on the model expressions whether this is problem or not.
On second thought, maybe for the time being it's better to invest more effort into recognizing explicitly time-dependent events as such and handling them without the root finding.

FFroehlich: I think adapting the jax implementation of disabling root detection until a root with a derivative with an opposite sign occurs should work. We can produce errors if roots are smaller than some threshold (I don't recall what SBML has to say about such (second order?) roots, I think sundials states that all bets are off whether we can even detect them).

dweindl: Roots found pre-reinitialization are now ignored if they are found again in the post-reinitialization step (with a warning only in case of an upcrossing, i.e. where we could be skipping an event assignment). All others are processed. 
This does not change the root function handed to sundials solvers, just the amici-side processing. This should in my understanding still be consistent with the JAX side.
SBML isn't very specific there. Second-order events have to be processed, but it is noted that all the event handling is subject to numerical errors, and therefore, models should be built sufficiently robust.

PR #3270, opened 2026-09-10
Fix zero dtcldp sensitivity without reintroducing dwdp blow-up

### LINKED ISSUE #3249: Forward sensitivities silently wrong (reported as 0) when a boundary-condition species' initial value depends on a parameter

### LINKED ISSUE #3268: Always-symbolic dtcldp (#3259) blows up generated code size, hangs compilation for large models
🤖 Drafted with AI assistance.

#3259 made `dtcldp` (sensitivity of a conservation law's total abundance w.r.t. free parameters) always a live symbol, never folded to a literal `0`, to fix incorrect zero sensitivities for single-species conservation laws with parameter-dependent initial values.

This is currently hanging/timing out compilation for large models. Confirmed on the `CS_Signalling_ERBB_RAS_AKT` performance-test benchmark:

- `main` before #3259: generated model code ~758 KB, compiles fine.
- `main` at #3259: generated model code ~15.5 MB. `dwdp.cpp` alone grows from 693 KB / 10,012 lines to 148 MB / 2,228,620 lines. Compilation doesn't finish within the CI timeout (observed >2h, still running after 5h in one merge-queue attempt).
- The merge-queue run for #3259 alone (no other PR involved) reproduces the hang, ruling out any other change as the cause.

**Root cause:** `_dtcldp_symbols()` (`de_model.py`, `DEModel._dtcldp_symbols`) builds `dtcldp` as a fully **dense** matrix: one fresh, opaque symbol per `(conservation law, parameter)` pair, regardless of whether that parameter could plausibly affect that law. Because these symbols are opaque, sympy can never simplify them to zero downstream, so every pair gets threaded through `w`/`dwdp`. For this model (~1600 single-species/boundary conservation laws × ~460 free parameters), that's ~740k forced-nonzero entries added to `dwdp`'s sparsity pattern (nnz: 4,216 → 743,752) — nearly all of them trivial 1:1 copies (`dSP_10_..._ = stcl_SP_10_..._;`), but the sheer count overwhelms the compiler.

This currently blocks the merge queue for any PR targeting `main` behind #3259 (Performance Test job hangs until CI timeout).

cc @dweindl

dweindl:
Fixes #3249, fixes #3268.

`dtcldp` was hardcoded to `0` for every parameter whenever a conservation law spanned a single species — wrong whenever that species' own initial value depends on a parameter (#3249). #3259 fixed this by making `dtcldp` always symbolic for every `(conservation law, parameter)` pair, which blew up `dwdp`'s sparsity pattern for models with many trivial single-species conservation laws, hanging compilation (#3268); it was reverted.

Single-species conservation laws always come from states with `dx/dt == 0` identically, so their `dtcldp` sensitivity is fully decidable from the state's own initial-value formula: this emits a symbol only for parameters that actually appear there, literal zero otherwise. Multi-species conservation laws stay fully symbolic, since preequilibration followed by partial reinitialization can make their total depend on parameters invisible to any static formula — this is what PEtab test suite cases 0010/0017 protect against.

🤖 Generated with AI assistance.

:eyes: https://github.com/dweindl/AMICI/actions/runs/34533796660

### DIFF
--- a/python/sdist/amici/_symbolic/de_model.py
+++ b/python/sdist/amici/_symbolic/de_model.py
@@ -1167,22 +1167,7 @@ class DEModel:
             )
             return
         elif name == "dtcldp":
-            # check, whether the CL consists of only one state. Then,
-            # sensitivities drop out, otherwise generate symbols
-            self._syms[name] = sp.Matrix(
-                [
-                    [
-                        sp.Symbol(
-                            f"s{tcl.get_id()}__{par.get_id()}",
-                            real=True,
-                        )
-                        for par in self._free_parameters
-                    ]
-                    if self.conservation_law_has_multispecies(tcl)
-                    else [0] * self.num_par()
-                    for tcl in self._conservation_laws
-                ]
-            )
+            self._syms[name] = self._dtcldp_symbols()
             return
         elif name == "x_old":
             length = len(self.eq("xdot"))
@@ -2473,6 +2458,49 @@ class DEModel:
         state_set = set(self.sym("x_rdata"))
         n_species = len(state_set.intersection(tcl.get_val().free_symbols))
         return n_species > 1
+
+    def _dtcldp_symbols(self) -> sp.Matrix:
+        """
+        Builds the symbol matrix for ``dtcldp``, the sensitivity of each
+        conservation law's total abundance w.r.t. the free parameters.
+
+        Multi-species conservation laws keep every entry as a live symbol:
+        if only some of their states get reinitialized after
+        preequilibration while others keep their (formula-less,
+        dynamics-derived) preequilibration values, the recomputed total can
+        depend on parameters invisible to the model's own ``x0``/``sx0``
+        formulas. Single-species conservation laws only ever arise from
+        states with ``dx/dt == 0`` identically, so they're never touched by
+        dynamics in the first place: their sensitivity is fully determined
+        by the free parameters in their own initial-value formula, and all
+        other entries can stay a literal zero. This keeps ``dtcldp`` (and
+        downstream quantities like ``dwdp``) sparse for models with many
+        trivial single-species conservation laws.
+
+        :return:
+            symbol matrix, one row per conservation law
+        """
+        state_by_sym = {
+            state.get_sym(): state for state in self._differential_states
+        }
+
+        def row(tcl: ConservationLaw) -> list[sp.Expr]:
+            if self.conservation_law_has_multispecies(tcl):
+                return [
+                    symbol_with_assumptions(f"s{tcl.get_id()}__{par.get_id()}")
+                    for par in self._free_parameters
+                ]
+
+            (state_sym,) = tcl.get_val().free_symbols
+            init_free_syms = state_by_sym[state_sym].get_val().free_symbols
+            return [
+                symbol_with_assumptions(f"s{tcl.get_id()}__{par.get_id()}")
+                if par.get_sym() in init_free_syms
+                else 0
+                for par in self._free_parameters
+            ]
+
+        return sp.Matrix([row(tcl) for tcl in self._conservation_laws])
 
     def _expr_is_time_dependent(self, expr: sp.Expr) -> bool:
         """Determine whether an expression is time-dependent.
--- a/tests/sbml/testSBMLSuite.py
+++ b/tests/sbml/testSBMLSuite.py
@@ -191,15 +191,6 @@ def _sensitivity_preflight_checks(
             "be wrong -- see "
             "https://github.com/AMICI-dev/AMICI/issues/3250"
         )
-    if any(
-        species.getBoundaryCondition() or species.getConstant()
-        for species in sbml_model.getListOfSpecies()
-    ):
-        pytest.skip(
-            "Sensitivities for boundary-condition/constant species "
-            "are known to be wrong -- see "
-            "https://github.com/AMICI-dev/AMICI/issues/3249"
-        )
     if uses_adjoint and model.nx_rdata == 0:
         pytest.skip(
             "Adjoint sensitivities for zero-state models are known to crash."


python/sdist/amici/_symbolic/de_model.py
```diff
@@ -2474,6 +2440,47 @@ def conservation_law_has_multispecies(self, tcl: ConservationLaw) -> bool:
...
+        """
+        Builds the symbol matrix for ``dtcldp``, the sensitivity of each
+        conservation law's total abundance w.r.t. the free parameters.
+
+        Multi-species conservation laws keep every entry as a live symbol:
+        preequilibration can make their total depend on any parameter that
+        affects the dynamics, not just ones visible in the model's own
+        ``x0``/``sx0`` formulas. Single-species conservation laws only ever
+        arise from states with ``dx/dt == 0`` identically, which
+        preequilibration can't move -- so their sensitivity is fully
```
FFroehlich: Even though this implementation looks correct, I just realised that I don’t really agree with the remit that preeq can introduce non-zero sensitivities for totals. Either those entities are conserved and the total does not change and preeq can’t introduce non-zero sensitivities (eqs probably simplify to zero in that case anyways?), or they are not conserved. Are we missing something else? Reinitialisation?

FFroehlich (approved):

python/sdist/amici/_symbolic/de_model.py
dweindl: Yeah, you are right, it's not preeq as such causing the problem, it's the following partial reinitialization. Will reword.

